PR #2, opened 2022-02-17
Added changes when handling 500 and path

Danwakeem:
## Description
I made a small change to also mark any requests as an error request that respond with a status code of 500 or greater.

I also made a change to not save the `/{proxy+}` as the `httpPath` and instead save the full path. This change was made so if we are using an express/koa app and we try and request a route that doesn't exist we will send the full path rather than the `/{proxy+}` path to our telemetry backend 👍


medikoo (approved):
Thank you @Danwakeem !

PR #10, opened 2022-03-09
bug: Fixed more integration tests

Danwakeem:
## Description
Sorry I messed up with the integration test fix 🤦‍♂️ This should actually fix it

### DIFF
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/helper.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/helper.js
@@ -107,8 +107,8 @@ const resourceAttributes = [
   },
   {
     key: 'faas.collector_version',
-    value: '@serverless/aws-lambda-otel-extension-0.1.8',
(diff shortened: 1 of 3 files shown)


packages/aws-lambda-otel-extension/opt/otel-extension/helper.js
```diff
@@ -107,8 +107,8 @@ const resourceAttributes = [
   },
   {
     key: 'faas.collector_version',
-    value: '@serverless/aws-lambda-otel-extension-0.1.8',
-    source: '@serverless/aws-lambda-otel-extension-0.1.8',
+    value: '@serverless/aws-lambda-otel-extension-0.1.9',
+    source: '@serverless/aws-lambda-otel-extension-0.1.9',
```
medikoo, line 111: This ideally should be read automatically, or generated automatically for publication.

I'll solve when working on layer improvements

PR #12, opened 2022-03-15
Split internal and extensions into individual directories

medikoo:
Reorganize code, so:
- `/opt/otel-extension/external` host code specific for external extension
- `/opt/otel-extension/internal` hosts code specific for internal extension (wrapper)

As exec handler of wrapper was moved into `/opt/otel-extension/internal` it implies a breaking change, so it should be followed with `0.2 `release, and update in a framework so new path to handler is used

Additional changes:
- Remove logic that rewritten `OTEL_RESOURCE_ATTRIBUTES` if it didn't include `service.name`, it appears, as following secures `service.name` to match function name: https://github.com/serverless/runtime/blob/7c37c33cf63034bb9d3443171729bbbfeaa0f483/packages/aws-lambda-otel-extension/opt/otel-extension/helper.js#L45-L46 and https://github.com/serverless/runtime/blob/7c37c33cf63034bb9d3443171729bbbfeaa0f483/packages/aws-lambda-otel-extension/opt/otel-extension/wrapper.js#L288
- Unify file naming, so there's just hyphen separated names
- Ensure tests are not affected by state of cache files eventually being created locally

-- 

Run, integration test all the reports are generated as before


Danwakeem (approved):
Looking good 👍

PR #14, opened 2022-03-16
chore: added original properties and rawHttp path

Danwakeem:
## Description
Added original properties and raw http path attributes.

- original properties are added so that we know which attributes were added via otel and which attributes were added by our event detector
- raw http path was added so that we would know what the requested path is supposed to be without introducing cardinality issues with influxDB
- Also removed the prod build thing since it was broken anyway from the file restructure


packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
```diff
@@ -171,13 +171,17 @@ const responseHandler = async (span, { res, err }, isTimeout) => {
         const startTime = val.startTime || [0, 0];
         const endTime = val.endTime || [0, 0];
 
-        let attributes = val.attributes;
+        let attributes = {
+          ...val.attributes,
+          'sls.original_properties': Object.keys(val.attributes).join(','),
```
medikoo, line 176: What's the purpose of this change?

medikoo (commented):
Thanks @Danwakeem 

Important fix, I have just one minor comment, also let's maybe not add `.DS_Store` files (you can add them to global ignore file to avoid that: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
Danwakeem: This is added so that we can distinguish between which attributes come directly from open telemetry and which come come from our instrumentation 😎

medikoo: Thanks, why exactly backend needs to know that?

Danwakeem: There are a few places in the UI where we want to only show open telemetry tags. More specifically this view in console 👇 
Screen Shot 2022-03-16 at 10 56 50 AM

medikoo: And why exactly it's just otel originated properties here? Isn't it more about those specific set of properties?

e.g. if otel would add `headers` is it ok for us, that it'll auto-magically appear in that section ?

medikoo: Sorry, for being picky. I'm also asking because goal is to have this package Serverless Console agnostic. We should also advertise as it as good to e.g. use with Datadog or other monitoring services.

I wonder if it's the right logical place for this logic

Danwakeem: > e.g. if otel would add headers is it ok for us, that it'll auto-magically appear in that section ?

from my understanding yes. We want a way to determine which attributes were generated by us and open telemetry. For context in one of the helper files we attach a set of custom attributes to each span so we wouldn't know which ones are attached from our instrumentation or the open telemetry library. 

> I wonder if it's the right logical place for this logic

No worries these are good questions. If we didn’t somehow list out the keys that came from the open telemetry library here I don’t know how else we could determine which attributes came from otel and which came from our helper functions.

We really won’t know what attributes are being created by the otel libraries ahead of time so unless we saved what keys came out of the span attributes here I don’t know how else we could figure out which attributes are from otel or from us.

> We should also advertise as it as good to e.g. use with Datadog or other monitoring services.

This would not break the integration with any other platform. In each platform (data dog, etc.) they will have an option to specify which attributes they care about and which attributes they don’t. If they don’t want this one (or any of the many other attributes we are adding) they can simply ignore this.

medikoo: Ok, and `source` on attribute value objects doesn't already answer that? Maybe we can distinguish it somehow that way, or add another property to those values?

It doesn't feel clean to create otel property, which is not about telemetry but about the categorization of otel properties.

Danwakeem: > Ok, and source on attribute value objects doesn't already answer that? Maybe we can distinguish it somehow that way, or add another property to those values?

Unfortunately no because that is not the shape of the object that is sent out to our APIs. And the trouble with adding new properties to the payload is that we are trying to keep this as otel compliant as possible so when we package up the trace/metrics payloads in the open telemeptry proto format so we cannot add properties to that payload since it has to match the proto format.

> It doesn't feel clean to create otel property, which is not about telemetry but about the categorization of otel properties.

I do see what you are saying but this is just another attribute that our platform needs. We are already using the attribute section to define other sls platform specific properties so I am not sure I understand what issue there is with adding one more attribute here and it seems to make the most sense to add this attribute in the same place we are attaching other sls specific properties.

Is the issue that we don't like where attribute is being aded in the code or that we don't want this attribute at all?

Danwakeem: I might have misunderstood what you meant by this 👇 

> Ok, and source on attribute value objects doesn't already answer that? Maybe we can distinguish it somehow that way, or add another property to those values?

Are you suggesting we should add this attribute in the `otel-payloads` file instead of adding it here?

medikoo: > Is the issue that we don't like where attribute is being aded in the code or that we don't want this attribute at all?

Issue is that we mix namespaces, and produce dirty dataset, by example following feels wrong:

```javascript
const personMeta = {
  properties: {
    name: "John",
    surname: "Doe",
    age: 37,
    nameRelatedPropertyNames: ["name", "surname"]
  }
}
```

It's quite problematic for any generic processor (which may not follow the meaning of each defined property, but would like to present them all). I believe way better to construct this would be as:

```javascript
const personMeta = {
  properties: {
    name: "John",
    surname: "Doe",
    age: 37,
  },
  nameRelatedPropertyNames: ["name", "surname"]
}
```

Still, question is, can we do something like that here?

medikoo: > Are you suggesting we should add this attribute in the otel-payloads ile instead of adding it here?


if I see correctly, attributes are constructed as:

```javascript
[
  {
    key: 'faas.id',
    source: 'computeCustomArn',
    type: 'stringValue',
  },
  {
    key: 'faas.name',
    source: 'functionName',
    type: 'stringValue',
  }
]
```

So I was thinking that maybe if we'd prefix `source` of each of our property with `sls.` it'll give same effect. Or maybe we can add extra property e.g. `isSlsProperty: true` 

But I'm not yet that familiar with otel schema, not sure what can work best

Danwakeem: I don't think so. The otel schema doesn't support anything outside of these attributes.

```
export interface ReadableSpan {
    readonly name: string;
    readonly kind: SpanKind;
    readonly parentSpanId?: string;
    readonly startTime: HrTime;
    readonly endTime: HrTime;
    readonly status: SpanStatus;
    readonly attributes: SpanAttributes;
    readonly links: Link[];
    readonly events: TimedEvent[];
    readonly duration: HrTime;
    readonly ended: boolean;
    readonly resource: Resource;
    readonly instrumentationLibrary: InstrumentationLibrary;
}
```

medikoo: Where exactly I can find this schemas?

Danwakeem: ~TBH I don't know where they have this posted online this has been one of the more frustrating things looking at the OTEL spec.~

~To get this info I have been been referencing the typescript files on the trace library 👇~

```
node_modules/@opentelemetry/sdk-trace-base/build/src/export/ReadableSpan.d.ts
```

Ignore this comment we need to be looking at the `trace.proto` span definitions 🤦

Danwakeem: The other place to reference would be the `trace.proto` file. This describes what the data must look like when we send it out to our API (Or any other otel supported API)

medikoo: I found it in code: https://github.com/open-telemetry/opentelemetry-js/blob/038bbdb67ad16d6fd8a26a3a81e42726ceccf04e/packages/opentelemetry-sdk-trace-base/src/export/ReadableSpan.ts

Still, it appears we already break that, at least I see we define properties not listed above, here:
https://github.com/serverless/runtime/blob/14631f0784528046f655b0030fed7a70635311a6/packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js#L184-L199

Do I miss something?

Danwakeem: Yeah sorry I probably pointed you toward the wrong definition. Looking at this again that `ReadableSpan` definition is different from the one in the `trace.proto`.

The definition in the `trace.proto` type is the definition that is used when you send data to an API and that is what this trace payload is being mapped to.

medikoo: Is content of `trace.proto` copied from some other location? (if it is, then from where exactly?)

If it's part of this package, then it means we control it, and we can extend it, isn't ?

Danwakeem: It is [from here](https://github.com/open-telemetry/opentelemetry-proto). I just wasn’t sure how to pull that proto file into the project from that repo since I couldn’t find an npm package for it. 🤷‍♂️ So while it is in the project we do not dictate what the data structure looks like.

medikoo: I see it's not an npm package, can we put comments in those files indicating where are they copied from?

Otherwise, let's take it in, eventually, when refactoring I'll think on whether we can improve that (it smells as extra technical debt now)

Danwakeem: Yeah that is a good call I can add that in this PR 👍

Also just for more context there is one more challenge with those files. We need to be able to keep those in sync with our backend ingestion app. The backend app is using the same proto files to decode payloads. We have not quite figured out how to keep both of these projects in sync. Just wanted to share this so you know that at some point we will need to figure that out as well 👍

medikoo: Great point @Danwakeem I'll take that into account

medikoo (dismissed):
👍

PR #15, opened 2022-03-18
feat: Added support for log streaming

Danwakeem:
## Description
This PR is dropping in a hook that will send logs to a logging endpoint as soon as we receive a log message from the lambda API.

> We are only going to share the `SLS_OTEL_REPORT_LOGS_URL` internally for now 👍


packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -4,6 +4,13 @@ const http = require('http');
 const { logMessage } = require('../../lib/helper');
 const { SAVE_FILE } = require('../helper');
 const { writeFileSync } = require('fs');
+const reportOtelData = require('./../report-otel-data');
+const { createLogPayload } = require('../otel-payloads');
+
+// eslint-disable-next-line prefer-const
+let liveLogData = [];
+// eslint-disable-next-line prefer-const
+let mainEventData = {};
```
medikoo: Let's remove those eslint disable comments

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -23,10 +30,17 @@ function listen({ port, address, logsQueue, callback }) {
           }
           if (batch.length > 0) {
             const logBatch = batch.filter((log) => {
+              logMessage('Filters: ', JSON.stringify(log));
               if (log.type === 'platform.report') {
                 return true;
-              } else if (log.type === 'function' && log.record.includes('⚡.')) {
+              } else if (
+                log.type === 'function' &&
+                typeof log.record === 'string' &&
```
medikoo: Isn't `log.record` always there and always a string in such case?

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -23,10 +30,17 @@ function listen({ port, address, logsQueue, callback }) {
...
+                log.record.includes('⚡.')
+              ) {
                 return true;
+              } else if (log.type === 'function' && (log.record || {}).recordType === 'eventData') {
```
medikoo: Is there a possoble scenarion when `log.type === 'function'` and `log.record` is not a string?

When I follow this doc: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-logs-api.html#runtimes-logs-api-msg-function I don't see such option (but I'm also not sure how complete are those docs)

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -43,6 +57,36 @@ function listen({ port, address, logsQueue, callback }) {
           if (!address) {
             logMessage('FROM CUSTOM HTTP SERVER: ', JSON.stringify(logsQueue));
           }
+
+          const reportedLogs = batch.filter((log) => {
+            if (/platform/i.test(log.type)) {
```
medikoo: Probably cleaner would be `log.type.startsWith('platform.')`

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -43,6 +57,36 @@ function listen({ port, address, logsQueue, callback }) {
...
+              return false;
+            } else if (
+              log.type === 'function' &&
+              typeof log.record === 'object' &&
+              log.record.recordType === 'eventData' &&
+              log.record.eventData
+            ) {
+              return false;
+            }
+            return true;
```
medikoo: Is that filter a duplicate of logic at L33-L43 ? If so maybe we can generalize it?

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -23,10 +30,17 @@ function listen({ port, address, logsQueue, callback }) {
...
           if (batch.length > 0) {
             const logBatch = batch.filter((log) => {
+              logMessage('Filters: ', JSON.stringify(log));
               if (log.type === 'platform.report') {
                 return true;
-              } else if (log.type === 'function' && log.record.includes('⚡.')) {
+              } else if (
+                log.type === 'function' &&
+                typeof log.record === 'string' &&
+                log.record.includes('⚡.')
```
medikoo: Do we still write `'⚡.'`  log? Is this still needed?

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
```diff
@@ -388,6 +460,9 @@ const createTracePayload = (groupedByRequestId, sentRequests) =>
     .reduce((arr, originalList) => [...arr, ...originalList], []);
 
 module.exports = {
+  createMetricAttributes,
+  createResourceAttributes,
```
medikoo: Why do we export those two properties?

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
```diff
@@ -118,6 +118,78 @@ const createResourceAttributes = (fun) =>
     }
   });
 
+const createLogPayload = (fun, logs) => {
```
medikoo, line 121: This file is named `otel-payloads` and if I see correctly this function is not about hotel. Maybe we can instead put it in another file as e.g. `generate-logs-payload.js` (?)

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
```diff
@@ -329,6 +329,28 @@ const instrumentations = [
         eventData[context.awsRequestId].eventCustomRequestTimeEpoch =
           event.requestContext.timeEpoch;
       }
+
+      // Send request data to external so that we can attach this data to logs
+      await fetch(`http://localhost:${process.env.MOCK_PORT || OTEL_SERVER_PORT}`, {
```
medikoo: What's the purpose of `process.env.MOCK_PORT` support?

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
```diff
@@ -329,6 +329,28 @@ const instrumentations = [
...
+            record: {
+              recordType: 'eventData',
+              eventData,
+              span: {
+                traceId: span.spanContext().traceId,
+                spanId: span.spanContext().spanId,
+              },
+            },
+          },
+        ]),
```
medikoo: Are we trying here to send non-standard AWS log request into endpoint that listens for AWS logs as reported by native AWS logs API?

If that's the case I would not mix it, as it becomes really confusing, and we risk colliding. Can we setup other server for that? Maybe we can use same server we've used to send function reports (so they're no longer read through logs) ?

packages/aws-lambda-otel-extension/opt/otel-extension/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "@opentelemetry/api": "~1.0.4",
+    "@opentelemetry/api": "~1.1.0",
```
medikoo, line 3: Let's also bump `@opentelemetry/core` to `^1.1.0`

Note I've intentionally narrowed this to "~1.0.4"` as `@opentelemetry/core` at v1.0.1 didn't support `@opentelemetry/api` at v1.1 (I see that v1.1 of `@opentelemetry/core` it's fixed)

packages/aws-lambda-otel-extension/test/unit/internal/index.test.js
```diff
@@ -35,19 +35,26 @@ describe('internal', () => {
           response.end('OK');
           server.close();
           const batch = JSON.parse(body);
-          logsQueue = batch;
+          logsQueue.push(...batch);
 
-          expect(logsQueue.length).to.equal(1);
-          const reportLog = logsQueue[0].record.split('\t')[2];
+          if (logsQueue.length === 2) {
```
medikoo: Wouldn't safer be `logsQueue.length > 1` (?)

medikoo (commented):
Thanks @Danwakeem ! I've posted some comments, let me know what you think

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
Danwakeem: It is always there but now always a string. Platform logs are usually an object and function logs are (i think but not sure) always a string

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
Danwakeem: Yeah this is me just being afraid of the unknown 😉 I haven’t seen function logs come in as anything other than string but I don’t know for sure if that is always the case.

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: This is following the [otel log format](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#example-log-records) 👍

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
Danwakeem: No, this is using the custom http server we are sending the other `⚡.` data through 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
Danwakeem: This is the inverse of L33-L43. In L33-L43 we are looking for specific log events. In this filter we are looking for all logs except those specific log events.

I could combine these two lists if we would rather do it that way. Either way does not matter to me 🤷

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
Danwakeem: We don't write it as a log but this http server file is the one for the custom http server and the logs API server so we would still want to do the check here 👍

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
Danwakeem: This is just copying behavior from the other fetch happing in this file. If I am remembering correctly it is because we set the port to something different during one of the tests 🤷

packages/aws-lambda-otel-extension/test/unit/internal/index.test.js
Danwakeem: ~~No I need to wait for both logs to be added to the queue (Normal log message and lambda metadata event) before our test is valid 👍~~

Scratch that I just did a test with your suggestion and it worked fine 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: This was a mistake 👍

packages/aws-lambda-otel-extension/opt/otel-extension/package.json
Danwakeem: Yeah I had to bump this because when I would run `npm install` it would throw an error about a version mismatch. I just bumped the `@opentelemetry/core` to `^1.1.0` as well 😎

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
medikoo: If it's the inverse then we can create one function named e.g. `isMeaningfulLog`, and then in one filter use it as `filter(isMeaningfulLog)` and in other `filter((...args) => !isMeaningfulLog(...args))`

It'll improve maintenance and reduce risk of unwanted error

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
medikoo: Ok, I'll confirm with AWS support, and eventually clean this later

Danwakeem: Good idea! When you reach out do you just do that through the AWS dashboard or do we have a partner portal or something?

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
medikoo: I see thanks!

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
```diff
@@ -118,6 +118,78 @@ const createResourceAttributes = (fun) =>
...
+    TRACE: 1,
+    DEBUG: 5,
+    INFO: 9,
+    WARN: 13,
+    ERROR: 17,
+    FATAL: 21,
+  };
+
+  return logs.map((log) => {
+    const split = (log.record || '').split('\t');
```
medikoo, line 170: I think format as `2019-04-19T08:07:05.666Z c82d8e46-fcc6-4186-979d-2bc896c1b0b2 INFO <logMessage>` is produced only by logs as written via `console` method.

If someone in code uses `process.std*` (or some log util that uses that internally) then log will be simply as `<logMessage>`.

I assume it's better to take that into account here, to avoid weird processing issues

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
medikoo: Ok, thanks for clarification!

medikoo: But why then we use AWS Lambda API log-like request format. 

It's a bit confusing

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: Ohh okay I didn't realize that. I will do some tests with process.std and make some adjustments 👍

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
medikoo: We don't set MOCK_PORT in any test. It looks as some never used left over

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
medikoo: We have a paid dev tech support in one of the accounts. I just open a ticket there.

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
Danwakeem: `MOCK_PORT` is set in the `packages/aws-lambda-otel-extension/test/unit/internal/index.test.js` file

medikoo: Indeed, sorry I didn't realize that (by mistake I've search in context of extension, ignoring test files)

Danwakeem: All good and looking into this more I don't see why we couldn't use the already defined `OTEL_SERVER_PORT` so I went ahead and removed this `MOCK_PORT` nonsense and the unit tests still pass 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
Danwakeem: Alright I updated this 👍 Looking at both filters there was only one minor difference. Let me know what you think about the change I made

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
medikoo: Great call, thanks @Danwakeem ! 👍

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
Danwakeem: I just though it would make the log filtering easier in the file where we are filtering events from the http server.

I could change the format to something else if we would like. Do you have a preference on the format?

medikoo: What information is important in that payload? Wouldn't payload as `{ eventData, traceId, spanId }` be good enough?

And if it's one of the different requests for this custom server, I'd probably use different paths for it (e.g. `/event-data/` in this case)

medikoo: > I just though it would make the log filtering easier.

Why exactly easier, do we mix these requests, with requests that are coming from AWS logs API? (If so, I would definitely use different processors for them, so it's not that spaghetti)

Danwakeem: > Why exactly easier, do we mix these requests, with requests that are coming from AWS logs API? (If so, I would definitely use different processors for them, so it's not that spaghetti)

Yeah we are using the same http server code for the Lambda Logs API and the custom HTTP server for telemetry data.

I can split these two http servers out but if you wouldn't mind I would prefer to do that work in a separate PR as I think it will be somewhat involved.

medikoo: Actually, that was my question at the top :) I think we should not send some custom requests into a server that's configured to receive data from native AWS logs API.

In my eyes, it's really confusing (it's hard to separate and understand different layers of the logic), and also error-prone.

How important is it to have this PR in shortly?

Maybe we can start with PR that cleans this up, and then follow up with this addition?

Danwakeem: > How important is it to have this PR in shortly?

The other console team mates were hoping to get their hands on this today so they could start testing infra as quickly as possible. I guess if we want to make that refactor first I can just start working on the other PR now.

Danwakeem: Actually I would also prefer to merge this in first so I don't have to refactor both the existing code in main and the code in this PR. If I merge this in then I could just refactor one codebase.

medikoo: Ok, no problem 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: This should be set up to support `process.std*`. It won't pick up the log level but we will have to make some adjustments for the logging frameworks we want to support. I don't have this list just yet but I can create a follow up PR once we know 👍

medikoo: Ok, I would probably address it on case by case basis when demand is shown. 

And anyway we need to agree on how to report typical stdout/err write, which authors may even do directly in lambda.

Can we have log level optional? As then it's technically not a message that is subject to level categorization?

Danwakeem: Yeah I totally agree ☝️ I have a feeling Austen and the gang may already have some libraries in mind they wanted to support (I have heard pino and winston throw arround) but I figured we will have to add support for others as we hear of demand.

> Can we have log level optional? As then it's technically not a message that is subject to level categorization?

Looking back over the OTEL spec it does say that the severity fields are optional so I will take these out if they are not in our log message 👍

Danwakeem: Alright the severity fields should be excluded from the request if they are undefined 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -4,6 +4,26 @@ const http = require('http');
 const { logMessage } = require('../../lib/helper');
 const { SAVE_FILE } = require('../helper');
 const { writeFileSync } = require('fs');
+const reportOtelData = require('./../report-otel-data');
+const { createLogPayload } = require('../otel-payloads');
+
+let liveLogData = [];
+let mainEventData = {};
+
+const meaningfulLog = (log) => {
```
medikoo: `meaningfulLog` suggests that function is a _meaningful log_, maybe we can name it as `isMeaningfulLog` ?

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -43,6 +56,20 @@ function listen({ port, address, logsQueue, callback }) {
...
+          const reportedLogs = batch.filter(
+            (log) => !meaningfulLog(log) && !log.type.startsWith('platform.')
+          );
+
+          liveLogData.push(...(reportedLogs || []));
+
+          if (liveLogData.length > 0 && Object.keys(mainEventData).length > 0) {
+            const sendData = [...liveLogData];
+            liveLogData = [];
+            await reportOtelData.logs(createLogPayload(mainEventData, sendData)).catch((error) => {
```
medikoo: Let's not use `then` or `catch` when we practice async/await (It may make sense when we do some parallel operations, but here it's not the case)

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
```diff
@@ -118,6 +118,76 @@ const createResourceAttributes = (fun) =>
...
+        'faas.api_gateway_app_id',
+      ].includes(attr.key);
+    })
+    .reduce(
+      (obj, attr) => ({
+        ...obj,
+        [attr.key]: Object.values(attr.value)[0],
+      }),
+      {}
+    );
```
medikoo: Wouldn't be more readable as:

```javascript
const metricAttributeNames = new Set(['faas.arch', ....])
const metricAttributes = {};
for (const attribute of createMetricAttributes({ record: fun.eventData[key] }, {})) {
  if (!metricAtributeNames.has(attr.key)) continue;
  metricsAttributes[attr.key] = atr.value;
}
```

I think `reduce` serves a bit different use cases, and with logic as proposed currently, it takes a while to seclude intention behind the logic.

What do you think?

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
```diff
@@ -118,6 +118,76 @@ const createResourceAttributes = (fun) =>
...
+
+  return logs.map((log) => {
+    const split = (log.record || '').split('\t');
+    return {
+      Timestamp: new Date(split[0]).getTime(),
+      Attributes: resourceAtt,
+      Resource: metricsAtt,
+      TraceId: spanData.traceId,
+      SpanId: spanData.spanId,
+      SeverityText: Object.keys(severityNumberMap).includes(split[2]) ? split[2] : undefined,
```
medikoo: It still will pick a string from logs as `foo\tbar\tnonsense` producing an invalid `ServerityText`.

I'd confirm that we definitely deal here with format as `<date>\t<requestId>\t <level>\t<message>` and only in such case resolve level

packages/aws-lambda-otel-extension/opt/otel-extension/external/index.js
```diff
@@ -227,9 +226,22 @@ module.exports = (async function main() {
     logMessage('Remaining logs queue: ', JSON.stringify(logList));
   }
 
-  const { server: otelServer } = listen({
+  const postLiveLogs = async () => {
+    if (liveLogData.logs.length > 0 && Object.keys(mainEventData.data).length > 0) {
+      const sendData = [...liveLogData.logs];
+      liveLogData.logs = [];
+      await reportOtelData.logs(createLogPayload(mainEventData.data, sendData)).catch((error) => {
```
medikoo: Let's not use `catch` (it's like mixing `then` with async/await).

Same can be constructed as:
```javascript
try {
  await reportOtelData.log(...)
} catch (error) {
  logMessage('Failed to send logs', error);
 }
```

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
```diff
@@ -5,31 +5,28 @@ const { logMessage } = require('../../lib/helper');
 const { SAVE_FILE } = require('../helper');
 const { writeFileSync } = require('fs');
 
-function listen({ port, address, logsQueue, callback }) {
+const meaningfulLog = (log) => {
+  if (log.type === 'platform.report') {
+    return true;
+  }
+  return false;
```
medikoo: It can be simplified to `reutrn (log.type === 'platform.report')`

packages/aws-lambda-otel-extension/test/unit/external.test.js
```diff
@@ -96,9 +92,22 @@ describe('external', () => {
       },
     ]);
 
+    await fetch(`http://localhost:${OTEL_SERVER_PORT}`, {
```
medikoo: I think to have proper order, this should be put before previous `logs` emit

medikoo (commented):
Thanks @Danwakeem! 

I've posted a few comments, can you also see my input from the previous review?

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
medikoo: Dan, did you see that one?

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
medikoo: And that one?

Danwakeem: It shouldn't. The `Object.keys(severityNumberMap).includes(split[2])` is ensuring we only sent a valid `SeverityText` value.

Sorry if I am commenting on an older message 😄

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: Ahh sorry I missed this one I can get this updated 👍  I do like your suggestion as you are not the first person to not like my use of the `reduce` function 😉

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
medikoo: ah true :) But can we make pre-prepared list for that (it's a bit weird to generate a new array, every time to confirm)

So something as:

```javascript
const severityLevelNames = new Set( Object.keys(severityNumberMap));
...
severityLevelNames.has(split[2])
```

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: Alright this should be updated to your suggestion 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: Yeah no problem! Just dropped in a commit for this 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
Danwakeem: I went ahead and got rid of this function all together since it is only used once 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
Danwakeem: This was moved around and should not be using `.catch` anymore 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/http-listener.js
medikoo: Great call 👍

packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js
Danwakeem: Im going to resolve this conversation since we ended up doing this in this PR 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/custom-http-listener.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+const http = require('http');
+const { logMessage } = require('../lib/helper');
+const { SAVE_FILE } = require('./helper');
+const { writeFileSync } = require('fs');
+
+// mainEventData is assigned in this file and then used in the parent file
+// eslint-disable-next-line no-unused-vars
```
medikoo: Let's remove this eslint-disable comment (ideally they never should be committed in).

Also I don't understand the comment above

packages/aws-lambda-otel-extension/opt/otel-extension/external/custom-http-listener.js
```diff
@@ -0,0 +1,53 @@
...
+function customListen({ port, logsQueue, mainEventData, callback, liveLogCallback }) {
+  // init HTTP server for the Logs API subscription
```
medikoo: I'd remove this comment, and instead make `customListen` variable name self explanatory.

How about `initializeLogsApiListenerServer` ?

packages/aws-lambda-otel-extension/opt/otel-extension/external/custom-http-listener.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
```
medikoo: Can we give a more meaningful name for this file?

e.g. `logs-api-listener-server.js` ?

packages/aws-lambda-otel-extension/opt/otel-extension/external/custom-http-listener.js
```diff
@@ -0,0 +1,53 @@
...
+      response.end('OK');
+    }
+  });
+
+  server.listen(port);
+  return { logsQueue, server };
+}
+
+module.exports = {
+  customListen,
```
medikoo: If it's the only value, we can export the function directly (no need to expose dictionary)

packages/aws-lambda-otel-extension/opt/otel-extension/external/custom-http-listener.js
Danwakeem: I went ahead and renamed this to `initializeTelemetryListener` and renamed the file to `initialize-telemetry-listener`. Let me know what you think about those names 👍

Danwakeem: I didn't want to call it `LogsApiListener` because this is listening for telemetry data from OTEL not from the Lambda logs API 👍

medikoo (commented):
Thanks @Danwakeem ! Did you confirm that the integration test pass locally with those changes?

Danwakeem:
@medikoo No let me try and run that real quick 🤔

Danwakeem:
@medikoo Okay I just ran the integration tests and they passed 👍

medikoo (approved):
👍

PR #16, opened 2022-03-22
Fix otel log payload

Danwakeem:
## Description
The format of the resources/attributes was not correct so this PR just just adjusting the format of the output 👍

### DIFF
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
@@ -107,8 +107,8 @@ const resourceAttributes = [
   },
   {
     key: 'faas.collector_version',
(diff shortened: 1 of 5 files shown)


medikoo (commented):
Can we back with some regression test?

Danwakeem:
@medikoo No problem I added a log test here 😎 Thanks for the tip on adding that test 👏 

There is one kind of odd thing I did here which was add an environment variable called `SLS_TEST_PRINT_LOG_EVENT`. I added this because the current version of the serverless cli will not add a `SLS_OTEL_REPORT_LOGS_URL` yet and I didn't want this to do a console.log in a live serverless app 👍 

Let me know if you think I should change that in some way.

medikoo (approved):
Looks great 👍

PR #17, opened 2022-03-24
Remove repeated log tags (severity and timestamp)

rfaga:


medikoo (commented):
Hi @rfaga as mentioned on Slack, assuming that this extension is intended to be backend agnostic, I wonder whether we should manipulate AWS logs for our specific case here.

@Danwakeem what are your thoughts?

Danwakeem:
@medikoo This is a really good point 🤔 Given @rfaga mentioned we are already cutting this data out on the backend I am thinking we should not do this here.

Danwakeem:
I also did some testing this morning and this kind of formatting would only apply to messages written with `console.*` functions so it also would vary from logging framework to logging framework 🤔

rfaga:
Agree, I'm closing this change then

PR #18, opened 2022-03-24
chore: Fixed timestamp and add order property

Danwakeem:
## Description
I am fixing the timestamp property to use the timestamp that is attached to the log api record and adding an order property to the log record that is sent to the API so we know what order the logs came in at.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202009458998787

### DIFF
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/index.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/index.js
@@ -253,7 +253,6 @@ module.exports = (async function main() {
   const { server } = listen({
     port: RECEIVER_PORT,
     address: receiverAddress(),
-    mainEventData,
     logsQueue,
     liveLogData,
     liveLogCallback: postLiveLogs,
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
@@ -177,7 +177,7 @@ const createLogPayload = (fun, logs) => {
       SeverityText: severityLevelNames.has(split[2]) ? split[2] : undefined,
       SeverityNumber: severityNumberMap[split[2]],
       Body: log.record || '',
-      ReceptionOrderId: process.hrtime.bigint().toString(),
+      ProcessingOrderId: process.hrtime.bigint().toString(),
     };
   });
 };
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
@@ -177,6 +177,7 @@ const createLogPayload = (fun, logs) => {
       SeverityText: severityLevelNames.has(split[2]) ? split[2] : undefined,
       SeverityNumber: severityNumberMap[split[2]],
       Body: log.record || '',
+      order: process.hrtime.bigint().toString(),
     };
   });
 };
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
@@ -177,7 +177,7 @@ const createLogPayload = (fun, logs) => {
       SeverityText: severityLevelNames.has(split[2]) ? split[2] : undefined,
       SeverityNumber: severityNumberMap[split[2]],
       Body: log.record || '',
-      order: process.hrtime.bigint().toString(),
+      ReceptionOrderId: process.hrtime.bigint().toString(),
     };
   });
 };
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
@@ -169,7 +169,7 @@ const createLogPayload = (fun, logs) => {
   return logs.map((log) => {
     const split = (log.record || '').split('\t');
     return {
-      Timestamp: split[0] ? new Date(split[0]).getTime() : new Date().getTime(),
+      Timestamp: new Date(log.time).getTime(),
       Attributes: resourceAtt,
       Resource: metricsAtt,
       TraceId: spanData.traceId,
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
@@ -107,8 +107,8 @@ const resourceAttributes = [
   },
   {
     key: 'faas.collector_version',
-    value: '@serverless/aws-lambda-otel-extension-0.2.4',
-    source: '@serverless/aws-lambda-otel-extension-0.2.4',
+    value: '@serverless/aws-lambda-otel-extension-0.2.5',
+    source: '@serverless/aws-lambda-otel-extension-0.2.5',
     type: 'stringValue',
   },
 ];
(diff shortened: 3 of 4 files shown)


packages/aws-lambda-otel-extension/opt/otel-extension/external/index.js
```diff
@@ -253,7 +253,6 @@ module.exports = (async function main() {
   const { server } = listen({
     port: RECEIVER_PORT,
     address: receiverAddress(),
-    mainEventData,
```
Danwakeem, line 256: Removing this because it was not needed in this file and was left over from a previous commit 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
```diff
@@ -169,14 +169,15 @@ const createLogPayload = (fun, logs) => {
   return logs.map((log) => {
     const split = (log.record || '').split('\t');
     return {
-      Timestamp: split[0] ? new Date(split[0]).getTime() : new Date().getTime(),
```
medikoo, line 172: It'll be good to confirm it's actually a date stamp, otherwise we may produce a `NaN` for any non `console.*` log that contains `\t` character

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
```diff
@@ -169,14 +169,15 @@ const createLogPayload = (fun, logs) => {
...
+      Timestamp: new Date(log.time).getTime(),
       Attributes: resourceAtt,
       Resource: metricsAtt,
       TraceId: spanData.traceId,
       SpanId: spanData.spanId,
       SeverityText: severityLevelNames.has(split[2]) ? split[2] : undefined,
       SeverityNumber: severityNumberMap[split[2]],
       Body: log.record || '',
+      order: process.hrtime.bigint().toString(),
```
medikoo: I assume we need that, as for logs for which timestamp is same, order gets broken, right?

Also, aren't we supposed here to follow some Otel spec, for generated payload?

Additionally `order` name (1) doesn't follow naming convention (it's not upper case), and it feels pretty vague (it's hard to assume what it's about just from name), maybe we can name it as e.g. `ReceptionHighResolutionTimestamp` ?

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: @medikoo This is the old code that I am trying to replace with `Timestamp: new Date(log.time).getTime()`. Did you mean to comment on this part?

medikoo: ah.. sorry, not sure where I was looking at, ignore this comment :)

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: > I assume we need that, as for logs for which timestamp is same, order gets broken, right?

This is correct

> Also, aren't we supposed here to follow some Otel spec, for generated payload?

Yes as close as possible but since the OTEL spec if experimental we decided this was an okay change since the current spec doesn't cover the case when timestamps are the same. We are going to work on getting a proposal for the OTEL spec together for adding this property to the log payload 👍 

> Additionally order name (1) doesn't follow naming convention (it's not upper case), and it feels pretty vague (it's hard to assume what it's about just from name), maybe we can name it as
> 
> Wouldn't it be better to name it as e.g. ReceptionHighResolutionTimestamp ?

This is a great suggestion. I am not sure what the `Reception` part of that name would mean but I do like the name `HighResolutionTimestamp` to better fit in with the other attributes 😎

medikoo: >  I do like the name HighResolutionTimestamp to better fit in with the other attributes 

I meant _Reception_ as it's a timestamp when our extension processed that log, and it's not really a timestamp when log was produced (afaik in most cases it'll be 100ms+ delays, right?).

Anyway in such case maybe we can call it `ReceptionOrderId` (?) So we do not attach to fact that it's a timestamp, as _time of reception_ is not a key here, but _order_

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: Awesome, no worries man 😎

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: Okay thanks for the clarification. Yeah this makes sense to me 👍 I like `ReceptionOrderId`. I will make that change here.

Danwakeem:
```suggestion
      ReceptionOrderId: process.hrtime.bigint().toString(),
```

Danwakeem: > afaik in most cases it'll be 100ms+ delays, right? 

This is correct 👍 This is in no way a timestamp for when the log was written because is will almost always be 100ms+ delayed from when the log is actually written. The only accurate time would be the one from `log.time`

Danwakeem: @medikoo actually after thinking on this a bit more I think the word `Reception` is kind of confusing. Maybe we just call it `ProcessOrderId` instead?

medikoo: Maybe `ProcessingOrderId` ? :) I wonder whether _process_ will not get confused with e.g. node.js process.

Still I leave it up to you

Danwakeem: I think I like `ProcessingOrderId` 👍 I think the `OrderId` part makes it distinct enough against the node.js process 😎

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
```diff
@@ -169,14 +169,15 @@ const createLogPayload = (fun, logs) => {
...
-      Timestamp: split[0] ? new Date(split[0]).getTime() : new Date().getTime(),
+      Timestamp: new Date(log.time).getTime(),
       Attributes: resourceAtt,
       Resource: metricsAtt,
       TraceId: spanData.traceId,
       SpanId: spanData.spanId,
       SeverityText: severityLevelNames.has(split[2]) ? split[2] : undefined,
       SeverityNumber: severityNumberMap[split[2]],
       Body: log.record || '',
+      ReceptionOrderId: process.hrtime.bigint().toString(),
```
Danwakeem:
```suggestion
      ProcessingOrderId: process.hrtime.bigint().toString(),
```

packages/aws-lambda-otel-extension/opt/otel-extension/external/otel-payloads.js
Danwakeem: Alright @medikoo this change is made 😎

Danwakeem: I’m going to go ahead and merge this in since all the feedback has been addressed 😎

medikoo (approved):
👍

PR #19, opened 2022-03-25
bug: Scoped log data so it is sent for appropriate event data

Danwakeem:
## Description
After doing some testing we were noticing that logs were being sent to a single trace id when a lambda function is under load. This PR scopes the event data to a specific request id so that we know for sure that a log will be associated with the correct trace id.

I also went ahead and added `@opentelemetry/instrumentation-fastify` instrumentation per request 🤷

### DIFF
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/runtime-api.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/runtime-api.js
@@ -24,6 +24,22 @@ async function getEventData() {
   };
 }
 
+async function getRuntimeEventData() {
+  const res = await fetch(`${baseUrl}/invocation/next`, {
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  });
+
+  if (!res.ok) {
+    console.error('next failed', await res.text());
+    return null;
+  }
+
+  return res;
+}
+
 async function getResponse(requestId) {
   const res = await fetch(`${baseUrl}/invocation/${requestId}/response`, {
     method: 'POST',
@@ -44,4 +60,5 @@ async function getResponse(requestId) {
 module.exports = {
   getEventData,
   getResponse,
+  getRuntimeEventData,
 };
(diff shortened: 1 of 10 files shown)


packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/runtime-api.js
```diff
@@ -24,6 +24,22 @@ async function getEventData() {
   };
 }
 
+async function getRuntimeEventData() {
+  const res = await fetch(`${baseUrl}/invocation/next`, {
```
medikoo: If I see correctly this is a Lambda runtime (not extension) API call.

Why exactly do we fall into that? Are there no implication on relying on it in lambda extensions ? (afaik it's dedicated for custom runtime implementations not extensions).

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/runtime-api.js
```diff
@@ -24,6 +24,22 @@ async function getEventData() {
   };
 }
 
+async function getRuntimeEventData() {
```
medikoo: I see this method is added, while we already have `getEventData`. Why we do not reuse `getEventData` (which if I see correctly is not used anywhere, so if not used should be removed)

packages/aws-lambda-otel-extension/opt/otel-extension/external/lambda-apis/runtime-api.js
Mmarzex: I don't think we need to use the runtime API for the `requestId`, it looks like we can get the `requestId` from the extension API call. I'll test it and update this if it works since Dan is out for the week.

Mmarzex: This should be fixed now to use the `requestId` that comes from the extension API

medikoo (approved):
Thank you @Mmarzex! It looks way better now.

I've also removed the not needed changes to test files

PR #24, opened 2022-04-01
Restrict eventData for only current requestId

rfaga:


medikoo (approved):
Thanks @rfaga ! I trust this does the fix. 

Tests at this point do not confirm on `requestId`, I'll add such checks next week

PR #25, opened 2022-04-04
Collect Lambda Request and Response Data

Mmarzex:
This PR enables the sending of the raw lambda event data to our ingest, as well as response data.


Danwakeem (commented):
This is looking good 👍 I wasn't sure what the `SLS_OTEL_REPORT_REQUEST_RESPONSE_URL` should be so I haven't tested with that just yet.

Danwakeem (approved):
Looks good and works fine with or without our environment variable set 👍 I think it is good enough to ship so we can start getting the backend configured 😎

PR #27, opened 2022-04-05
Initial documentation

medikoo:
Describes how to set up and test the extension.

Soon to be extended with complete reference on generated payload

/cc @Danwakeem @hev

### DIFF
--- a/packages/aws-lambda-otel-extension/README.md
+++ b/packages/aws-lambda-otel-extension/README.md
@@ -1,5 +1,100 @@
 # @serverless/aws-lambda-otel-extension
 
-## AWS Lambda extension that gathers otel telemetry and sends to designated url.
+## AWS Lambda extension that gathers [OpenTelemetry](https://opentelemetry.io/) data and sends it to the designated destination
 
-_Note: Package it's still in experimental stage and subject to major changes._
+_Package is still in an experimental stage and subject to major changes._
+
+_At this point, only Node.js runtime is supported, but this support will be extended in the near future_
+
+### How does it work?
+
+The extension is a lambda layer with a source maintained in [opt](opt) folder. There are two extensions configured within a layer:
+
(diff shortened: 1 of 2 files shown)


packages/aws-lambda-otel-extension/README.md
```diff
@@ -1,5 +1,100 @@
...
+
+Ensure that layer version ARN is listed in Lambda layers.
+
+Ensure that internal extension of a layer is pre-loaded by configuring `AWS_LAMBDA_EXEC_WRAPPER` environment variable with `/opt/otel-extension/internal/exec-wrapper.sh`
+
+##### Configure destination endpoints of telemetry payloads
+
+Generated payload can be propagated to the following destinations:
+
+_At this point, only one destination type can be configured, but that's subject to change in the future_
```
Danwakeem, line 52: It might be worth mentioning which environment variables take precedent over the others. 

If you set these environment variables (`SLS_OTEL_REPORT_<type>_URL`, `SLS_OTEL_REPORT_S3_BUCKET`, and `SLS_TEST_PRINT_LOG_EVENT`) we try to send them in the following order

1. External server
2. S3 Bucket
3. Inline logs

medikoo: I wonder if this really important, and can't we treat it as an implementation detail (?)

If we mention that we only support one configuration type, It's clear that any configuration where more than one is configured can be treated as erroneous and may produce surprising results.

Documenting precedence order, adds the feeling that it's an intended _feature_, where it isn't,  also I don't want to create too broad documentation on that part, as to how it works currently is not our goal. I believe we should support all of those destinations at once, and I believe it'll be the case quite soon.

Danwakeem: Okay yeah that also sounds good 👍

Danwakeem (approved):
LGTM 👍 I think it will be good to get Adam's input though since I have been working on this quite a bit I already know most of the ins and outs 😎

medikoo:
> LGTM 👍 I think it will be good to get Adam's input though since I have been working on this quite a bit I already know most of the ins and outs 😎

Sure. It's the initial version, nothing to be set in stone, and I expect to introduce constant improvements to it (especially that  a lot work will happen in the extension in the next weeks)

Anyway let's wait for @hev for his suggestions

hev:
This all LGTM and is a big improvement.  

As follow up we need to document what console endpoints exist, and what Authentication is present on them. These details probably should live within console docs.

PR #28, opened 2022-04-06
Parse Http Route for Fastify

Mmarzex:
This makes fastify routes parse correctly into `http.path`


Danwakeem (approved):
🧠 Oh good catch I totally forgot to do this 👍

PR #29, opened 2022-04-06
Otel instrumentation optimizations

medikoo:
When learning more about how instrumentation internals work, I've come up with few improvements:

- Register initialized instrumentations once (it's fine to extend _resource_ of tracer after it's initialized)
- Introduce patch that will prevent misleading `No modules instrumentation has been defined, nothing will be patched` warning as produced by AWS Lambda instrumentation (we replaced its patcher with ours, so indeed otel internally sees no modules to patch)


Danwakeem (approved):
This seems to work fine but I am a little nervous there is an edge case we don't know about with the change to `internal/index.js`. I had followed the implementation directly from the [otel lambda library here](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/nodejs/packages/layer/src/wrapper.ts).

When I am testing this change in my test lambda functions it does seem to work alright though 🤷

medikoo:
> When I am testing this change in my test lambda functions it does seem to work alright though 🤷

I followed the internals carefully, and that should work without issues, also I've confirmed on that with integration tests.

PR #31, opened 2022-04-07
Automate extension version resolution

medikoo:
So we don't have to update it manually with every release


Danwakeem (approved):
Ahh very nice 😎

PR #33, opened 2022-04-07
Test: Workaround retryable errors

medikoo:
Last CI build failed due to `Rate exceeded` error: https://github.com/serverless/runtime/runs/5868467556?check_suite_focus=true

This patch refactors our integration test, so we rely on util that mitigates eventual retryable errors


Danwakeem (approved):
Another cool utility I will have to try out 👍

PR #34, opened 2022-04-08
Remove obsolete inner env payload compression

medikoo:
We originally used compression to have a written log (which ended in CloudWatch) as small as possible, so it doesn't affect customers' billing that much.

But after @Danwakeem brought excellent improvement of not writing that log, but instead of sending it directly to server-initiated by external extension, this additional compression is no longer needed, and just makes maintenance more difficult.

This change ensures we no longer do that extra compression.


Danwakeem (approved):
🎸

PR #35, opened 2022-04-13
fix: Include spanId in req/res data

Danwakeem:
## Description
We ended up needing the spanId in the request/response payload so we can accurately match it to a specific trace span 👍


Mmarzex (approved):
🚀

PR #38, opened 2022-04-21
CI: Fix release notes publications step

medikoo:
CI fail exposed that resolution of Tag from Github ref did not work as expected. This patch fixes that


Danwakeem (approved):
🔨

PR #39, opened 2022-04-21
Centralize log listener server handling and support complex user configuration input

medikoo:
- Centralize log listener server setup into single module
- Cleanup internals of log listener server
- Support complex user configuration input to be injected via `SLS_OTEL_USER_SETTINGS` environment variable, with `disableLogsMonitoring` and `disableRequestResponseMonitoring` settings


packages/aws-lambda-otel-extension/test/utils/get-extension-server-mock.js
```diff
@@ -109,8 +109,8 @@ module.exports = (emitter) => {
                 state: 'Subscribed',
                 types: data.types,
               },
-            }
-          ).then((res) => res.text());
+            },
+          ]).then((res) => res.text());
```
Danwakeem, line 113: Did you mean to change the second param to an array?

Danwakeem (approved):
This is looking good to me 👍 

There was only one change in the test that I wasn't sure was needed but beyond that I think it is a good feature and organizational change 😎

packages/aws-lambda-otel-extension/test/utils/get-extension-server-mock.js
medikoo: Yes. It's then passed as logs payload, and and AWS side in all cases it's an array.

When I was refactoring I observed that due this not being an array, we had one failure that was muted by try/catch clause which wasn't picking this up.

Danwakeem: Ahh okay nice catch 👍

PR #44, opened 2022-04-27
feat: Clean responses that are not JSON

Danwakeem:
## Description
We were looking to clean the responses from lambda so that they do not contain unusable/unwanted objects like image blobs, strings, etc.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202191846621943

### DIFF
--- a/packages/aws-lambda-otel-extension/README.md
+++ b/packages/aws-lambda-otel-extension/README.md
@@ -62,6 +62,34 @@ Configure following environment variables with designated server urls:
 
 Additionally, through `SLS_OTEL_REPORT_REQUEST_HEADERS` environment variable, extra request headers can be configured, that will be sent with every request to each of the configured urls
 
+It is important to note that lambda response data sent to `SLS_OTEL_REPORT_REQUEST_RESPONSE_URL` will ignore any non JSON objects in either a simple response or a HTTP response payload.
+
+For example, the following will be sent to the `SLS_OTEL_REPORT_REQUEST_RESPONSE_URL` without any modifications 👇
+
+```json
+{ "message": "lambda response" }
+```
+
+```json
+{ "statusCode": 200, "body": "{\"message\": \"lambda response\"}" }
+```
+
+The following responses will be modified when sent to the `SLS_OTEL_REPORT_REQUEST_RESPONSE_URL` endpoint 👇
+
+_Simple non JSON object responses will be ignored_
+
+| Input             | Output    |
+| ----------------- | --------- |
+| "Lambda response" | _Ignored_ |
+| true              | _Ignored_ |
+| 12                | _Ignored_ |
+
+_HTTP Responses with a non JSON body will be ignored_
+
+| Input                                              | Output                  |
+| -------------------------------------------------- | ----------------------- |
+| `{ "statusCode": 200, "body": "lambda response" }` | `{ "statusCode": 200 }` |
+
 ###### S3 bucket
 
 Configure `SLS_OTEL_REPORT_S3_BUCKET` with bucket name (and ensure that Lambda has needed rights to write to the bucket)
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
@@ -1,4 +1,5 @@
 'use strict';
+const isObject = require('lodash.isobject');
 
 const extensionVersion = (() => {
   try {
@@ -9,6 +10,33 @@ const extensionVersion = (() => {
   // eslint-disable-next-line import/no-unresolved
   return require('../version');
 })();
+
+const isJson = (data) => {
+  try {
+    JSON.parse(data);
+    return true;
+  } catch (error) {
+    return null;
+  }
+};
+const stripResponseBlobData = (data) => {
+  const raw = data.responseData;
+
+  if (!isObject(raw)) {
+    delete data.responseData;
+    return data;
+  }
+  if (typeof raw.body !== 'string') return data;
+  if (!isJson(raw.body)) {
+    delete raw.body;
+    data.isBodyExcluded = true;
+    return {
+      ...data,
+      responseData: JSON.stringify(raw),
+    };
+  }
+  return data;
+};
 
 const resourceAttributes = [
   {
@@ -220,4 +248,5 @@ module.exports = {
   EventType,
   resourceAttributes,
   measureAttributes,
+  stripResponseBlobData,
 };
--- a/packages/aws-lambda-otel-extension/test/unit/external/helper.test.js
+++ b/packages/aws-lambda-otel-extension/test/unit/external/helper.test.js
@@ -1,0 +1,64 @@
+'use strict';
+
+const { expect } = require('chai');
+const { stripResponseBlobData } = require('../../../opt/otel-extension/external/helper');
+
+describe('external helper', () => {
+  describe('stripResponseBlobData', () => {
+    it('should redact entire body if it is not json', () => {
+      const data = {
+        responseData: 'This is just a standard text string',
+      };
+      const out = stripResponseBlobData(data);
+
+      expect('isBodyExcluded' in out).to.equal(false);
+      expect('responseData' in out).to.equal(false);
+    });
+
+    it('should redact body from http response if it is not json', () => {
+      const data = {
+        responseData: {
+          headers: {
+            'Content-Type': 'text/html',
+          },
+          body: 'Some random string...',
+        },
+      };
+      const out = stripResponseBlobData(data);
+
+      expect(out.isBodyExcluded).to.equal(true);
+      expect(out.responseData).to.deep.equal(
+        JSON.stringify({
+          headers: data.responseData.headers,
+        })
+      );
+    });
+
+    it('should keep body from http response if it is json', () => {
+      const data = {
+        responseData: {
+          headers: {
+            'Content-Type': 'text/html',
+          },
+          body: JSON.stringify({ message: 'I am json' }),
+        },
+      };
+      const out = stripResponseBlobData(data);
+
+      expect('isBodyExcluded' in out).to.equal(false);
+      expect(out.responseData).to.deep.equal(data.responseData);
+    });
+
+    it('should keep body if it is json and not an http response', () => {
+      const data = {
+        responseData: {
+          message: 'hello there',
+        },
+      };
+      const out = stripResponseBlobData(data);
+
+      expect('isBodyExcluded' in out).to.equal(false);
+      expect(out.responseData).to.deep.equal(data.responseData);
+    });
+  });
+});
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/external/index.js
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/external/index.js
@@ -9,7 +9,7 @@ const setupLogListenerServer = require('./setup-log-listener-server');
 const initializeTelemetryListener = require('./initialize-telemetry-listener');
 const reportOtelData = require('./report-otel-data');
 const { logMessage, OTEL_SERVER_PORT } = require('../lib/helper');
-const { EventType, SAVE_FILE, SENT_FILE } = require('./helper');
+const { EventType, SAVE_FILE, SENT_FILE, stripResponseBlobData } = require('./helper');
 const { createMetricsPayload, createTracePayload, createLogPayload } = require('./otel-payloads');
 
 function handleShutdown() {
@@ -179,7 +179,8 @@ module.exports = (async function main() {
       for (const responseEvent of Object.values(responseEvents)) {
         try {
           if (!sentResponseEvents.includes(responseEvent.executionId)) {
-            await reportOtelData.requestResponse(responseEvent);
+            // Strip response blob data before sending it to the req/res endpoint
+            await reportOtelData.requestResponse(stripResponseBlobData(responseEvent));
             sentResponseEvents.push(responseEvent.executionId);
           }
         } catch (error) {
--- a/packages/aws-lambda-otel-extension/opt/otel-extension/package.json
+++ b/packages/aws-lambda-otel-extension/opt/otel-extension/package.json
@@ -25,6 +25,7 @@
     "@opentelemetry/sdk-trace-node": "^1.0.1",
     "lodash.clone": "^4.5.0",
     "lodash.get": "^4.4.2",
+    "lodash.isobject": "^3.0.2",
     "node-fetch": "^2.6.7",
     "protobufjs": "^6.11.2",
     "uuid": "^8.3.2"
--- a/packages/aws-lambda-otel-extension/test/unit/external/index.test.js
+++ b/packages/aws-lambda-otel-extension/test/unit/external/index.test.js
@@ -7,12 +7,12 @@ const evilDns = require('evil-dns');
 const log = require('log').get('test');
 const requireUncached = require('ncjsm/require-uncached');
 const overwriteStdoutWrite = require('process-utils/override-stdout-write');
-const getExtensionServerMock = require('../utils/get-extension-server-mock');
-const normalizeOtelAttributes = require('../utils/normalize-otel-attributes');
-const { SAVE_FILE, SENT_FILE } = require('../../opt/otel-extension/external/helper');
+const getExtensionServerMock = require('../../utils/get-extension-server-mock');
+const normalizeOtelAttributes = require('../../utils/normalize-otel-attributes');
+const { SAVE_FILE, SENT_FILE } = require('../../../opt/otel-extension/external/helper');
 const { default: fetch } = require('node-fetch');
-const { OTEL_SERVER_PORT } = require('../../opt/otel-extension/lib/helper');
-const ensureNpmDependencies = require('../../scripts/lib/ensure-npm-dependencies');
+const { OTEL_SERVER_PORT } = require('../../../opt/otel-extension/lib/helper');
+const ensureNpmDependencies = require('../../../scripts/lib/ensure-npm-dependencies');
 
 const port = 9001;
 
@@ -36,7 +36,7 @@ describe('external', () => {
     let stdoutData = '';
     const extensionProcess = overwriteStdoutWrite(
       (data) => (stdoutData += data),
-      async () => requireUncached(() => require('../../opt/otel-extension/external'))
+      async () => requireUncached(() => require('../../../opt/otel-extension/external'))
     );
 
     await new Promise((resolve) => listenerEmitter.once('listener', resolve));


Mmarzex (dismissed):
Looks good to me 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +10,39 @@ const extensionVersion = (() => {
...
+const cleanResponseData = (data) => {
+  let response = '';
+  data.excludedBody = false;
+
+  let responseData = data.responseData;
+  if (typeof responseData === 'string') {
+    try {
+      responseData = JSON.parse(responseData);
+    } catch (error) {
+      responseData = data.responseData;
```
medikoo: I think there's no need for this . if `JSON.parse` fails, `responseData` will remain `data.responseData`

(and for lint purposes we may add `/* noop */` comment)

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +10,39 @@ const extensionVersion = (() => {
...
+    }
+  }
+
+  // Only save responses if it is an object
+  if (responseData && typeof responseData === 'object') {
+    if ('body' in responseData && typeof responseData.body === 'string') {
```
medikoo: It'll be nice to add `lodash` as dependency, and form it simpler as:

```
if (typeof _.get(responseData, 'body') === 'string') {
```

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +10,39 @@ const extensionVersion = (() => {
...
+      try {
+        JSON.parse(responseData.body);
```
medikoo: What's the purpose of this line?

If this logic is meant to be run only if we parsed `responseData` out of JSON. It'll be good to form it more explicitly. In given form, it's hard to get this intention

packages/aws-lambda-otel-extension/test/unit/external/external.test.js
```diff
@@ -7,12 +7,12 @@ const evilDns = require('evil-dns');
 const log = require('log').get('test');
```
medikoo: In this case I'd name it `index.test.js` (to reflect module we're testing)

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
Danwakeem: Ahh nice yeah the `/* noop */` this is exactly what I was looking for. Thanks man 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
Danwakeem: This line is meant to check if the body of an http request is JSON or not.

At this point in the check the `responseData` would look something like this

```json
{
    "traceId": "xxxx",
    "spanId": "xxxxx",
    "executionId": "xxx-xxxx-xxxx-xxxx-xxxxxx",
    "orgUid": "xxxxx",
    "requestData": "",
    "responseData": "{\"statusCode\":200,\"headers\":{\"access-control-allow-origin\":\"*\",\"content-type\":\"application/json; charset=utf-8\"},\"body\":\"{\\\"message\\\":\\\"Success\\\"}}\"}",
    "timestamp": "2022-04-27T19:29:47.151000022Z"
  }
```

And we would want to check that the `body` inside the `responseData` is JSON or not. If it is not JSON we would want to drop the body.

Danwakeem: @medikoo I refactored this a bit to make this more clear. Let me know what you think 👍

packages/aws-lambda-otel-extension/test/unit/external/helper.test.js
```diff
@@ -0,0 +1,65 @@
...
+describe('external helper', () => {
+  describe('cleanResponseData', () => {
+    it('should redact entire body if it is not json', () => {
+      const data = {
+        responseData: 'This is just a standard text string',
+      };
+      const out = cleanResponseData(data);
+
+      expect(out.excludedBody).to.equal(true);
+      expect(out.responseData).to.equal('');
```
medikoo: Wouldn't it be cleaner to simply remove `responseData` property in such case? And in such case we should also not need `exludedBody` (I think this applies only where we exclude part of the response)

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +11,44 @@ const extensionVersion = (() => {
...
+      data.excludedBody = true;
+    }
+  } else if (isJSONResponse(responseData)) {
+    response = JSON.stringify(responseData);
+  } else {
+    data.excludedBody = true;
+  }
+
+  data.responseData = response;
+  return data;
```
medikoo: How about going with approach, where we do not mix _raw_ and parsed out of JSON response under same variable, so it's clear in all places what value is behind the variables, e.g.:


```javascript
const raw = data.responseData;

const json = resolveJsonOrNull(raw);
if (!json) {
  if (typeof raw === 'string') delete data.responsData;
  return data;
}
const jsonHttpResponse = resolveJsonOrNull(json.body);
if (!jsonHttpResponse) return data;
delete json.body;
return Object.assign(data, {
  excludedBody: true,
  responseData: JSON.stringify(json)
});
```

medikoo (commented):
@Danwakeem I've added some comments, but maybe ignore them for now, and let's clarify first what exactly we're addressing here, as it seems as not perfectly clear.

I understand that we want to strip some special cases of responses from data we don't want to propagate to the ingestion server.

Having that can we specify what exactly responses we want to strip.

e.g.HTTP response would can be object as follow:

```javascript
{ statusCode: 200, body: "{\"message\":\"ok\"}" }
```

I understand that we deal with direct lambda results, as returned by lambda handler, so is there a valid case of returning a HTTP response as a JSON string in entity as follows:

```javascript
"{\"statusCode\":200,\"body\":\"{\\\"message\\\":\\\"ok\\\"}\"}"
```

It'll be also good that this utility documents in what exactly cases it addresses and strips.

Danwakeem:
> I understand that we deal with direct lambda results, as returned by lambda handler, so is there a valid case of returning a HTTP response as a JSON string in entity as follows:

```
"{\"statusCode\":200,\"body\":\"{\\\"message\\\":\\\"ok\\\"}\"}"
```

@medikoo yes. When you return a lambda response with this format you typically have to JSON stringify the body of your response. So it looks like something like this.

```javascript
{
  statusCode: 200,
  body: JSON.stringify({
    message: 'ok',
  }),
}
```

I am not sure this `JSON.stringify` step is required or not so I figured it is simple enough to handle cases where it is stringified or not 👍 

Since we have to follow this format this means what we end up parsing out is in that double JSON stringified format. Here is what these look like in our opensearch db 👇 

Screen Shot 2022-04-28 at 8 02 20 AM

Danwakeem:
@medikoo I went ahead and updated this to the format you had suggested. Let me know what you think

medikoo:
@Danwakeem I understand then that we want to trim responses for the following cases:

1. Unclear case, where response are plain string objects
2. HTTP API case, where the response is an object that contains _string_ body property.

Is that correct? Do I miss anything else?

Danwakeem:
@medikoo Yes that is all correct 👍 For the second one I would probably say something like 

HTTP API case, where the response is an object that contains string body property _that is no a JSON string_.

But I am kind of assuming this is what you mean.

medikoo:
@Danwakeem can we add documentation on that? I think it's pretty important that's part of public documentation.

> HTTP API case, where the response is an object that contains string body property that is no a JSON string.

Do you mean that we intend to strip following:

```javascript
{ statusCode: 200, body: 'foo' /* non-json string */ }
```

but not one as as:

```javascript
{ statusCode: 200, body: '{"message":"ok"}' /* JSON string */ }
```

?

Danwakeem:
@medikoo Exactly 👍

medikoo (commented):
> @medikoo Exactly 👍

Ok, so whole intention is to strip eventual blob data, and nothing else. It's clearer for me now.


I've realized ow, that we do not have yet any docs on produced payload, so it's hard to suddenly jump out with documentation on that.

Still, let's maybe improve the name of the util from `cleanResponseData` to `stripResponseBlobData`.

Danwakeem:
@medikoo I went ahead and changed the name of the function. I also went ahead and added a section in the README that mentions something about stripping this kind of data from the response.

I think it is helpful to have it in there but If you would rather me remove that until we have payloads documented just let me know 🤷

packages/aws-lambda-otel-extension/README.md
```diff
@@ -62,6 +62,34 @@ Configure following environment variables with designated server urls:
...
+```
+"Lambda response"
+true
+12
+```
+
+_HTTP Responses with a non JSON body will be ignored_
+
+```json
+{ "statusCode": 200, "body": "lambda response" } -> { "statusCode": 200 }
```
medikoo: Comments are not valid in JSON, so maybe let's declare it as _javascript_ and make a valid JS comment

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +10,43 @@ const extensionVersion = (() => {
...
+    if (!jsonHttpResponse) {
+      delete json.body;
+      return Object.assign(data, {
+        excludedBody: true,
+        responseData: JSON.stringify(json),
+      });
+    }
+  }
+  return data;
+};
```
medikoo: After having full understanding I think we can simplify it further. What do you think about:

```javascript
if (!_.isObject(raw)) {
  delete data.responseData;
  return data;
}
if (typeof raw.body !== 'string') return data;
if (!isJsonString(raw.body) {
  delete raw.body
  data.isBodyExcluded = true;
}
return data;
```

I've also proposed above to rename `excludedBody` to `isBodyExcluded` to ensure meaning is clear (`excludedBody` suggests it's a property where we host an excluded response body).

Also maybe `isBodyStripped` will be better?

medikoo (commented):
>  I think it is helpful to have it in there but If you would rather me remove that until we have payloads documented just let me know 🤷

If we have it, that's great, it looks very good to me in a way you've proposed 👍 

I think we're getting closer to having this merged, still I think we still have a room for simplification, please see my comment :)

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
Danwakeem: I think I like `isBodyExcluded` of the two options 👍

medikoo: Yeah, `isBodyStripped` doesn't feel as a great choice 🤦 🤣

packages/aws-lambda-otel-extension/README.md
Danwakeem: I moved this to a table format where we have an input and and an output. Let me know what you think about this approach 🙏

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
Danwakeem: 😆 This should be updated 😎

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +11,37 @@ const extensionVersion = (() => {
...
+    const res = JSON.parse(data);
+    return res;
+  } catch (error) {
+    return null;
+  }
+};
+const isJson = (data) => !!resolveJson(data);
+
+const stripResponseBlobData = (data) => {
+  data.isBodyExcluded = false;
```
medikoo: Is there a need to have this property set, if we do not exclude the body?

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +11,37 @@ const extensionVersion = (() => {
...
+  const raw = resolveJson(data.responseData);
```
medikoo: Why do we attempt to resolve JSON out of the eventual top level string? This doesn't serve any use case we've discussed (for HTTP API case, it's objects with JSON string at body property that are returned and not JSON strings - do I miss something?)

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
Danwakeem: Just so the payload going to our open search cluster is consistent.

medikoo: hm.. but we already have a problem with bloated payload. Ideally, if we send only properties that add some information, and in general backend should not expect exactly same set of properties every time.

Is this a problem for a backend to not have this property?

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
Danwakeem: For http app api data it is a JSON string so we need to parse the string before we can inspect the body property. This body property is also stringified. 

If you take a look at the tests that are attached to this PR it describes exactly what the payloads look like for each use case.

medikoo: Hmm.. in my understanding HTTP API response as send from lambda is in this form:

```javascript
{ statusCode: 200, body: '{"message":"ok"}' }
```
And not this:

```javascript
"{\"statusCode\":200,\"body\":\"{\\\"message\\\":\\\"ok\\\"}\"}"
```

Or is returning JSON string from  lambda, is also valid in that case, an it'll be parsed out by AWS (?)

Danwakeem: By the time it gets to our function the payload has been stringified at some point either by an otel library or by us but when I am printing the response payload from this function I am getting the following 👇 

```
"{\"statusCode\":200,\"body\":\"{\\\"message\\\":\\\"ok\\\"}\"}"
```

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
Danwakeem: We have a bloated payload problem on traces but that does not apply to req/res data. Also I would rather be explicit here than not given how small this data is.

Danwakeem: I will talk to max and the gang about the idea of removing this 👍

medikoo: Sure, I believe our rule of thumb should be, to keep payloads minimal and only introduce properties that have some extra information to describe.

In this case `isBodyExcluded: false` doesn't add anything. Our interpretation of the data will be the same wether this property is set to `false` or is not set at all.

Danwakeem: This is true 🤷 I will see what the ingest gang thinks about excluding properties in this payload.

Danwakeem: Double checked with ingest and it should be fine if we exclude this property so I went ahead and removed it 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
medikoo: Hmmm.. as I check it doesn't seem to be the case.

1. Otel instrumentation provides us with  Lambda response as-is (so object in case of HTTP API) ->  https://github.com/serverless/runtime/blob/4e329cd5471e878efa30283e7fa44bee874c304e/packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js#L387
2. We're passing this as-is to external extension -> https://github.com/serverless/runtime/blob/4e329cd5471e878efa30283e7fa44bee874c304e/packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js#L262
3. ofc for transport whole request payload is serialized -> https://github.com/serverless/runtime/blob/4e329cd5471e878efa30283e7fa44bee874c304e/packages/aws-lambda-otel-extension/opt/otel-extension/internal/index.js#L380 and unserialized in external extension -> https://github.com/serverless/runtime/blob/4e329cd5471e878efa30283e7fa44bee874c304e/packages/aws-lambda-otel-extension/opt/otel-extension/external/initialize-telemetry-listener.js#L25

Then this data is never seralized by us in context of external extension, so what lands in `stripResponseBlobData` is lambda response in exact shape as it was returned by lambda.

I've also doubled confirmed that, by using an extension with extra logs added and deploying this service: https://github.com/medikoo/test-serverless/tree/test-http-api-console-response

And it's an object as follows that ends as `responseData` in external extension at point where `stripResponseBlobData` would be run

```javascript
{ statusCode: 200, body: '{"message":"ok"}' }
```

Danwakeem: Ahh okay cool I must be getting confused with how ingestion works 👍

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +11,34 @@ const extensionVersion = (() => {
   return require('../version');
 })();
 
+const isJson = (data) => {
+  try {
+    const res = JSON.parse(data);
+    return res;
```
medikoo: I think we don't need `res` to be returned. We just need to return `true`

packages/aws-lambda-otel-extension/opt/otel-extension/external/helper.js
```diff
@@ -10,6 +11,34 @@ const extensionVersion = (() => {
...
+  } catch (error) {
+    return null;
+  }
+};
+const stripResponseBlobData = (data) => {
+  const raw = data.responseData;
+
+  if (!isObject(raw)) {
+    delete data.responseData;
+    data.isBodyExcluded = true;
```
medikoo: In this case we do not return `responseData` at all, so I think setting `isBodyExcluded` is not needed (?)

medikoo (commented):
Looks great @Danwakeem ! Can we rebase it into one commit?

Danwakeem:
@medikoo Yeah no problem. Just rebased and left it as the same commit message as the first `feat` one. If you think I should change the message just let me know 😎

medikoo (dismissed):
Looks great @Danwakeem 

I'd probably name this as: `Avoid sending binary response data`

medikoo:
As technically data in all cases is JSON (plain string is also a JSON valid value)

Danwakeem:
@medikoo Yeah I like that description better 👍 I went ahead and updated this

medikoo (approved):
👍

PR #46, opened 2022-05-04
Fix handling of initialization errors

medikoo:
I've discovered that initialization errors, when thrown from pre-required preparation module are exposed as Runtime errors to users.

This patch ensures the error is cached and thrown from patched handler instead. Having that user faces same effect, no matter whether lambda function is instrumented or not


Danwakeem (approved):
Seems straightforward 👍 Good find!

PR #47, opened 2022-05-05
Nest Node.js projects/packages in `node` folder

medikoo:
It's to create room for _python_ specific utilities

PR #48, opened 2022-05-05
Fix internal wrapper

medikoo:
An accidental rename error (copy/paste to blame) sneaked with #47, luckily exposed by the integration test.

Fix that, and improve error handling in tests


Danwakeem (approved):
Ahh yeah looking back at #46 I see where this happened 🤦 Sorry I didn't see this either

PR #49, opened 2022-05-11
Simplify and generalize external extension

medikoo:
Rewrite of external extension logic, maintaining the current contract for communication with the ingestion server, which as I deducted is as follows:

- `requestEventPayload` is sent immediately to ingestion server upon receiving from internal extension
- Function logs are sent immediately on reception, assuming we already have `eventData` as sent from the internal extension. If we don't have `eventData` logs are stored in a queue and sent immediately on reception of `eventData`
- `responseEventPayload`  is sent immediately to ingestion server upon receiving from internal extension
- Metrics that are read from `telemetryData` are sent immediately on reception from the internal extension, and they're sent a second time with `platform.report` once it's obtained
- Traces that are read from `telemetryData` are sent immediately to reception
- `platform.report` is sent immediately on reception (together with metrics data)

In case of invocation failure (as a timeout) `telemetryData` is stored to a temporary file, so on reinitialization, we can send Metrics together with `platform.report`

I've tested it for basic invocations cases and timeout error, for last I've configured an additional integration test.

This PR cuts ca 50% of the codebase of external extension logic, also as I observed it reduces by 150-200ms duration of each lambda invocation


Danwakeem (approved):
This is looking great man! 🎸 ✅ 

I tested this with quite a few different app configurations and everything seems to be working as expected. I am seeing errors, timeouts, successful invocations, log streaming, etc.

Danwakeem:
Well actually sorry give me a few more minutes I think I need to check one more thing 👍

Danwakeem (approved):
Lol never mind this looks good 😆 I thought I was seeing dropped durations but it was just a delay before it was joined with the trace details 👍

PR #50, opened 2022-05-20
Split extension and configure runtime agnostic build script

medikoo:
- Ensure internal and external extensions logic are independent of each other (remove any common code)
- Separate both extensions at the root folder (that also introduces a breaking change, therefore the following release will be `0.3`)
- Configure a script that builds a runtime agnostic version of the external extension (cc @whardier)
- Remove `lodash.isobject` dependency

_There are no changes to any internal logic in this PR_


Danwakeem (approved):
This looks good but more to Shane's point the final external extension zip is pretty large 🤷

medikoo:
> This looks good but more to Shane's point the final external extension zip is pretty large 🤷

Yes, as we discussed it's a temporary step until we have it written in Rust

PR #51, opened 2022-05-23
Remove `node-fetch` dependency in external extension

medikoo:
So it's lighter and has better initialization time (less external requires), I've tested it by deploying and confirming that all requests work as expected (as this is not really covered by integration tests at this point)

Additionally improved CI setup:
- So far we were updating `node_modules` with `npm update` which won't clean dependencies that were removed from `package.json`. For internal and external extensions I've ensured we have a clean install
- Remove obsolete installation of dependencies in test fixtures for publish job


Danwakeem (approved):
Awesome 😎

PR #54, opened 2022-05-25
Fix dependencies configuration

medikoo:
_Addressing issue reported internally by @whardier (observed `Error: Cannot find module 'yargs-parser'`)_

Current conventions are the following:
- All _dev_ dependencies that are used for testing, linting etc. (operations that should not be done on the package when it's npm installed) at the root package
- All `aws-lambda-otel-extension` build script dependencies are treated as _prod_ (not _dev_) dependencies (as users should be able to build the extension binaries after installing this package).

This PR fixes issues in dependencies layout

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -25,6 +25,7 @@
     "chai": "^4.3.6",
     "chalk": "^4.1.2",
     "child-process-ext": "^2.1.1",
+    "deasync": "^0.1.26",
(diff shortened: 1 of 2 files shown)


Danwakeem (approved):
Ahh sorry I forgot to hit approve on this yesterday 🙄 ✅

PR #55, opened 2022-05-26
Bundle internal and external extensions into a single file

medikoo:
Bundle extensions but without minifying the source code, so eventual crashes can easily be debugged (I've also tested that minification doesn't give us a real gain in any area)

As I tested this cuts down ca 260ms from initialization time (30ms on external extension and 230ns on internal)

Additionally, it cuts down extension size from 8.7MB to 284KB


Danwakeem (approved):
This is awesome 😎 Pretty amazing what moving some files around can do

PR #56, opened 2022-05-26
 Improve extension overhead logs tag line

medikoo:
`Extension duration` suggested that it's a total time that extension took, while it reports only the overhead of what extension adds to the execution time. Updated log message to reflect that


Danwakeem (approved):
😎

PR #57, opened 2022-05-26
Support new user settings format

medikoo:
Switch to the format of settings we've agreed on internally.

This feels like a breaking change, but as we considered those settings experimental we do not propagate this as breaking at this point


Danwakeem (dismissed):
Looks good to me 👍 And framework will just drop these into the user settings environment variable correct?

medikoo:
> Looks good to me 👍 And framework will just drop these into the user settings environment variable correct?

The framework also needs the update, and it's ready here: https://github.com/serverless/serverless/pull/11115

PR #58, opened 2022-05-26
Fix reliability of integration tests

medikoo:
Addresses two recent fails:

- https://github.com/serverless/runtime/actions/runs/2391380978
- https://github.com/serverless/runtime/actions/runs/2391391575

Fails happened because we started to process reports when just _metrics_ report was read from S3 bucket. 

This patch ensures we do not move to processing until there's also _traces_ and final _metrics_ report


Danwakeem (approved):
Oh nice 👍 I guess sometimes it was producing all the reports fast enough and sometimes it wasn't?

PR #62, opened 2022-06-01
Generalize user settings and request handling

medikoo:
- Unify reports handling to one central function
- Move all the settings to user settings JSON read from `SLS_OTEL_USER_SETTINGS` environment variable


Danwakeem (approved):
LGTM 👍 The combined method for sending reports is nice 😎

PR #63, opened 2022-06-02
Fix request and response hooks handling

medikoo:
_Reported internally by @Mmarzex_

It appears that AWS Lambda instrumentation supports only _sync_ request and response hooks, and eventual promises as returned by them are not awaited

It was not that problematic for lamdbas which resolved with a _callback_ and over there `context.callbackWaitsForEmptyEventLoop` is set to `true`, so AWS naturally waited for that hook to resolve before settling the invocation. Still when result of lambda is _thenable_ this setting is set to _false_ and AWS immediately settles the invocation.

With this PR I patched handling of responses so in all cases we wait for hooks to resolve. Additionally I've ensured we handle the right way the edge case of callback/thenable race (where given lambda is attempted to be resolved with both callback and thenable).

I've also added as simple test script that allows to invoke internal extension locally (for debugging purposes)

### DIFF
--- a/node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
+++ b/node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
@@ -44,6 +44,88 @@ const eventData = {};
 let timeoutHandler;
 let transactionCount = 0;
 
+const requestHandler = async (span, { event, context }) => {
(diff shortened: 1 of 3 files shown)


Danwakeem (approved):
Appears to work with callbacks now 👍 I am kind of surprised we didn't see this earlier 🤷‍♂️

PR #65, opened 2022-06-06
`serverless/console` content

medikoo:
Content of https://github.com/serverless/console (as a first step of merging two repositories)


hev:
LGTM. I confirmed that full history appears to be there, and my latest commits are there also. No concerns merging this from my end.

PR #66, opened 2022-06-06
Unify debug log handling

medikoo:
In context of extension, no `console.log`'s are prefixed with timestamp and request id, also `console.log` may interfere with `stdout` of the lambda logic which we may confirm on in tests. Therefore changed `debugLog` so it writes to `stderr` via `_rawDebug` (this method is not susceptible to some third party patches).

Having that I've reconfigured current logs which we wrote with `_rawDebug` to use `debugLog` function


Danwakeem (approved):
Looking good 👍

PR #67, opened 2022-06-09
Test reorganization and improvements (pave path towards benchmarks)

medikoo:
- Improve function fixture names, so they follow `<outcome>-<resolution-type>-<case>` naming
- Ensure logged report end as single item in CloudWatch logs (running `process.stdout.write` multiple times, resulted in multiple items in some cases)
- Read reports from CloudWatch logs instead of S3 (as writing them to S3 in extension context affects visibly extension performance)
- Improve generated resource basenames. So far it was random string for every test run, which could lead to pollution of account with stranded resources if tests is run multiple times locally without proper cleanup. With this change only in CI environment resource basenames will stay random, but locally local machine is used (or basename token is read form `TEST_UID` environment variable), thanks to that when playing locally we'd rather recreate/update left over resources instead of creating new ones.
- Ensure test handle gentle scenarios if given resources already exist (it's important after introduction of persistent `TEST_UID`)
- Seclude standalone `cleanup` script, through which we can remove all eventually left over test resources
- Confirm more precisely on generated reports (ensure that given reports are provided in scope of expected invocation frame - this covers issue that was fixed with #63 )
- Improve coverage of debug logs in external and internal extension
- Cleanup resources also on test fails (it appears we will still have access to deleted functions CloudWatch logs, and that should be good enough for eventual debugging)
- Improve code organization by secluding logical parts in separate modules
- Process all tested functions simultaneously (that should improve test run time and allow to keep it at similar level as we will add more test cases)


Danwakeem (approved):
🎸

PR #68, opened 2022-06-10
Init AWS IAM role CFN template

ole3021:
Init the AWS Iam role CFN template for today and future usage. The template are made by consider:

* use Yaml for easy add comments in the yaml file.
* Grant reads related permissions for AWS services we use today and tomorrow.
* Remove the Tag related permissions.
* Limited the Read permissions on AWS S3 to avoid abusing the permission to read user files.
* Add Specific permissions about AWS Lambda that used in Integration services.

[Jira Task](https://serverlessinc.atlassian.net/browse/SC-85)

### DIFF
--- a/instrumentation/aws/iam-role-cfn-template.yaml
+++ b/instrumentation/aws/iam-role-cfn-template.yaml
@@ -1,5 +1,24 @@
+#
+# Serverless Inc.'s Observability AWS IAM Role CloudFormation Template
+#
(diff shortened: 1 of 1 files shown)


austencollins (approved):
This looks like a great step toward improving the security and especially the user-friendliness and transparency of our IAM Role.  We'll continue to improve on this 👍

PR #69, opened 2022-06-13
Fix ESM support for `nodejs16.x` runtime

medikoo:
AWS apparently uses a different internal module loader in v14 than in v16, and our wrapper patch does not work for ESM in the context of node16.x runtime.

This PR fixes that and ensures we run tests against all supported runtimes.

Additionally:
- Ensure to handle well scenarios where the report is split among multiple CloudWatch logs (observed such situation locally)
- Improve debug logs for lambda invocation
- Fix express app test setup (after recent refactor, it didn't receive API Gateway payload - ensure that also test crashes if we have regression on that - we just tested that express instrumentation loaded and not that it generated spans specific to request, it was improved)
- Fix time range configuration for CW logs retrieval (by mistake, we've used start time recorded before the last invocation and not before the first one - I've observed one test locally due to that - otherwise, it usually worked as start time was intentionally subtracted by additional 3s)
- Handle eventual race conditions on AWS side in the context of `Lambda.invoke` (in response to failure observed locally)
- Refactor tests setup, so different testing scenarios can be set for each handler (so far, it was one scenario per handler)


Danwakeem (approved):
LGTM 👍 The map will make it easy to add/remove support for each version of node. Nice work

PR #72, opened 2022-06-14
fix: Updated timeout error message

Danwakeem:
## Description
The error message is a bit vague in that it sounds like the timeout has not happened yet even though it most likely has happened. So we decided to update the message to be a bit more definite given this error is showing up in the console UI.

### DIFF
--- a/node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
+++ b/node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
@@ -201,7 +201,7 @@ const responseHandler = async (span, { res, err }, isTimeout) => {
     functionData.errorExceptionType = 'TimeoutError';
     functionData.errorExceptionMessage = '';
     functionData.errorExceptionStacktrace =
-      'Function execution is going to exceeded configured timeout limit.';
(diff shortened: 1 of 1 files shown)


Danwakeem:
@medikoo this is such a small change I am going to go ahead and merge this in 👍

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
```diff
@@ -201,7 +201,7 @@ const responseHandler = async (span, { res, err }, isTimeout) => {
     functionData.errorExceptionType = 'TimeoutError';
     functionData.errorExceptionMessage = '';
     functionData.errorExceptionStacktrace =
-      'Function execution is going to exceeded configured timeout limit.';
+      'Function execution exceeded the configured timeout limit.';
```
medikoo, line 204: To be perfectly clear, we don't know that yet, at this point. It's still possible that the function will manage to succeed on time.

I think that the right approach here is to not report `errorExceptionStacktrace` for timeouts, as it's clear we can't have a stack trace here, so this data is N/A

Also, currently our timeout reporting is unreliable, I've described the issue in more detail in `SC-263`

PR #73, opened 2022-06-14
chore: Release v0.4.2

Danwakeem:
## Description
Releasing the update to the timeout message

### DIFF
--- a/node/packages/aws-lambda-otel-extension/CHANGELOG.md
+++ b/node/packages/aws-lambda-otel-extension/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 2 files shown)


medikoo (commented):
@Danwakeem please see https://github.com/serverless/console/pull/72#discussion_r897652966

I think we should rather not include `errorExceptionStacktrace` in case of a timeout

Danwakeem:
@medikoo i agree but we do end up using this field on the frontend to display the timeout error message. I think switching the error message and error stack trace fields is another issue all together.

I could work on that bug today and include that in this release 👍

medikoo:
> @medikoo i agree but we do end up using this field on the frontend to display the timeout error message. I think switching the error message and error stack trace fields is another issue all together.


I think on a frontend logic should be as:

```javascript
if (errorType === 'timeout') {
  show("Lambda timed out")
} else {
  show(errorMesssage):
  if (errorStackTrace) show(errorStackTrace)
}
```

Danwakeem:
@medikoo sure we could do that. I will just close this out.

PR #74, opened 2022-06-15
fix: Updated error field mapping and names

Danwakeem:
## Description
Adjusted the error message field mapping and names


node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
```diff
@@ -198,26 +198,23 @@ const responseHandler = async (span, { res, err }, isTimeout) => {
...
-    functionData.errorExceptionStacktrace =
-      'Function execution exceeded the configured timeout limit.';
+    functionData.errorType = 'TimeoutError';
   } else if (err) {
     functionData.error = true;
     functionData.errorCulprit = err.message;
-    functionData.errorExceptionType = typeof err;
-    functionData.errorExceptionMessage = err.message;
-    functionData.errorExceptionStacktrace = err.stack;
+    functionData.errorType = typeof err;
```
medikoo: I think it doesn't make much sense to use `typeof err` as there's no value in it

I think what we want to know here, is whether it was: _timeout_, _unhandled_, _handled_ or _initialization_ error (and this case we speak of _handled_ case)

Danwakeem: Okay so in this case we should maybe put unhandled?

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
```diff
@@ -198,26 +198,23 @@ const responseHandler = async (span, { res, err }, isTimeout) => {
...
     pathData['http.status_code'] >= 500 &&
     ['aws.apigateway.http', 'aws.apigatewayv2.http'].includes(functionData.eventType)
   ) {
     // This happens if we get a 500 status code set explicity within in the app
     functionData.error = true;
     functionData.errorCulprit = 'internal server error';
-    functionData.errorExceptionType = typeof new Error();
-    functionData.errorExceptionMessage = 'internal server error';
-    functionData.errorExceptionStacktrace = 'internal server error';
+    functionData.errorType = typeof new Error();
```
Danwakeem: And then here maybe out “handled”?

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
medikoo: It's actually _handled_ here. Unhandled is when the original handler doesn't report the error back, but crashes dirty way and kills lambda instance (those can only be tracked from external extension)

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
```diff
@@ -198,26 +198,23 @@ const responseHandler = async (span, { res, err }, isTimeout) => {
...
-    functionData.errorExceptionStacktrace =
-      'Function execution exceeded the configured timeout limit.';
+    functionData.errorType = 'TimeoutError';
   } else if (err) {
     functionData.error = true;
     functionData.errorCulprit = err.message;
-    functionData.errorExceptionType = typeof err;
-    functionData.errorExceptionMessage = err.message;
-    functionData.errorExceptionStacktrace = err.stack;
+    functionData.errorType = 'unhandled';
```
Danwakeem:
```suggestion
    functionData.errorType = 'handled';
```

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
medikoo: I think both of them are handled.

What's the real difference? Is that one is response for API Gateway? and other is not?

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
Danwakeem: Got it so really in this file the error should always be marked as handled 👍

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
medikoo: Ok, I looked closer. 

Second is when we report `500` with success resolution.

I guess it's a situation where the user app might have crashed and HTTP decorator converted it to `500`which is reported with _success_ response (so not really an error).

That's an interesting case, as technically, just by looking at lambda response, it's a _success_  resolution.

Danwakeem: In this if block the user is returning a `500` error response from their API.

The if statement above is one where you would throw an error from an async function or pass something like `callback(new Error(), null)` from your lambda function

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
```diff
@@ -198,26 +198,23 @@ const responseHandler = async (span, { res, err }, isTimeout) => {
...
-    functionData.errorExceptionStacktrace =
-      'Function execution exceeded the configured timeout limit.';
+    functionData.errorType = 'TimeoutError';
   } else if (err) {
     functionData.error = true;
     functionData.errorCulprit = err.message;
-    functionData.errorExceptionType = typeof err;
-    functionData.errorExceptionMessage = err.message;
-    functionData.errorExceptionStacktrace = err.stack;
+    functionData.errorType = 'handled';
```
medikoo, line 205: I think it'll be nice to have unified naming convention for _types_.

This PR proposes two types: `TimeoutError` and `handled` that doesn't look good.

What would be nice is `timeout` and `handled` but can we do that?

Danwakeem: Yeah that should be alright 👍 We are not really taking advantage of the `errorType` in our ingest/frontend app right now so we can just look for `TimeoutError` or `timeout`

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
Danwakeem: @medikoo thinking about this a bit more it might actually make sense for this to be marked as `unhandled`. This block of code could be executed in the case that an error is thrown from the lambda function that was not caught by the customers code 🤔

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
Danwakeem: I went ahead and updated the timeout `errorType` to `timeout` 😎

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
medikoo: > This block of code could be executed in the case that an error is thrown from the lambda function that was not caught by the customers code 🤔

But that'll mean that all errors reported from lambda _unhandled_ while we need to differentiate from those returned from Lambda and those which crash lambda.

So we need type name for the following:
- Timeout
- Lambda crashed during invocation
- Lambda crashed during initialization
- Lambda invocation resolved with an error.

How would name those 4 types? Note here, we're dealing with the last one. And Lambda crashes by AWS are reported as _Unhandled exceptions_

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
```diff
@@ -198,26 +198,23 @@ const responseHandler = async (span, { res, err }, isTimeout) => {
   if (isTimeout) {
     functionData.error = true;
     functionData.errorCulprit = 'timeout';
-    functionData.errorExceptionType = 'TimeoutError';
-    functionData.errorExceptionMessage = '';
-    functionData.errorExceptionStacktrace =
-      'Function execution exceeded the configured timeout limit.';
+    functionData.errorType = 'timeout';
```
medikoo, line 201: Great, I think it's way better name, still we need to also adjust integration test :)

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
Danwakeem: 🤔 Maybe we should have error types named basically like you outlined above

- `timeout`: Set when we have a timeout error
- `unhandled_invocation`: Lambda crashed during invocation
- `unhandled_initialization`: Lambda crashed during initialization
- `unhandled_resolved`: Lambda invocation resolved with an error
- `handled`: Set if we receive a handled error (Ex. we see a response with a 500 error)

They are kind of verbose but I think they accurately describe what is happening 🤔

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
Danwakeem: Ahh thanks! Updated the test ✅

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
Danwakeem: @medikoo yeah exactly 👍 For context, this was a requirement that successful lambda invocations that return a 500 error as an error so they would show up as an error in console.

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
medikoo: > `HandledError`: Set if we receive a handled error (Ex. we see a response with a 500 error)

Problem is that, in this case, we didn't receive an error. We read from the success lambda response that `500` is reported to the API GW integration.

And `UnhandledResolvedError`, technically in all cases is _handled_ as it was propagated through _error_ channel and not thrown in the air (_unhandled_ - again, it's how AWS calls errors which crash lambdas)

Clean (low-level) I think would be to not report the last one as _error_, but at a high level (in console) introduce special handling in for `500` statuses, so they're shown to users as errors. Still, I get it probably demands a bigger discussion, so I wouldn't mess with it now

medikoo (approved):
👍

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
Danwakeem: Yeah exactly it was just a requirement from console that was why we ended up surfacing successful lambda invocations that return a 500 error as an error ✅

medikoo: I think we will have to revisit it at some point.

As ideally if lambdas can be instrumented only with external extension, and over there (not knowing the shape of the response), such case will be reported as _success_.

And for data coherence ideally if internal extension doesn't change the report as produced by the external but enriches instead (so internal could add info on HTTP statuses, and basing on that we show extra alarms to users)

Danwakeem: Yeah makes sense given how the extension is evolving 👍

PR #80, opened 2022-06-21
feat: Add attributes and resource to req/res

Danwakeem:
## Description
We are hoping to add req/res data to the dev mode time line in serverless console but right now the req/res payloads do not contain enough information to surface/filter these in the dev mode console. This PR adds the attribute/resource data to the req/res payloads.

### DIFF
--- a/node/packages/aws-lambda-otel-extension/external/otel-extension-external/index.js
+++ b/node/packages/aws-lambda-otel-extension/external/otel-extension-external/index.js
@@ -39,7 +39,7 @@ module.exports = (async () => {
 
   // Rotate current request data
   // Each new "platform.start" or invoke event resets the object
-  const getCurrentRequestData = (() => {
+  const getCurrentRequestContext = (() => {
     let current;
     return (uniqueEventName) => {
       if (!current) current = { logsQueue: [] };
@@ -88,13 +88,13 @@ module.exports = (async () => {
               );
             }
             if (functionLogEvents.length) {
-              const currentRequestData = getCurrentRequestData();
-              if (!currentRequestData.eventData) {
-                currentRequestData.logsQueue.push(...functionLogEvents);
+              const currentRequestContext = getCurrentRequestContext();
+              if (!currentRequestContext.requestData) {
+                currentRequestContext.logsQueue.push(...functionLogEvents);
               } else {
                 sendReport(
                   'logs',
-                  createLogPayload(currentRequestData.eventData, functionLogEvents)
+                  createLogPayload(currentRequestContext.requestData, functionLogEvents)
                 );
               }
             }
@@ -103,7 +103,7 @@ module.exports = (async () => {
               switch (event.type) {
                 case 'platform.start':
                   debugLog('Extension platform log: start');
-                  getCurrentRequestData('start');
+                  getCurrentRequestContext('start');
                   // eslint-disable-next-line no-loop-func
                   ongoingInvocationDeferred = new Promise((resolve) => {
                     resolveOngoingInvocationDeferred = resolve;
@@ -243,7 +243,7 @@ module.exports = (async () => {
           await Promise.resolve(ongoingInvocationDeferred).then(waitUntilAllReportsAreSent);
           break;
         case 'INVOKE':
-          getCurrentRequestData('invoke');
+          getCurrentRequestContext('invoke');
           await new Promise((resolve) => {
             runtimeEventEmitter.once('runtimeDone', resolve);
           })
@@ -276,20 +276,23 @@ module.exports = (async () => {
                   if (data.record.requestEventPayload) {
                     sendReport('request', createRequestPayload(data.record.requestEventPayload));
                   }
-                  const currentRequestData = getCurrentRequestData('request');
-                  currentRequestData.eventData = data.record;
-                  if (currentRequestData.logsQueue.length) {
-                    sendReport('logs', createLogPayload(data.record, currentRequestData.logsQueue));
+                  const currentRequestContext = getCurrentRequestContext('request');
+                  currentRequestContext.requestData = data.record;
+                  if (currentRequestContext.logsQueue.length) {
+                    sendReport(
+                      'logs',
+                      createLogPayload(data.record, currentRequestContext.logsQueue)
+                    );
                   }
                 }
                 break;
               case 'telemetryData':
                 lastTelemetryData = data;
                 if (data.record.responseEventPayload) {
-                  const { eventData: currentRequestData } = getCurrentRequestData();
+                  const { requestData } = getCurrentRequestContext();
                   sendReport(
                     'response',
-                    createResponsePayload(data.record.responseEventPayload, currentRequestData)
+                    createResponsePayload(data.record.responseEventPayload, requestData)
                   );
                 }
                 sendReport('metrics', createMetricsPayload(data.requestId, data.record.function));
--- a/node/packages/aws-lambda-otel-extension/external/otel-extension-external/index.js
+++ b/node/packages/aws-lambda-otel-extension/external/otel-extension-external/index.js
@@ -3,10 +3,8 @@
 'use strict';
 
 const processStartTime = process.hrtime.bigint();
-const processStartDate = new Date().getTime();
 let isInitializing = true;
 let invocationStartTime;
-let invocationStartDate;
 let shutdownStartTime;
 
 module.exports = (async () => {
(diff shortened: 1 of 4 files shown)


node/packages/aws-lambda-otel-extension/external/otel-extension-external/index.js
```diff
@@ -268,7 +273,15 @@ module.exports = (async () => {
               case 'eventData':
                 {
                   if (data.record.requestEventPayload) {
-                    sendReport('request', data.record.requestEventPayload);
+                    const { resourceAtt, metricsAtt } = createAttributes(data.record);
+                    sendReport('request', {
+                      ...data.record.requestEventPayload,
+                      timestamp: currentRequestData.logsQueue[0]
+                        ? currentRequestData.logsQueue[0].time
+                        : new Date().getTime() - 100,
```
Danwakeem, line 281: Looking for a way to replace this with the invocation time 🤔

Danwakeem: Basically I need to make sure the request payload has a timestamp at invocation time (Before any other log timestamps). The request doesn't come with a timestamp and I can't seem to find something reliable on external that will give me a timestamp before any log messages 🤔

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/wrapper.js
```diff
@@ -18,6 +18,7 @@ const wrapHandler = (handlerFunction) => {
   let requestStartTime;
   let responseStartTime;
   let currentInvocationId = 0;
+  global.invokeStartDate = new Date().getTime();
```
Danwakeem: I am not a fan of this. @medikoo Do you know of a more reliable way to get the invocation time from the external part of the extension? I need to get an invocation time for the request that has a timestamp before the first log statement 🤔

node/packages/aws-lambda-otel-extension/external/otel-extension-external/index.js
```diff
@@ -268,7 +273,12 @@ module.exports = (async () => {
               case 'eventData':
                 {
                   if (data.record.requestEventPayload) {
-                    sendReport('request', data.record.requestEventPayload);
+                    const { resourceAtt, metricsAtt } = createAttributes(data.record);
+                    sendReport('request', {
+                      ...data.record.requestEventPayload,
+                      attributes: resourceAtt,
+                      resource: metricsAtt,
+                    });
```
medikoo: Can we create `createRequestPayload` and `createResponsePayload` which will cover payload generation in whole (same as we do for _metrics_, _traces_ etc.)?

Leaking into this file payload generation doesn't feel right

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/wrapper.js
medikoo: > I need to get an invocation time for the request that has a timestamp before the first log statement 🤔

Logs in lambda can be written before first invocation (during the initialization phase), e.g. with provisioned concurrency, it means that logs can be written minutes before the first invocation (or even with no invocations at all, when lambda won't get invoked).

So the question is, do we want to track the start time of each invocation? Or start time of lambda initialization? This code point is not right for any of that :)

Danwakeem: That is okay if the initialization logs are published before the request. We basically just want to get a timestamp as close to the invocation time as possible so any logs that happen after that the invocation has started will show up after the req payload

node/packages/aws-lambda-otel-extension/external/otel-extension-external/index.js
Danwakeem: Yeah no problem 👍

Danwakeem: Moved these to the same file we are generating the log payload in 😎 ✅

node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/wrapper.js
medikoo: Ok, so it's about the timestamp that will mark the start of the invocation and which will be updated at the start of every next invocation.

In this case, this is the right codepoint: https://github.com/serverless/console/blob/547d0d377b7fd1f3b9f536b3139faf27b4861b50/node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/wrapper.js#L58

I would definitely not expose it as global, but e.g., we can expose it as `EvalError.$serverlessIncocationStart` (it's sort of hacky, but not intrusive. We're adding a property on a global object, which is for sure not used in any code) and `Date.now()` is more natural way to get current timestamp

Danwakeem: Awesome, I moved off of the global variable to the `EvalError` variable and switched both date instances to use `Date.now()`.

Both `global` and `EvalError` feel hacky to me but at least `EvalError` is far less likely to be used in customer code so I like this suggestion 😎

node/packages/aws-lambda-otel-extension/external/otel-extension-external/index.js
```diff
@@ -280,7 +286,11 @@ module.exports = (async () => {
               case 'telemetryData':
                 lastTelemetryData = data;
                 if (data.record.responseEventPayload) {
-                  sendReport('response', stripResponseBlobData(data.record.responseEventPayload));
+                  const { eventData: currentRequestData } = getCurrentRequestData();
+                  sendReport(
+                    'response',
+                    createResponsePayload(data.record.responseEventPayload, currentRequestData)
```
medikoo: I'd stick to existing variable naming, and not name _eventData_ as _currentRequestData_, as in other parts `currentRequestData` is a different object.

Also I don't think it's necessary to create a variable here.

--- 

Notheless if some variables are not named right, we definitely should improve that, but I'd ensure we're using same naming convention in all parts

Danwakeem: @medikoo Alright I changed the name of that function `getCurrentRequestData` to `getCurrentRequestContext` and changed the `eventData` attribute to `requestData`.

I thought it would make more sense to call this `getCurrentRequestContext` since we are doing more than just storing request data in that object. Similarly I thought changing `eventData` to `requestData` would make that payload a bit more specific.

Let me know what you think 🙏

node/packages/aws-lambda-otel-extension/external/otel-extension-external/otel-payloads.js
```diff
@@ -111,8 +111,7 @@ const createResourceAttributes = (fun) =>
     }
   });
 
-const createLogPayload = (eventData, logs) => {
-  const spanData = eventData.span;
+const createAttributes = (eventData) => {
```
medikoo: Can we name it `requestData` here as well?

node/packages/aws-lambda-otel-extension/external/otel-extension-external/otel-payloads.js
```diff
@@ -149,6 +148,16 @@ const createLogPayload = (eventData, logs) => {
     resourceAtt[attribute.key] = Object.values(attribute.value || {})[0];
   }
 
+  return {
+    resourceAtt,
+    metricsAtt,
+  };
+};
+
+const createLogPayload = (eventData, logs) => {
```
medikoo: Same here and in other places in this file (?)

medikoo (commented):
@Danwakeem looks very good, I've just proposed one last var name unification

node/packages/aws-lambda-otel-extension/external/otel-extension-external/otel-payloads.js
Danwakeem: Ahh yeah good call 👍

node/packages/aws-lambda-otel-extension/external/otel-extension-external/otel-payloads.js
Danwakeem: Another good call. I think there is one other place in this file where this needs to change I will be sure to change that as well 😎

Danwakeem: This should be done ✅

node/packages/aws-lambda-otel-extension/external/otel-extension-external/otel-payloads.js
Danwakeem: Done ✅

medikoo (approved):
Looks great 👍 

Have you confirmed locally that integration test pass?

Danwakeem:
@medikoo yeah I had run them on the last commit and I reran them just now to double check 👍

PR #83, opened 2022-06-22
Docs reorg

hev:
This is a significant set of docs changes. It will likely
be hard to review this in the PR but here is what is included

* Removes using section
* Merges concepts with glossary into a product section
* Re-orgs some AWS items into main glossary
* Moves optimization related items from glossary to duration 
* Comments out some unused glossary items relatd to benchmarks (plan add these back later but would be confusing to end users atm)
* Reorgs FAQs to be more linkable
* Adds explainer about transaction limits to FAQs

Planned for future PR
* More links and navigation updates expected
* Fleshing out the platform section and organization.
* Revisiting what's in duration guide vs platform section
* Fleshing out more runtime details including benchmark glossary
* A round of updates to the root repo README.md


hev:
I think I can merge here without formal approval. About to do so.

PR #87, opened 2022-06-23
Test: Fix performance script after internal renames

medikoo:


Danwakeem (approved):
Running `test:performace` works great 👍

Screen Shot 2022-06-23 at 11 05 30 AM

PR #95, opened 2022-06-24
CI: Ensure needed token

medikoo:
I've configured the secret, but forgot it needs to be literally copied to env in the workflow file.


Danwakeem (approved):
Nice 😎 Which org does this correspond to?

medikoo:
> Nice 😎 Which org does this correspond to?

`test` -> https://github.com/serverless/console/blob/1aaad64ebcb4040f673a0e14989b934605e618cc/.github/workflows/node-integrate.yml#L198

PR #96, opened 2022-06-24
Optimise protobuf conversion

medikoo:
It cuts of overhead from 1s-500ms to max 100ms at invocation.

Additionally added workaround for AWS issue I observe from time to time locally


Danwakeem (commented):
This is a great idea 👍 I am noticing this issue when I test with console

The metrics don't seem to be coming trough to console. I am wondering if this is because of the changes to how we are using protobuf or something? 🤔 
Screen Shot 2022-06-24 at 10 59 48 AM

medikoo:
@Danwakeem I was actually testing it with console, and everything seemed to show. Below it's what I got just now by running benchmark on one case

Screenshot 2022-06-24 at 18 43 50


What exactly are you missing? You don't see invocations, or invocations are show, but you miss some data?

Danwakeem:
Huh maybe it isn't related to this PR then 🤔 For some reason I am not seeing metrics for invocations come through.

I am getting trace data but just not the metrics for some reason 🤷‍♂️ 

It might be something wrong with my test or something if you are seeing what you posted above after running benchmarks.

Danwakeem (approved):
The code looks good to me 👍 I am not sure what is going on with my test but I am fairly certain the issue is on my side somewhere because I am seeing some metrics dropped even when testing without using a custom built extension

PR #109, opened 2022-06-28
Bug fixes and benchmark improvements

medikoo:
### Extension fixes and improvements
- Fix regression introduced with #80. In this PR dependency on `eventData` was added when handling `telemetryData`, yet there are scenarios observed when `telemetryData` arrives to the extension before `eventData`. Logic was not prepared for that. I've added a patch which ensures that in such case we wait for the `eventData` and propagate response data after we have it.
- Remove `node-fetch` dependency from internal extension
- Ensure to reuse HTTP(s) request sockets (enforce `keepAlive`)

### Benchmark fixes and improvements
- Add `internalOnly` benchmark variant case, where only overhead of internal extension is measured
- Add support for `--memory-size` param, through which we can test different memory size scenarios
- Improve benchmark internals

### DIFF
--- a/node/packages/aws-lambda-otel-extension/external/otel-extension-external/helper.js
+++ b/node/packages/aws-lambda-otel-extension/external/otel-extension-external/helper.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const http = require('http');
+const https = require('https');
+
 const isObject = (value) => value && typeof value === 'object';
 
 const extensionVersion = (() => {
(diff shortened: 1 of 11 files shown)


Danwakeem (approved):
This looks good 👍 That is interesting that the response would come in in that order 🤔 Was this happening on lambda functions that would just immediately respond with something or on longer running lambda functions?

medikoo:
> Was this happening on lambda functions that would just immediately respond with something or on longer running lambda functions?

Yes, that's the case

PR #114, opened 2022-06-29
Fix promise resolution handling

medikoo:
Fix for a bug reported internally.

In wrapper we're overriding `context` methods, to catch user resolutions, yet same methods are used internally by AWS in case of promise resolution. Additionally, we have dedicated logic to ensure correct resolution if lambda is attempted to be resolved with boith callback and promise, and that one broke those methods for internal AWS resolver.

This patch ensures that for internal AWS resolver those methods work as if not patched


Mmarzex (approved):

Danwakeem (approved):
Looks good to me 👍 What exactly was happening before that was causing an issue?

medikoo:
@Danwakeem I've just updated main description with explanations

PR #116, opened 2022-06-29
manual-instrumentation(WIP)

hev:
This still needs work but is getting towards a more digest-able manual install process.

### DIFF
--- a/docs/integrations/aws/index.md
+++ b/docs/integrations/aws/index.md
@@ -32,8 +32,97 @@ In additional to runtime libraries installed by the Serverless Extension, an ext
 Lambda Extension is used to forward logs, metrics, and traces to
 Serverless Console. This component is independent from the language runtime.
 
-### Configuring the Serverless Console Extension
+## Configuring the Serverless Console Extension
 [Serverless Framework](../index.md) is the easiest way to instrument your app or
 service with the Serverless Console Extension. If your not using 
 Serverless Framework you'll need to build, and deploy the extension 
 yourself for each of your Lambdas functions. 
+
+### Pre-requisites
+Before you get started make sure you have the [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
+and a recent version of [Node.js](https://nodejs.dev/learn/how-to-install-nodejs) and a 
+[Serverless Console Organization](https://console.serverless.com?ref_website=https%3A%2F%2Fwww.serverless.com%2Fconsole%2Fdocs%2F) and [jq]().
+
+### Clone Repo and Install Node Modules
+Once you have everything you need you'll need to clone this repo
+and install node modules.
+
+```text
+git clone https://github.com/serverless/console.git
+cd console
+npm install
+```
+
+### Build the extension locally
+Next you'll need to run a set of build commands to build
+the internal and external extensions.
+
+```
+cd node/packages/aws-lambda-otel-extension/external/otel-extension-external
+npm install
+cd ../../internal/otel-extension-internal-node
+npm install
+cd ../../
+npm run build
+```
+This will generate an artifact in `dist/extension.zip`
+
+### Publish the extension to your AWS Account
+Once you have built the extension, you can publish it 
+to your AWS account.
+
+```text
+aws lambda publish-layer-version \
+   --layer-name sls-otel-extension \
+   --zip-file fileb://./dist/extension.zip 
+   --region us-east-1 
+```
+
+### Request an ingest token
+Next you'll need to request an ingest token based on your 
+organization id, service name, and environment you wish to
+deploy to. It's helpful to set these as environment variables.
+
+1. Go to your org settings page to get your org id and org token
+1. Set the following environment variables (used later)
+
+```text
+ORG_ID=<your-org-id> 
+ORG_TOKEN=<your-org-token>
+SERVICE_NAME=<your-service-name>
+FUNCTION_NAME<name-of-lambda-function>
+ENVIRONMENT=dev
+REGION=us-east-1
+```
+
+1. Curl the following endpoint to get an ingest token.
+```
+curl -X POST \
+  'https://core.serverless-dev.com/ingestion/kinesis/token' \
+  --header 'Accept: application/json' \
+  --header 'sls-token-type: orgToken' \
+  --header 'Content-Type: application/json' \
+  --header 'Authorization: Bearer 620cea8d-1d78-4e65-bd93-32d4609734f1' \
+  --data-raw '{
+	"orgId": "$ORG_ID",
+	"serviceId": "test-service",
+	"stage": "test"
+}'
+
+```
+### Configure necessary environment variables and tags
+
+Configure the AWS environment variables 
+
+
+aws lambda   update-function-configuration --function-name $FUNCTION_NAME --region us-east-1 --environment Variables=\
+"
+{
+  "common": { "destination": { "requestHeaders": "serverless_token=#YOUR INGEST TOKEN" } },
+  "logs": { "destination": "https://core.serverless.com/v1/logs" },
+  "metrics": { "destination": "https://core.serverless.com/v1/metrics" },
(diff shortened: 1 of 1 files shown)


hev:
@medikoo @Danwakeem @Mmarzex  - I was hoping to get your review of the following PR as a more approachable set of instructions for installing our Lambda Extension. A few things to note (that are called in the [Jira user story](https://serverlessinc.atlassian.net/browse/SC-412)). 

* We need to [surface the ORG ID](https://serverlessinc.atlassian.net/browse/SC-429) on our org page in order for users to be able to follow this. 
* We really should [re-organize our environment variables](https://serverlessinc.atlassian.net/browse/SC-442). This should probably happen regardless of how we proceed with this script. 

A few more things we could do as well.

* Eliminating the build and publish steps. @medikoo I think your making progress here.
* Adding ability to manage ingest tokens on the UI

cc @ac360

docs/integrations/aws/index.md
```diff
@@ -32,8 +32,92 @@ In additional to runtime libraries installed by the Serverless Extension, an ext
...
+## Configuring the Serverless Console Extension
 [Serverless Framework](../index.md) is the easiest way to instrument your app or
 service with the Serverless Console Extension. If your not using 
 Serverless Framework you'll need to build, and deploy the extension 
 yourself for each of your Lambdas functions. 
+
+### Pre-requisites
+Before you get started make sure you have the [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
+and a recent version of [Node.js](https://nodejs.dev/learn/how-to-install-nodejs) and a 
+[Serverless Console Organization](https://console.serverless.com?ref_website=https%3A%2F%2Fwww.serverless.com%2Fconsole%2Fdocs%2F) and [jq]().
```
Danwakeem:
```suggestion
[Serverless Console Organization](https://console.serverless.com?ref_website=https%3A%2F%2Fwww.serverless.com%2Fconsole%2Fdocs%2F) and [jq](https://stedolan.github.io/jq/download/).
```

docs/integrations/aws/index.md
```diff
@@ -32,8 +32,92 @@ In additional to runtime libraries installed by the Serverless Extension, an ext
...
+This will generate an artifact in `dist/extension.zip`
+
+### 3.Publish the extension to your AWS Account
+Once you have built the extension, you can publish it 
+to your AWS account.
+
+```text
+aws lambda publish-layer-version \
+   --layer-name sls-otel-extension \
+   --zip-file fileb://./dist/extension.zip
```
Danwakeem, line 77: Is this supposed to be `fileb://` and not `file://`?

docs/integrations/aws/index.md
```diff
@@ -32,8 +32,92 @@ In additional to runtime libraries installed by the Serverless Extension, an ext
...
+  --header 'Accept: application/json' \
+  --header 'sls-token-type: orgToken' \
+  --header 'Content-Type: application/json' \
+  --header 'Authorization: Bearer $ORG_TOKEN' \
+  --data-raw '{
+        "orgId": "$ORG_ID",
+        "serviceId": "test-service",
+        "stage": "$ENVIRONMENT"
+}' | jq -r .token.accessToken)
+```
```
Danwakeem, line 111: We need to add some details about the ingest token lifecycle to prevent customers from accidentally revoking a token they are currently using in another service or even on the same service once they decided they want to redeploy.

Here are a few rules/thing to keep in mind with these tokens

These tokens are rotated every 30 days if you make this same request with the same org, service, and stage to this endpoint.

For example 👇 

If you call this endpoint for the first time you will get a new token
If you call this endpoint some time after the first call within 30 days (Say on day 20) you will receive the same ingest token
If you call this endpoint some time after the first call and it has been more than 30 days (Say on day 35) you will receive a new ingest token and the old token will no longer be valid

Basically a good rule of thumb for these tokens is to generate 1 token per stack (Cloudformation, terraform, etc).

This is so you don't have to worry about deploying multiple stacks if you end up generating a new token with the 30 day policy I mentioned above.

Also it might be a good idea to not include this find/create token endpoint in any CI/CD process since a failed deploy might result in the existing service's token being revoked and not replaced if you call this endpoint outside of the 30 day period and the deployment fails.

Does this make sense?

Danwakeem (changes requested):
Looking good but I think we might want to add some warnings/disclaimers in the `Request an ingest token` section 👍

docs/integrations/aws/index.md
hev: oddly the fileb is actually what you want. i thought it was a typo when @medikoo shared it with me originally.

Danwakeem: Oh nice lol AWS just wants the be different I guess 😆

docs/integrations/aws/index.md
hev: I think this makes sense and I'll add some additional context here. Thanks for the rundown.

docs/integrations/aws/index.md
```diff
@@ -32,8 +32,92 @@ In additional to runtime libraries installed by the Serverless Extension, an ext
...
+[Serverless Console Organization](https://console.serverless.com?ref_website=https%3A%2F%2Fwww.serverless.com%2Fconsole%2Fdocs%2F) and [jq](https://stedolan.github.io/jq/download/).
+
+### 1.Clone the repo and install Node modules
+Once you have everything you need you'll need to clone this repo
+and install node modules.
+
+```text
+git clone https://github.com/serverless/console.git
+cd console
+npm install
```
medikoo, line 53: This installs the main dev dependencies of the repository, but won't install the dependencies for the `aws-lambda-otel-extension` package. For that additional installation step is needed:

```
cd node/packages/aws-lambda-otel-extension
npm install
```

docs/integrations/aws/index.md
```diff
@@ -32,8 +32,92 @@ In additional to runtime libraries installed by the Serverless Extension, an ext
...
+
+### 3.Publish the extension to your AWS Account
+Once you have built the extension, you can publish it 
+to your AWS account.
+
+```text
+aws lambda publish-layer-version \
+   --layer-name sls-otel-extension \
+   --zip-file fileb://./dist/extension.zip 
+   --region us-east-1
```
medikoo, line 78: This assumes user publishes the lambdas to `us-east-1` account.

If user publishes to different region, then this layer won't be accessible

docs/integrations/aws/index.md
```diff
@@ -32,8 +32,92 @@ In additional to runtime libraries installed by the Serverless Extension, an ext
...
+        "serviceId": "test-service",
+        "stage": "$ENVIRONMENT"
+}' | jq -r .token.accessToken)
+```
+### 5.Configure necessary environment variables and tags
+
+Configure the AWS environment variables using the AWS CLI
+
+```text
+aws lambda update-function-configuration --function-name $FUNCTION_NAME --region $REGION --environment Variables= {NAMESPACE=$NAMESPACE}
```
medikoo, line 117: It doesn't feel complete, or I miss something.

Layers, as built with above instructions, require `OTEL_RESOURCE_ATTRIBUTES`, `SLS_OTEL_USER_SETTINGS` and `AWS_LAMBDA_EXEC_WRAPPER` to be configured on lambda to work.

medikoo (commented):
@hev I've commented on proposed documentation which assumes user will build the layer manually directly from our repo.

However, I've just prepared a PR with which we will publish Console ready layers, which (aside of AWS specific `AWS_LAMBDA_EXEC_WRAPPER`) will require just one simple environment variable. Check: https://github.com/serverless/console/pull/119

docs/integrations/aws/index.md
hev: Yeah, I was imagining a future state. I am going to use the syntax you came up with in #119

docs/integrations/aws/index.md
hev: good point. I will use the env var I recommend setting.

hev:
I am going to close this for now and instead open a PR against the details in #119

PR #117, opened 2022-07-04
Test: Refactoring towards inclusion of custom benchmarks

medikoo:
Introduce changes, that make configuration of different (extended) benchmark setups easy, in a way that current functionalities can be reused.

Additionally:
- Improved layers cleanup logic (now we attempt to delete only layer version that are confirmed to exist, and not all numbers below the latest versions - that can be slow with large versions)
- Rename `protoConsole` variant to `console`
- Improve debug logs


Danwakeem (approved):
Looking good 👍 and the benchmarks appear to run just fine 😎

PR #118, opened 2022-07-04
Support bundling of custom settings into layers

medikoo:
This will allow us to pre-generate layers ready for the Console (with hardcoded urls) and publish it to our community AWS accounts


Danwakeem (approved):
✅

PR #119, opened 2022-07-06
Publish Console ready layers to Serverless community accounts

medikoo:
_Depends on #121 being merged first_

With every release of the extension, we will publish three variants of the Console-ready layers to _DEV_ and _PROD_ AWS Serverless community accounts.

Configured variants to be published
- `sls-otel-extension-node`- Node.js runtime specific extension that covers both external and internal part
- `sls-otel-extension-external`-  External runtime agnostic extension, currently written in Node.js
- `sls-otel-extension-internal-node`- Node.js runtime specific internal extension (which can eventually be used together with `sls-otel-extension-external`)

Removed the `@serverless/aws-lambda-otel-extension-dist` package, as once layers are distributed to AWS accounts, we no longer need to rely on the npm channel.

Needed secrets were already configured, and required infrastructure on AWS accounts is already setup

--- 

Once we have the first successful release, I'll update the Framework to rely on layers as published to our AWS accounts


hev:
@medikoo how does this affect #116 steps 1-3?

Currently this is...

1. Clone the repo and install Node modules
2. Build the extension locally
3. Publish the extension to your AWS Account

I would assume all of this could just be removed? The user would simply need to setup appropriate env vars including ingest token?

medikoo:
> @medikoo how does this affect #116 steps 1-3?

If user is meant to rely on layers as published in context of our AWS accounts, then there's no need to clone repo and build anything.

First thing that needs to be done is finding an ARN of latest version of the layer in given region.

In initial version we will only offer publicly available (hosted on S3) JSON file, that exposes all ARNS for the regions .so user will technically have to  (1) download our layers registry JSON (2) find layer ARN that's needed.

It's pretty bad to ask user to do that, so it'll be way nicer if we can setup a dedicated page which will list ARN's for latest versions in all regions.

hev:
> (1) download our layers registry JSON 

Can you share a command to do this somehow?

> It's pretty bad to ask user to do that, so it'll be way nicer if we can setup a dedicated page which will list ARN's for latest versions in all regions.

Can we work this into the release process somehow? Ideally the page for this would be https://github.com/serverless/console/releases IMO.

medikoo:
> Can we work this into the release process somehow? Ideally the page for this would be https://github.com/serverless/console/releases IMO.

That's an interesting idea, and indeed it can be done, I'll cover it in #116

medikoo:
> Can we work this into the release process somehow? Ideally, the page for this would be https://github.com/serverless/console/releases IMO.

@hev I've configured it that way, so with each release, we'll add a JSON registry file to release assets, and by inspecting them user will be able to seclude the region. In our documentation, we should be able to link to this JSON directly

I think this should be good enough. And I think it's nicer in a way it's separated from the release description, which otherwise will get polluted with info of tens of ARNs for different layers

Danwakeem (approved):
Looking good 👍

PR #120, opened 2022-07-07
Allow to run benchmark against separate external extension

medikoo:
Introduce `--extension-layers-mode` which allow to run benchmarks against separate layers hosting internal and external extension. Additionally path to custom external layer can be passed through `TEST_EXTERNAL_LAYER_FILENAME` environment variable.

Additionally:
- Set memory size default to 1024. That's also the default in the Framework, and it's probably closest value to reflect real world usage

### DIFF
--- a/node/packages/aws-lambda-otel-extension/test/benchmark/README.md
+++ b/node/packages/aws-lambda-otel-extension/test/benchmark/README.md
@@ -201,8 +201,16 @@ Generated benchmark results are output to the console in CSV format. To store th
 ./test/scripts/benchmark.js > benchmark.csv
 ```
 
-Benchmark run can additionally be configured with following CLI params
+### 4.1 Run customization
 
-- `--use-cases` Function use cases to test (e.g. `callback,express`)
-- `--benchmark-variants` Benchmark variants to test (e.g. `bare,externalOnly`)
+Benchmark run can additionally be customized with following CLI params
+
+- `--use-cases` Comma separated list of function use cases to test (e.g. `callback,express`)
+- `--benchmark-variants` Comma separated list of benchmark variants to test (e.g. `bare,externalOnly`)
 - `--memory-size` Memory size to provide to lambdas (by default it's 128MB).
+- `--extension-layers-mode` By default benchmark are run against single layer which contains Node.js external and internal extension. By setting `dual` value for this mode, benchmark will be run with those two extensions provided as separate layers (and external layer will come with Node.js binary bundled)
+
+Additionally extension layers that are used for testing can be overridden with following environment variables:
+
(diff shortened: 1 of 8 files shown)


node/packages/aws-lambda-otel-extension/test/benchmark/README.md
```diff
@@ -201,8 +201,17 @@ Generated benchmark results are output to the console in CSV format. To store th
...
+- `--benchmark-variants` Comma separated list of benchmark variants to test (e.g. `bare,externalOnly`)
 - `--memory-size` Memory size to provide to lambdas (by default it's 128MB).
+- `--extension-layers-mode` By default benchmark are run against single layer which contains Node.js external and internal extension. By setting `dual` value for this mode, benchmark will be run with those two extensions provided as separate layers (and external layer will come with Node.js binary bundled)
+
+Additionally extension layers that are used for testing can be overriden with following environment variables:
+
+- `TEST_LAYER_FILENAME` - Path to `.zip` file of a layer whih contains both external and internal Node.js extensions. Ineffective with `--extension-layers-mode=dual`
+- `TEST_EXTERNAL_LAYER_FILENAME` - Path to `.zip` file of a layer whih contains external (runtime agnostic) extension. Effective only with `--extension-layers-mode=dual`
+- `TEST_INTERNAL_LAYER_FILENAME` - Path to `.zip` file of a layer whih contains just internal Node.js extension. It's effective for `internalOnly` benchmark variant, and in other cases effective only with `--extension-layers-mode=dual`
```
Danwakeem:
```suggestion
Additionally extension layers that are used for testing can be overridden with following environment variables:

- `TEST_LAYER_FILENAME` - Path to `.zip` file of a layer which contains both external and internal Node.js extensions. Ineffective with `--extension-layers-mode=dual`
- `TEST_EXTERNAL_LAYER_FILENAME` - Path to `.zip` file of a layer which contains external (runtime agnostic) extension. Effective only with `--extension-layers-mode=dual`
- `TEST_INTERNAL_LAYER_FILENAME` - Path to `.zip` file of a layer which contains just internal Node.js extension. It's effective for `internalOnly` benchmark variant, and in other cases effective only with `--extension-layers-mode=dual`
```

Danwakeem (dismissed):
Looking good 👍 Are we splitting the extension into two physical layers or were we planning to bundle the go and node extension into one layer?

medikoo:
>  Are we splitting the extension into two physical layers or were we planning to bundle the go and node extension into one layer?

As layers are technically different projects I think we will provide them separately (that way fix that's needed in external part, can be just pushed for this single layer, and not that we need to update all runtime specific layers as well)

PR #121, opened 2022-07-08
Convert extension to be Serverless Console specific

medikoo:
As discussed internally we're converting the extension so it is Serverless Console specific.

In this PR I reshape user settings which now will be taken with `SLS_EXTENSION` (not `SLS_OTEL_USER_SETTINGS`) environment variable.

Additionally:
- (in response to above rename) improve the naming convention of internal (test purpose) environment variables
- Move test documentation to the test folder and document all test environment variables.

PR #122, opened 2022-07-11
Enable running unit tests against custom external extension binary

medikoo:
_Depends on #121 being merged first_

- Reconfigure external extension unit tests, so they're run against external extension as being run in a separate process
- Enable running unit tests against custom external extension binary


Danwakeem (approved):
This is just in preparation for the go extension correct?

medikoo:
> This is just in preparation for the go extension correct?

Yes, exactly

PR #123, opened 2022-07-11
How to instrument (revised)

hev:
This updated PR does a better job of getting us close to a simple process for instrumenting without framework now that #119 has been merged. That said, I still am running into a few gaps and/or snags.

1. We're still missing something to include the ARN we publish. @medikoo it came up separately that this could live on the release itself. I still think we're missing a step here but am not sure what it is. 
2. I am struggling to get the syntax right to set the environment variable `SLS_EXTENSION`. This seemingly should be easy but it's not for some reason.


medikoo (commented):
> We're still missing something to include the ARN we publish. @medikoo it came up separately that this could live on the release itself. I still think we're missing a step here but am not sure what it is.

After this is merged: https://github.com/serverless/console/pull/124 we should have first layers published. Let's follow up with instructions when it's in

> I am struggling to get the syntax right to set the environment variable SLS_EXTENSION. This seemingly should be easy but it's not for some reason.

Do you mean you have problems porting a JSON to the environment variable setting in `aws lambda update-function-configuration` CLI command?

As a side note, your commits in this PR are not recognized as yours, as if they're pushed using email address not attached to GitHub account

hev:
> Do you mean you have problems porting a JSON to the environment variable setting in aws lambda update-function-configuration CLI command?

Yes. My goal is to have people enter in a few local variables in their shell and then copy/paste the commands in the guide. 

>As a side note, your commits in this PR are not recognized as yours, as if they're pushed using email address not attached to GitHub account

Fixed this.

medikoo:
@hev we've just published the first layers, and instructions can be put to read layer ARN for desired region from registry file placed at https://github.com/serverless/console/releases/latest/download/sls-otel-extension-node.json

This address will always point the latest release.

It's valid until we rely current model where there's just one layer which host internal and external Node.js dedicated layers.

After switching to Go version, most likely we will have to revisit

hev:
@medikoo what aws-cli command do I need to reference the ARN for my function?

medikoo:
> @medikoo what aws-cli command do I need to reference the ARN for my function?

@hev I believe it's still just `aws lambda update-function-configuration` that should be used (unless I don't understand the question)

hev:
@medikoo that does seem like the obvious command, but I am unclear where to set it specifically reviewing [the docs](https://docs.aws.amazon.com/cli/latest/reference/lambda/update-function-configuration.html#options).  In fact I don't see any mention of extension on the page.

medikoo:
> @medikoo that does seem like the obvious command, but I am unclear where to set it specifically reviewing the docs. In fact I don't see any mention of extension on the page.

@hev extension is technically a layer, and its layer ARN needs to be passed to `--layers` CLI param (what's important here,  is that if given function has already layers attached, I think this function will remove those layers, unless user will also list them in that CLI command - it might be good to add info about that)

PR #126, opened 2022-07-13
Go external/fasthttp

rfaga:


rfaga:
> Why Go version creates a different payload? I think it should work exactly same

For logs, I'm not sure why node version is producing attributes starting in capital letter, should we keep this format? Ingest server recognizes both formats (it's not case sensitive)

PR #127, opened 2022-07-13
CI: Fix layers publication script

medikoo:
Apparently our first attempt for layers publication failed, due to few setup issues. This fixes that. 

By introducing temporary CI workflow I've ensured that we have layers for v0.5.0 published. It got finalized with: https://github.com/serverless/console/runs/7320061288?check_suite_focus=true

And we can see list of ARN's of published layers in JSON files attached to release: https://github.com/serverless/console/releases/tag/%40serverless%2Faws-lambda-otel-extension%400.5.0

I've also confirmed that layers registry in S3 bucket was updated, and that it's publicly accessible

### DIFF
--- a/.github/workflows/node-publish-aws-lambda-otel-extension.yml
+++ b/.github/workflows/node-publish-aws-lambda-otel-extension.yml
@@ -105,7 +105,7 @@ jobs:
             --bucket-name sls-dev-layers-registry \
             --layer-basename sls-otel-extension-node \
             --version $VERSION \
-            --layer-filename packages/aws-lambda-otel-extension/dist/extension.zip \
+            --layer-filename packages/aws-lambda-otel-extension/dist/extension.zip
           ./scripts/publish-extension-layers.js \
             --bucket-name sls-dev-layers-registry \
             --layer-basename sls-otel-extension-external \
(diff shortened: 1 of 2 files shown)


.github/workflows/node-publish-aws-lambda-otel-extension.yml
```diff
@@ -130,17 +130,17 @@ jobs:
             --bucket-name sls-layers-registry \
             --layer-basename sls-otel-extension-node \
             --version $VERSION \
-            --layer-filename packages/aws-lambda-otel-extension/dist/extension.zip
+            --layer-filename packages/aws-lambda-otel-extension/dist/extension.zip \
```
Danwakeem, line 133: This line also needs the back slash removed? The one above makes sense since it is the end of the script but since this is not the end of the line removing the slash here seems odd? 🤔

medikoo: This one has extra option following that line. This slash is actual fix :)

Danwakeem: Oh that is too funny 🤦 lol my bad totally misread the diff lol

Danwakeem (approved):
✅

PR #128, opened 2022-07-13
Improve `requre-in-the-middle` overhead

medikoo:
Introduce a patch which ensures that `require-in-the-middle` is initialized just once and not 20 times.

In benchmarks for case where `aws-sdk` is required (it implies a lot of require's internally), this cuts down overhead as introduced by `require-in-the-middle` from ca 590ms to 230ms.

Related benchmark data can be seen here: https://docs.google.com/spreadsheets/d/1Wg9aNEFWzZX1jTfc9g39ygCEcx6-0AAe8l4_QB4h_lo/edit?usp=sharing

I believe we can improve things further, but it'll require more sophisticated patch (most likely override of `require-in-the-middle` or revise of the approach as the whole).

I've confirmed with integration tests that instrumentation is not affected (we test it with `express` case)

-- 

Additionally improved `invoke-internal` script (up to recent changes) as I've used to observe improvement when testing the patch

### DIFF
--- a/node/packages/aws-lambda-otel-extension/test/fixtures/lambdas/requires.js
+++ b/node/packages/aws-lambda-otel-extension/test/fixtures/lambdas/requires.js
@@ -1,0 +1,8 @@
+'use strict';
+
+// eslint-disable-next-line import/no-unresolved
+require('aws-sdk');
+
+module.exports.handler = (event, context, callback) => {
+  callback(null, { statusCode: 200, body: JSON.stringify({ result: 'ok', filename: __filename }) });
+};
--- a/node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
+++ b/node/packages/aws-lambda-otel-extension/internal/otel-extension-internal-node/index.js
@@ -469,7 +469,10 @@ registerInstrumentations({
 
 module.exports = detectResources({
   detectors: [awsLambdaDetector, envDetector, processDetector],
-}).then((resource) => (tracerProvider.resource = tracerProvider.resource.merge(resource)));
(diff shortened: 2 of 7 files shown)


node/packages/aws-lambda-otel-extension/test/fixtures/lambdas/requires.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+// eslint-disable-next-line import/no-unresolved
+require('aws-sdk');
```
Danwakeem, line 4: It might make sense to require more than one supported library here that way we can see how the benchmarks are effected when loading multiple libraries.

If we include `apollo-server-lambda` this would also also include the graphql and express instrumentations and might make this app a bit closer to a real use case. In fact, if you would like I could turn this file into a small apollo server if you think that would be helpful?

medikoo: Yes, I totally agree we should create cases that test all those kind of resources we instrument, but that's not exactly scope of this PR (note we have JIRA tasks for creating those cases)

This specific case was to expose issue where a lot of requires are involved (and what is required is irrelevant here)

Technically issue was already seen in `express` case, but it was not that eminent, and requiring `aws-sdk` makes it way more visible.

Danwakeem: Okay sounds good 👍 If you need me to help create any of the other use cases for that other ticket I would be happy to help.

Danwakeem (approved):
:shipit:

PR #129, opened 2022-07-13
WIP - Add AWS Lambda docs for instrumentation v2

austencollins:


docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+AWS Lambda is a serverless compute service that lets you run code without provisioning or managing servers.  You can measure Metrics, Traces and Logs of AWS Lambda via Serverless Console.  
+
+# Tracing
+
+If you use Serverless Console's AWS Lambda Extension for Node.js, it will automatically trace your AWS Lambda invocations, providing a timeline rich with information we curated to help you assess, optimize and troubleshoot, as well as trace services and data that triggered your AWS Lambda functions.
+
+Before the handler of your AWS Lambda function runs, our Extension loads our SDK and performs the auto-instrumentation logic.  This is what it looks like:
+
+```javascript
+import serverless from 'serverless-sdk-aws-lambda'
```
medikoo: It feels more natural to stick to our npm namespace, so publish packages named as `@serverless/...`

In this case it could be named `@serverless/aws-lambda-sdk`

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+    * **express**
+      * **node.https**
+      * **node.http**
+      * **aws.sdk**
+
+# Spans
+
+## `aws.lambda`
+
+This is the parent Span (aka the Trace) for an AWS Lambda. It measures the combined lifecyle phases of AWS Lambda Initialization and Invocation, and any logic performed within the Invocation phase.
```
medikoo, line 35: It'll be good here to clarify how we interpret the case where the lambda process is initialized again in the context of the same Lambda instance after the crash. 

e.g., in case of timeouts or unhandled crashes, AWS reinitializes the process in the same container, and  this second initialization may perform differently (e.g. if on first initialization lambda cached some results to FS, second initialization can take advantage of that).

The question here is: Is reinitialization after crash continuation of same `aws.lambda` span, or new span is started.

I think accurate would be to continue under the same `aws.lambda` span. It'll also reflect what we see in AWS Cloudwatch (where the second initialization goes into the same log stream, and we take hash of log stream to be an id of this span).

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+
+Additionally, the duration of this Span is what AWS Lambda bills for, based on 1ms increments. Duration charges apply to initialization code that is declared outside of the handler in the Initiatlization phase, code that runs in the handler of a function during the Invocation phase, as well as the time it takes for code in any last running Extensions to finish executing during Shutdown phase.
```
medikoo, line 37: I think we need to clarify how are we going to measure the _Shutdown phase_. 

AWS advertises (at https://aws.amazon.com/lambda/pricing/)  that it charges for the time extension takes during the shutdown phase, yet this duration is not reported anywhere, and we're not capable of measuring it reliable internally.

I'll ask the AWS support about that

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+
+s.sdk.name: "s-aws-lambda-nodejs"
+s.sdk.version: "0.0.1"
+s.org.id: "53121673-361f-4181-3411-53e031312c09"
+s.platform: "aws-lambda"
+s.environment: "prod"
+s.namespace: "api"
+s.service: "aws-api-prod-getPoster"
+s.region: "us-east-1"
+s.request.id: "2be6c182-955a-4da9-9c39-d9e9d9febbaa" // Identifies this unique request
```
medikoo, line 58: Request is specific to invocation, so I think it doesn't belong to `aws.lambda` (unless I miss something)

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+aws.resource.arn: "arn:aws:lambda:us-east-1:423234:function:aws-api-prod-getPoster"
+
+/* Tags: aws.xray */
+
+aws.xray.trace_id: "Root=1-62136c8-56d0adcsafa323790afa;Parent=16b66safasf23e2ab3794;Sampled=0" // Optional. If available
+
+/* Tags: aws.lambda */
+
+aws.lambda.arch: "x64"
+aws.lambda.coldstart: false
```
medikoo, line 77: This feels as specific to `aws.lambda.invocation`

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+aws.lambda.duration: 1032 // Maps to the Cloudwatch Logs Report "Billed Duration"
+aws.lambda.error: false
+aws.lambda.error_culprit: "null"
+aws.lambda.error_exception_message: "null"
+aws.lambda.error_exception_stacktrace: "null"
+aws.lambda.error_exception_type: "null"
```
medikoo, line 83: This feels specific to invocation or initialization (technically initialization can also crash, but it cannot be a _handled_ error)

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+aws.lambda.event_source: "aws.apigateway"
+aws.lambda.error_timeout: false
+aws.lambda.event_type: "aws.apigatewayv2.http"
```
medikoo, line 86: Feels as specific to invocation (even if lambda is integrated to API Gateway, technically can be also directly invoked)

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+These are the Tags attached to this Span:
+
+```javascript
+/* Tags: Standard */
+
+s.span.id: "ja819kasfj10asfh" // This Span's ID
+s.span.parent.id: "e8a4e2a264145041" // This Span's parent Span ID.
+s.span.timestamp: "2022-07-13T20:10:48.794945024Z" // This Span's time.  Same as start_time.
+s.span.start_time: "2022-07-13T20:10:48.794945024Z" // This Span's start time.
+s.span.end_time: "2022-07-13T20:10:48.799375104Z" // This Span's end time.
```
medikoo, line 171: We may also add information on whether it's initialized under provisioned concurrency (so not with immediately following invocation) - still I'm not sure at this point whether we have means to have that information

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+
+Here is a table of contents of Spans we currently capture, in an example hierarchical format:
+
+* **aws.lambda**
+  * **aws.lambda.initialization**
+  * **aws.lambda.invocation**
+    * **express**
+      * **node.https**
+      * **node.http**
+      * **aws.sdk**
```
medikoo: If I understand correctly under one `aws.lambda` we can get multiple `aws.lambda.invocation` spans (and depending on decision on how we handled initialization after crashes, we could also have multiple `aws.lambda.intialization`)

So the flow could be as:

```
aws.lambda
  aws.lambda.initialization
  aws.lambda.invocation
    express
      ...
  aws.lambda.invocation
    express
      ...
  aws.lambda.initialization
  aws.lambda.invocation
    express
      ...
  aws.lambda.invocation
    express
      ...
```

Is my understanding correct?

Or do we want to create different `aws.lambda` span parent for each invocation?

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+
+s.span.id: "91ashafj0a91hasf" // This Span's ID
+s.span.parent.id: "e8a4e2a264145041" // This Span's parent Span ID.
+s.span.timestamp: "2022-07-13T20:10:48.794945024Z" // This Span's time.  Same as start_time.
+s.span.start_time: "2022-07-13T20:10:48.794945024Z" // This Span's start time.
+s.span.end_time: "2022-07-13T20:10:48.799375104Z" // This Span's end time.
+
+/* Tags: aws.lambda */
+
+aws.lambda.invocation.duration: 432 // Maps to the Cloudwatch Logs Report "Duration"
```
medikoo, line 209: It'll be good clarify if `aws.lambda` wraps just single invocation, or all that happens in given lambda process.

If it's the latter, then I'd expect it more as:

```
aws.lambda.processes[].invocation.duration: 432
```

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+
+This is a Span within a Trace for AWS Lambda. After Initialization, Extensions and the handler of the AWS Lambda function run in the Invocation phase. This phase includes:
+
+Running External Extensions in parallel with the function. These also continue running after the function has completed, enabling Serverless Console to capture diagnostic information and ingest metrics, traces and logs.
+
+Running the wrapper logic for Internal Extensions.
+
+Running the handler for your AWS Lambda.
+
+The Invocation phase is comprised mostly of your handler (i.e. your business logic), and you want to optimize that as best you can because its performance (combined with Initialization performance) will have the biggest impact on the experience for your users and customers.
```
medikoo, line 188: We may further tokenize t the invocation duration into:
- Handler duration
- Extension overhead duration
- Runtime duration (`totalDuration - (handlerDuration + extensionDuration)`)

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+s.platform: "aws-lambda"
+s.environment: "prod"
+s.namespace: "api"
+s.service: "aws-api-prod-getPoster"
+s.region: "us-east-1"
+s.request.id: "2be6c182-955a-4da9-9c39-d9e9d9febbaa" // Identifies this unique request
+s.trace.id: "db8e08ae5e6c71c35320ddad5ffb56a5"
+s.span.id: "e8a4e2a264145041" // This Span's ID
+s.span.timestamp: "2022-07-13T20:10:48.794945024Z" // This Span's time.  Same as start_time.
+s.span.start_time: "2022-07-13T20:10:48.794945024Z" // This Span's start time.
```
medikoo, line 62: Are we sure we want to follow `snake_case` naming convention instead of `camelCase` which is common in JS and JSON?

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+
+s.span.id: "80asjsaf10asdf" // This Span's ID
+s.span.parent.id: "e8a4e2a264145041" // This Span's parent Span ID.
+s.span.timestamp: "2022-07-13T20:10:48.794945024Z" // This Span's time.  Same as start_time.
+s.span.start_time: "2022-07-13T20:10:48.794945024Z" // This Span's start time.
+s.span.end_time: "2022-07-13T20:10:48.799375104Z" // This Span's end time.
+
+/* Tags: aws.sdk */
+
+aws.sdk.region: 'eu-west-1' // Region name for the request
```
medikoo, line 229: Is `aws.sdk` span a wrapper for all requests in given invocation (or lambda process), or is it specific to single request, and each request has its own `aws.sdk` parent span?

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+/* Tags: aws.sns */
+
+aws.sns.topic.name: "sns-lambda" // Taken from the TopicArn
+aws.sns.operation: "send" // Must be "send"
+aws.sns.message_ids: ["fja98jafs", ...] // Introspected from the response of this SDK operation
+```
+
+## `node.http`
+
+If you use the `http` module in Node.js, this Span is created.
```
medikoo: I assume  it's about `http.request` specifically, and we do not cover here usage of e.g. `http.createServer` 

Maybe then it'll better to name it `request` and in one of the properties store wether its _http_ or _https_ (?)

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,465 @@
...
+
+node.https.url: "http://myapp.com"
+node.https.method: "get"
+node.https.path: "/helloworld"
+node.https.status_code: 500
+```
+
+## `express`
+
+If an HTTP request is made from your AWS Lambda function, a Span is created.
```
medikoo, line 446: `express` feels specific to the Express web application framework. While I believe this span is about the handling of application endpoints (behind HTTP or HTTPS) in general (and that can be configured with many different frameworks).

Wouldn't it be better then to name it as `app` (some name generic to `http` and `https`), and in span property store wether its _http_ or _https_ and if Framework is used, name of the framework

docs/platforms/aws-lambda/nodejs/index.md
austencollins: Of course!  Change made.

docs/platforms/aws-lambda/nodejs/index.md
austencollins: Right.  You are speaking specifically of retries.  This is great feedback, and a place where users often need help.

I'm not sure I agree it should be within the same span because it is an additional invocation and reporting everything under one span is a happenstance that may affect our product experience and the clarity of what is happening.  For example, this will diverge from everything else which will indicate additional Invocations happened (our Metrics, the Metrics of other triggering infrastructure, etc.).  Numbers in charts will no longer match and I bet users will be more confused.

However, I do believe we can do something in the UX to link Traces that belong to a retry.  The user should be able to see the retry attempt number, and easily access the Trace of each retry. This would be hugely valuable.

Therefore, we should re-initialize everything, except somehow track a retry number and the Trace IDs of those retries, if possible.  Let me try and draft this out.

docs/platforms/aws-lambda/nodejs/index.md
austencollins: @medikoo I'm not sure we can, unfortunately.  I have omitted it for now.  It's also worth noting that it happens not with every invocation, and has a 300ms cap.  Therefore, this may be worth omitting.  Whereas Initialization and Invocation have a much larger performance, cost and end-user impact.

docs/platforms/aws-lambda/nodejs/index.md
austencollins: I believe this is required by our ingest cc @Mmarzex  It was previously known as `executionID`

docs/platforms/aws-lambda/nodejs/index.md
austencollins: I think it might be cleaner to keep most things on the parent Span, and limit those Lambda phases to just focus on duration.

docs/platforms/aws-lambda/nodejs/index.md
austencollins: In this specification, we've been careful to only focus on the three high-level Lambda lifecycle phases which AWS has made clear (Initialization, Invocation, Shutdown), stick only to AWS terminology, and only pull durations from those phases from an accurate piece of information given to use directly from Cloudwatch, if available.

I don't know what _Handler Duration_, _Runtime Duration_, and _Extension Overhead Duration_ is.  Can you point to AWS docs that use these terms and describe them?  Also, how would we get this information accurately?

docs/platforms/aws-lambda/nodejs/index.md
austencollins: Yes, this tag convention will be used across all languages.

docs/platforms/aws-lambda/nodejs/index.md
austencollins: Each AWS SDK operation (i.e. a request) should have its own `aws.sdk.<service>.<operation>` span.

docs/platforms/aws-lambda/nodejs/index.md
austencollins: That's right.  Though, I think we should name Spans after the module name first and foremost for explicitness and the clarity that follows.  Abstraction or clever terms will be too confusing ultimately.

Perhaps `node.http.request` is the right name here.

docs/platforms/aws-lambda/nodejs/index.md
austencollins: This is specific to `express`, and not about handling of application endpoints.  I still think we should not create abstract terms or get cute at this time.  People should be able to look at the name of the span and immediately understand what it is without relying on documentation to decode what it means.  This Span is intended to measure everything of Express.js (init, pre-middleware, request handler, post-middleware, etc.).

austencollins: Though, I do want to acknowledge I think your idea is clever!  Let's keep it simple for now, then figure out where to optimize conveying what is happening in users' logic.

docs/platforms/aws-lambda/nodejs/index.md
medikoo: Ok, so if I understand correctly, `aws.lambda` is to track a single invocation which in case of a cold start, will be accompanied by initialization that comes before.

If that's the case, then it doesn't perfectly fit the AWS Lambda lifecycle model. e.g. there can be lambda instances that were never invoked (so no request id) yet were initialized and possibly even errored, e.g. provisioned concurrency will trigger that.
Do we plan somehow to monitor such cases?

Also, shutdown phase in success cases will happen minutes after invocation (after lambda gets stale) so it's hard to attribute it to e.g. last invocation.

docs/platforms/aws-lambda/nodejs/index.md
Mmarzex: Yeah we need the `executionId`/`requestId` so that we have a way to link all the various upserts and logs as they come in.

docs/platforms/aws-lambda/nodejs/index.md
medikoo: It's sort of outlined in this graph: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#runtimes-lifecycle-invoke

The actual function invocation duration takes just part of the total invocation duration, especially with the extensions on board.

I think it's valuable to report to the user how long the actual function logic takes time, and how much is consumed by the runtime and eventual extensions

docs/platforms/aws-lambda/nodejs/index.md
austencollins: `aws.lambda` is intended to be the parent span, referring to handling a _request_ in Lambda.  There is nothing that accompanies it. There are only child spans that it contains, which are:

* `aws.lambda.initialization`
* `aws.lambda.invocation`

> If that's the case, then it doesn't perfectly fit the AWS Lambda lifecycle model. e.g. there can be lambda instances that were never invoked (so no request id) yet were initialized and possibly even errored, e.g. provisioned concurrency will trigger that.

`aws.lambda` and `aws.lambda.initialization` would still exist in this case.  That is, if we can still capture those in the above case.

>Also, shutdown phase in success cases will happen minutes after invocation (after lambda gets stale) so it's hard to attribute it to e.g. last invocation.

Shutdown has been ignored as a span because of this.

docs/platforms/aws-lambda/nodejs/index.md
```diff
@@ -0,0 +1,469 @@
...
+aws.dynamodb.scanned_count: 5 // The value of the ScannedCount request parameter.
+aws.dynamodb.attributes_to_get: ["lives", "id"] // The value of the AttributesToGet request parameter.
+aws.dynamodb.consistent_read: true // The value of the ConsistentRead request parameter.
+aws.dynamodb.index_name: "my_index" // The value of the IndexName request parameter.
+aws.dynamodb.limit: 50 // The value of the Limit request parameter.
+aws.dynamodb.projection: "Title, Description, RelatedItems, ProductReviews" // The value of the ProjectionExpression request parameter.
+aws.dynamodb.select: "ALL_ATTRIBUTES" // The value of the Select request parameter.
+```
+
+### `aws.sdk.dynamodb.updateitem`
```
medikoo, line 336: @ac360 Why not `updateItem`? We just want `[a-z0-9.]` as the allowed charset for span names?

PR #140, opened 2022-07-29
Base setup for `@serverless/aws-lambda-sdk` package

medikoo:
- Original handler decorator
- Internal extension that automatically decorates the handler
- Tests that confirm that automatic handler kind detection and decoration works properly
- CI/CD setup, that runs integration tests, publishes new version of package, and layers to our AWS accounts
- TS typings for public API's


Danwakeem (approved):
Looking good I was able to run integration tests and unit test without issue 😎

PR #141, opened 2022-08-01
SDK: CI and dependencies setup improvement

medikoo:
- Ensure to declare dev dependencies only at top level
- Do not publish to npm internal layer logic and build script
- CI: Ensure to not trigger validation checks for packages of which code was not touched


Danwakeem (approved):
🎸

PR #142, opened 2022-08-04
Base setup for `aws.lambda*` trace spans

medikoo:
- `TraceSpan` class implementation
- Configuration of `aws.lambda`, `aws.lambda.initialization` and `aws.lambda.invocation` spans and its integretion into internal extension (trace spans come with no tags assigned at this point, but interface is ready to host them)
- Integration tests that confirm that spans are created with each lambda invocation.

First number show, no overhead during initialization, and 1-2ms overhead for each invocation. That may go slightly bigger once we add other instrumentation and start writing tags, but I wouldn't expect it to go bigger than those few milliseconds, we'll see.


Danwakeem (dismissed):
The types are looking good to me 👍

PR #145, opened 2022-08-17
Fix `esbuild` reference in build script

medikoo:
Noticed that CI failed after the recent dev dependencies cleanup. It was caused by invalid esbuild path resolution. This patch fixes that


Danwakeem (approved):
👍

PR #148, opened 2022-08-19
Trace (global) tags and `aws.lambda*` spans tags resolution

medikoo:
- Rename `handler-decorator` module to `instrument`
- Require Serverless Console organization id for the SDK instrumentation
- Document setup of the SDK and all the auto generated spans (this documentation should be treated more as a draft which can serve as an inspiration for website documentation - once we have this documentation ready, I'll simplify documentation in this package cc @hev)
- Configure all global `slsTags` and all `aws.lambda*` tags, and document them
- Improve input validation in public API endpoints
- Improve typings
- Improve tests configuration

Note: There's still no writing of protobuf payloads configured it'll be taken care of in next PR


node/packages/aws-lambda-sdk/README.md
```diff
@@ -1,3 +1,63 @@
 # @serverless/aws-lambda-sdk
 
 ## AWS Lambda dedicated tracing utility
+
+Instruments AWS Lambda functions and propagates traces to the [Serverless Console](https://www.serverless.com/console/docs)
+
+### Setup
+
+#### 1. Register with [Serverless Console](https://console.serverless.com/)
```
Mmarzex, line 9: We should probably make it clear that you have to register at the Console website and setup the AWS integration.

node/packages/aws-lambda-sdk/.ts-types/index.d.ts
```diff
@@ -1,5 +1,7 @@
 import TraceSpan from './lib/trace-span';
 
+export const orgId: string;
```
Mmarzex, line 3: Something to keep in mind. I see you are using the `typesVersions` key in the `package.json`. This will only work when using typescript greater than v3.1. That is a pretty old version so I think it is safe to assume that this is fine but we may want to make a note somewhere about that.

medikoo: @Mmarzex if that may help TS users, it's a good idea. Can you propose a PR with a such note?

node/packages/aws-lambda-sdk/README.md
medikoo: This note explicitly asks to register with Serveless Console and provides a link.

What do you mean by AWS integration? (notes below state how Lambda function should be instrumented with SDK)

Note, it's not meant to be final documentation for Console and extension users. as I believe the final will be on our website, handled by @hev. Once it's there, we will definitely update this with a link

Mmarzex: In order to actual get data from the extension/SDK into Console you have to sign up and then have the role and log stream subscriptions created. So we just need to make sure that is clear here.

medikoo: >  you have to sign up and then have the role and log stream subscriptions created. 

Do we have some external doc on how it should be done? If not, can you explain here, and I'll add this to the doc

Mmarzex: I'm not sure that there is documentation on it yet. I'm okay to leave it for now and we can adjust after Adam has finished the documentation that will be on the main website.

Mmarzex (approved):

PR #149, opened 2022-08-22
Release sdk-schema v0.0.1

Mmarzex:


node/packages/sdk-schema/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@serverless/sdk-schema",
   "repository": "serverless/console",
-  "version": "0.0.0",
+  "version": "0.0.1",
```
medikoo: Why not `v1.0.0` ?

Using `v0.0.1` prevents us from relying on semver entirely (each release will have to be accompanied with explicit bumps in all dependent packages)

medikoo (commented):
We need also CHANGELOG.md

Ideally, if it's setup same way as e.g. in `aws-lambda-sdk` package. See: https://github.com/serverless/console/blob/bfb26dbf146755553bcf1a73dbb39a02ad05da49/node/packages/aws-lambda-sdk/package.json#L45

Also this need to be added https://github.com/serverless/console/blob/bfb26dbf146755553bcf1a73dbb39a02ad05da49/node/packages/aws-lambda-sdk/package.json#L17-L42

Sorry, somehow missed that in initial review

node/packages/sdk-schema/package.json
Mmarzex: Good catch, I wasn't thinking of that.

medikoo: If you feel that schema is not stable (and that we will possibly make some breaking changes to it at the beginning), we may start with `v0.1.0` and then shift to v1 once we will feel that all properties are used and not to be altered

Mmarzex: I say we do `v0.1.0` for now now that I'm thinking about it some more. Even though I feel it is pretty stable what we have defined, better to be extra safe while we are still setting up all the initial instrumentation.

medikoo (commented):
Looks great, let's just also add changelog and commit messages validation (to `validate.yml`) as we have in case of other packages:

https://github.com/serverless/console/blob/bfb26dbf146755553bcf1a73dbb39a02ad05da49/.github/workflows/node-validate.yml#L188-L192

https://github.com/serverless/console/blob/bfb26dbf146755553bcf1a73dbb39a02ad05da49/.github/workflows/node-validate.yml#L203-L211

medikoo (approved):
👍

PR #150, opened 2022-08-23
Revert "Release sdk-schema v0.0.1"

Mmarzex:
Need to fix the workflow for publishing this version, so going to revert and fix the workflow first and then re-apply.


medikoo (approved):
Instead of reverting you may just push PR with a fix, and push tag manually, that'll also work.

But that won't re-confirm that CI is fixed, so I'll leave it up to you

Mmarzex:
I did not think of that. I'll just do that instead of throwing a revert in the history. The workflow is a simple fix so I'm sure it will work next time.

PR #152, opened 2022-08-25
SDK Schema: Unify tags resolution handling

medikoo:
Top level tags are no longer expected to have underscores replaced with ".", and they were reorganized into nested structure

Additionally, fixed tests, so they detect as expected not defined properties (in previous version any structure will be reported as valid)


Mmarzex (approved):
Looks good 👍 

Might be a good idea to tag a new release with this change.

medikoo:
> Might be a good idea to tag a new release with this change.

There's one more PR I want to propose, and after that I think it'll be great to have release, unless you feel it'll be good to release this change promptly

Mmarzex:
I'm fine to wait if there are other changes you want to make.

PR #153, opened 2022-08-25
Updated to span_id

Danwakeem:
I think it might make more sense to just call this `span_id` instead of `id`. I know it is implied since it is attached to a span but given elastic already has an id property this might help reduce some confusion 👍


Danwakeem:
We will use the span_id as the elastic _id so we don't have to change this 👍

PR #156, opened 2022-08-26
Add workflow for generating documentation from protobufs

Mmarzex:


medikoo (commented):
Will add commits to PRs?

I think it's better to push generated docs directly to `main`, as e.g. we do in platform for SDK.

Also it might be good to include the current version of docs with this PR

.github/workflows/proto-generate-documentation.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          go install github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc@latest
+          which protoc-gen-doc
+          which protoc
+          cd proto/
+          mkdir -p ./docs/schema
+          protoc --doc_out=./docs/schema --doc_opt=markdown,docs.md ./serverless/instrumentation/tags/v1/*.proto ./serverless/instrumentation/v1/*.proto
+      - name: Add Doc Commit
+        uses: EndBug/add-and-commit@v9
+        add: ./proto/docs/schema/docs.md
```
medikoo: Is this path correct? It's different from one at which current docs are commit

.github/workflows/proto-generate-documentation.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Generate Docs
+        run: |
+          go install github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc@latest
+          which protoc-gen-doc
+          which protoc
+          cd proto/
+          mkdir -p ./docs/schema
+          protoc --doc_out=./docs/schema --doc_opt=markdown,docs.md ./serverless/instrumentation/tags/v1/*.proto ./serverless/instrumentation/v1/*.proto
+      - name: Add Doc Commit
+        uses: EndBug/add-and-commit@v9
```
medikoo, line 31: How will it behave if there are no changes in the generated doc?

proto/docs/docs.md
```diff
@@ -0,0 +1,688 @@
+# Protocol Documentation
```
medikoo: It feels a bit weird to name only doc `docs/docs.md`, how about  simply`README.md` ?

.github/workflows/proto-generate-documentation.yml
Mmarzex: That's the path I meant for it to be in. I just messed up the path locally.

proto/docs/docs.md
Mmarzex: Sure

.github/workflows/proto-generate-documentation.yml
Mmarzex: It will not commit if the diff is the same. That being said, we can change it that the workflow only runs against the folder that contains `.proto` files. That means it will always be a change.

medikoo: >  That means it will always be a change.

There may be just whitespace changes or changes to comments that I believe will not influence outcome doc.

Still if no empty commit will be pushed and CI will not crash it's ok

proto/docs/schema/README.md
```diff
@@ -0,0 +1,688 @@
+# Protocol Documentation
```
medikoo: In initial comment I meant to place it at `/proto/README.md` but I see we already have other documentation in proto context.

In light of that probably best name is `/proto/docs/schema.md`, wdyt ?

Mmarzex: That's fine with me, I'll make the change.

medikoo (approved):
Thanks @Mmarzex, that looks great 👍

PR #164, opened 2022-08-30
Add account_id to lambda root span tags

Mmarzex:


proto/serverless/instrumentation/tags/v1/aws.proto
```diff
@@ -69,6 +69,8 @@ message AwsLambdaTags {
   string request_id = 9;
   // The Lambda Function version.
   string version = 11;
+  // The AWS Account ID of the Lambda Function
+  optional string account_id = 12;
```
Danwakeem, line 73: Why optional? 🤔

medikoo (approved):
👍

proto/serverless/instrumentation/tags/v1/aws.proto
medikoo: So extension doesn't need to log it in payload :)

As I think of it, it might be nice to have two different payload schema endpoints, one for the payloads that extension writes and other for enriched payloads that end at the ingestion server

Mmarzex: > two different payload schema endpoints

Not sure I'm following this @medikoo, we have the singular endpoint that firehose writes to, not sure how we would split it.

Danwakeem: Also given an org could have multiple AWS account attached to it wouldn't the extension need to include the account id that the lambda function was invoked from?

If someone wanted to filter lambda invocations by one account we would need the account id (Or alias) included in the extension payloads to do this kind of filtering. Or is there some other way we are planning on grabbing the account id for an incoming span? 🤔

Mmarzex: The account id is included in the payload firehose forwards to us. See [here](https://gist.github.com/Mmarzex/4d76df2718b09deadcd86cc6e5a783c3)

Danwakeem: Ahh okay 🤔 if we are planning on using the account id from the firehose payload do we need the `account_id` here at all?

Mmarzex: Yup, we use these schemas in ingest still, so we need a field defined for it.

Danwakeem: Ahh okay, let me make sure I understand.

1. Extension will generate a span, package as a proto buff, and log to cloudwatch (No account id attached to span)
2. Firehose picks up the logs and forwards the logs to ingest (Account id attached to firehose record, No account id attached to span)
3. Ingest will parse the firehose record, attach the account id to the span, and forward the payload along to another function, database, whatever

Does that sound right?

Mmarzex: Yup that's correct. There will be some general process that attaches the tags from the inventory table, and tags from the firehose payloads.

Danwakeem: Gotcha cool thanks for the explanation 😎

Danwakeem (approved):
🚀

proto/serverless/instrumentation/tags/v1/aws.proto
medikoo: > Not sure I'm following this @medikoo, we have the singular endpoint that firehose writes to, not sure how we would split it.

@Mmarzex If I understand correctly we use this schema to process two different payloads (1) one that's written by the extension (simpler, which omits many otherwise required properties, which are other deducted from other CW properties) (2) one that's sent from a firehose to ingest, which contains full data.

So my point was, that if we want to ensure that properties such as `region`, are required in the payload as sent from a firehose, it'll need another variant of payload schema. Still, that ofc will make things more complex.

> account id that the lambda function was invoked from?

@Danwakeem I assume it's about the account where the function is hosted (which is part of its arn) and not the invoker account id, but maybe I miss something.

Danwakeem: @medikoo @Mmarzex answered my question above 👍

PR #165, opened 2022-08-31
AWS Lambda SDK: Instrument API Gateway REST API events

medikoo:


node/packages/aws-lambda-sdk/.ts-types/lib/trace-span.d.ts
```diff
@@ -1,5 +1,6 @@
 declare class TraceSpanTags extends Map {
   set(key: string, value: boolean | number | string | Date): TraceSpanTags;
+  setMany(Object, options?: { prefix?: string }): TraceSpanTags;
```
Mmarzex: Instead of Object here it would be more semantically correct to do,

```suggestion
  setMany(Record<string, boolean | number | string | Date>, options?: { prefix?: string }): TraceSpanTags;
```

Mmarzex (changes requested):

node/packages/aws-lambda-sdk/.ts-types/lib/trace-span.d.ts
medikoo: Great suggestion thanks!

Mmarzex (approved):

PR #185, opened 2022-09-12
Added dev mode rust extension

Danwakeem:
## Description
Just creating this PR to show how we might be able to build a dev mode extension that work independently from any other extension, cloudwatch subscription, etc. The extension might be a little big for us at 6MB but it might be possible to cut this down by modifying our toml file a bit 🤔 

There are instructions on how to install/build/deploy this layer in the README as well as info about the environment variables you would need.

Also for those more visual here is a loom video ;) 

[loom gif](https://www.loom.com/share/314dd09d10e543478dbe1b5684a96107)


medikoo (commented):
It's great to see Rust variant of the extension, and it looks pretty slick. Although as you mentioned it'll be probably nice to trim it's size, if possible.

Additionally in the end it'll be nice to run it through tests we have configured for external extension in otel version - there are a lot of race conditions that the extension needs to handle well - and our tests confirms on that.

The only problem I see is that we already have something similar prepared in Go by @rfaga and we definitely should not introduce two different external extensions to users (as each will add latency which we should keep at a minimum)

So the question is what is the purpose of that work? Do we want to postpone Go version and leverage Rust?

Danwakeem:
This is a good point @medikoo. I spoke with Austen and Max and they would like for me to revisit the Go version of the external extension in place of this Rust version. I will close out this PR and see if I can pick up with work with the Go extension.

PR #188, opened 2022-09-13
Dev Mode Extension

Danwakeem:
## Description
Rewrote the rust extension in go so we can start running some tests on this.

Demo Video 👇 
[](https://www.loom.com/share/a6b6caf57d414982a057eaa0f228ab58)


Danwakeem:
> It's great to see Rust variant of the extension, and it looks pretty slick. Although as you mentioned it'll be probably nice to trim it's size, if possible.
>
> Additionally in the end it'll be nice to run it through tests we have configured for external extension in otel version - there are a lot of race conditions that the extension needs to handle well - and our tests confirms on that.
> 
> The only problem I see is that we already have something similar prepared in Go by @rfaga and we definitely should not introduce two different external extensions to users (as each will add latency which we should keep at a minimum)
>
> So the question is what is the purpose of that work? Do we want to postpone Go version and leverage Rust?

I think @medikoo comment still applies here. It would be nice to get this link up to some of our integration tests and what not. But before I do that I am going to finish the PoC with some benchmarks and what not.

dev-mode/main.go
```diff
@@ -0,0 +1,116 @@
...
+	var logsStr string = ""
+	var requestId string = ""
+	flushLogQueue := func(force bool) {
+		for !(logQueue.Empty() && (force || strings.Contains(logsStr, string(logsapi.RuntimeDone)))) {
+			logs, err := logQueue.Get(1)
+			if err != nil {
+				logger.Error(printPrefix, err)
+				return
+			}
+			logsStr = fmt.Sprintf("%v", logs[0])
```
rfaga: I think `%v` should be used only for debug. It seems you should actually be parsing the log in its json format to get log type and flag as runtime done when it is done, like in https://github.com/serverless/console/blob/bfb77d9848e8fc267443afbe97e6e5f4dffa1140/go/packages/aws-lambda-otel-extension/external/reporter/parser_logs.go#L153
I know it should be pretty rare, but any platform runtimedone word in the same way inside log may wrongly flag as done
Also you are running a string compare check in the entire JSON log response from AWS for something that should be a .type field

dev-mode/go.mod
```diff
@@ -0,0 +1,12 @@
+module serverless/dev-mode-extension
+
+go 1.18
+
+require (
+	github.com/Workiva/go-datastructures v1.0.53 // indirect
+	github.com/golang-collections/go-datastructures v0.0.0-20150211160725-59788d5eb259 // indirect
+	github.com/google/uuid v1.3.0 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
+	github.com/sirupsen/logrus v1.9.0 // indirect
```
rfaga: would be good to use zap logger as it is the one we use for other golang packages

rfaga (commented):
I think it's fine to merge, but I have some suggestions that may worth to change in next versions

dev-mode/main.go
Danwakeem: I am going to hold off on parsing logs for now. I think the requirements for log parsing may have changed a bit since we were last building this layer. I will for sure keep this in mind when we add logs parsing back in 👍

.github/workflows/dev-mode-layer-publish.yml
```diff
@@ -0,0 +1,53 @@
+name: Publish Dev Mode Extension
+on:
+  push:
```
medikoo, line 3: I think it should rather be published after all tests pass, and ideally, if we also establish some release flow (as in the case of node packages)

dev-mode/.envrc
```diff
@@ -0,0 +1 @@
+AWS_PROFILE=sls-test
```
medikoo: This looks environment specific, do we need to commit it in?

.github/workflows/dev-mode-layer-publish.yml
medikoo, line 3: I think it should rather be published after all tests pass, and ideally, if we also establish some release flow (as in the case of node packages)

dev-mode/.envrc
Danwakeem: Huh I have this in the gitignore so I am not sure how this got added 🤔 Anyway you are correct this should be removed

.github/workflows/dev-mode-layer-publish.yml
Danwakeem: Yeah I am working on bringing some tests over now 👍

Danwakeem: Included unit tests 😎 Thanks for your help!

dev-mode/.envrc
Danwakeem: This should be gone now 😎

dev-mode/main.go
Danwakeem: Going to resolve this for now 👍 I will add more log parsing if we end up changing the schema. Thanks for your help with the benchmark stuff!

Danwakeem:
Got the unit tests working with the old system and added the release github action so I think this is ready to go 🙏 Let me know what you guys think or if there is anything else I should add here.

.github/workflows/dev-mode-layer-publish.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Build layer
+        run: make build
+      - name: Publish layers (dev)
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.OPEN_DEV_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.OPEN_DEV_AWS_SECRET_ACCESS_KEY }}
+        run: |
+          TEMP_ARRAY=($(echo $GITHUB_REF | tr "@" "\n"))
+          VERSION=${TEMP_ARRAY[@]: -1}
+          ./scripts/publish-extension-layers.js \
```
medikoo: For that to work I think we need to traverse into `node` directory, and also ensure all dependencies there are installed

.github/workflows/node-validate.yml
```diff
@@ -223,11 +223,6 @@ jobs:
             npx dump-release-notes-from-cc-changelog $NEW_VERSION
           fi
 
-      - name: Unit tests aws-lambda-otel-extension
```
medikoo: I think those steps should not be removed

.gitignore
```diff
@@ -3,3 +3,4 @@ coverage
 node_modules
 npm-debug.log
 package-lock.json
+.envrc
```
medikoo: Is `.envrc` generated or required by this project?

If it's more about developer preference of handling env vars, it'll probably be better to keep it in global gitignore file, and do not add it here (https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

node/packages/aws-lambda-otel-extension/test/README.md
```diff
@@ -49,17 +49,11 @@ How extensions behave, for various testing purposes, can be tweaked with followi
 
 ### Variables handled in extension logic
```
medikoo: I wouldn't interfere in a project that way, as it leaves it broken on its own, and that'll be problematic if decide to make e.g. _security_ or _deprecation_ release.

Ideally would be to simply bring those tests to `dev-mode` (written in go). 

Still if that's too much for now, we can create a copy of `test/unit/external/index.test.js` (e.g. to ` `test/unit/external/lambda-flow.test.js`), and just strip it out of internal extension reporting - this package should also pass this tests, as I think lack of internal extension is not blocking for external extension to work properly.

node/packages/aws-lambda-otel-extension/test/unit/external/index.test.js
```diff
@@ -551,258 +252,26 @@ describe('external', () => {
...
-
-        reports = await waitForReports();
-        expect(reports.map(([name]) => name)).to.deep.equal(['metrics']);
-      }
-    );
-    extensionProcess.stdout.on('data', (line) => stdoutLines.push(line));
-    await promise;
-  });
-
-  it('should handle case where platform.start arrives before invoke', async () => {
```
medikoo: This test is quite important

node/packages/aws-lambda-otel-extension/test/unit/external/index.test.js
```diff
@@ -551,258 +252,26 @@ describe('external', () => {
...
-
-        reports = await waitForReports();
-        expect(reports.map(([name]) => name)).to.deep.equal(['metrics']);
-      }
-    );
-    extensionProcess.stdout.on('data', (line) => stdoutLines.push(line));
-    await promise;
-  });
-
-  it('should handle case where platform.runtimeDone arrives before invoke', async () => {
```
medikoo: Same as that one

dev-mode/.gitignore
```diff
@@ -0,0 +1,3 @@
+bin
```
medikoo: We keep all node packages in `node` folder.

I think it might be good to also create some central point for `go` packages. So if we decide to introduce next one, they can share some common things. For that would place `dev-mode` in `go` folder. WDYT?

.github/workflows/dev-mode-layer-publish.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          TEMP_ARRAY=($(echo $GITHUB_REF | tr "@" "\n"))
+          VERSION=${TEMP_ARRAY[@]: -1}
+          TAG=${GITHUB_REF:10}
+          ./scripts/publish-extension-layers.js \
+            --bucket-name sls-layers-registry \
+            --layer-basename sls-dev-mode \
+            --version $VERSION \
+            --layer-filename dist/extension.zip \
+            --github-tag $TAG
```
medikoo: It'll be nice to ensure EOL's a EOF's

dev-mode/.gitignore
Danwakeem: Yeah this makes sense to me 👍 I was thinking that since dev mode wasn't specific to the language it is being deployed to it would make sense to put it in a folder that is named after the feature it is targeting.

node/packages/aws-lambda-otel-extension/test/unit/external/index.test.js
Danwakeem: Added these tests back since I moved the go tests will no longer share this file 👍

node/packages/aws-lambda-otel-extension/test/unit/external/index.test.js
Danwakeem: Added these tests back since I moved the go tests will no longer share this file 👍

node/packages/aws-lambda-otel-extension/test/README.md
Danwakeem: I moved the tests to go so this file should be reverted back to normal 👍 Sorry for the confusion

.github/workflows/dev-mode-layer-publish.yml
Danwakeem: Yeah missed this one 👍 Thanks!

dev-mode/.gitignore
Danwakeem: Alright I think this folder structure is better. Let me know what you think 🙏

.github/workflows/dev-mode-layer-publish.yml
Danwakeem: Yes 🤦‍♂️ when I was double checking some stuff I noticed this too I fixed this in another commit.

.github/workflows/node-validate.yml
Danwakeem: Added these back now that I moved the tests to the correct folder 👍

.gitignore
```diff
@@ -2,4 +2,4 @@
 coverage
 node_modules
 npm-debug.log
-package-lock.json
+package-lock.json
```
medikoo: nit: For good git diff'ing, let's maybe ensure EOL at EOF's (in all introduced files)

go/packages/dev-mode/tests/unit/index.test.js
```diff
@@ -0,0 +1,277 @@
+'use strict';
```
medikoo: I wouldn't host any JS here, as it's outside all the lint and prettier checks, and again we duplicate all dev deps setup

Instead can we have this file introduced in otel package, so both extensions passes? I think it'll be nice to validate  two of those external extension with it

go/packages/dev-mode/Makefile
```diff
@@ -0,0 +1,12 @@
...
+
+build: clean
+	GOOS=linux GOARCH=amd64 go build -ldflags "-s -w" -o dist/extensions/dev-mode-extension main.go
+	cd dist; zip extension.zip -r extensions/
+
+test:
+	npx mocha tests/unit/index.test.js
+
+deploy: build
+	aws lambda publish-layer-version --layer-name sls-dev-mode --description "This layer will collect logs and forward them to serverless console. This should only be used in development" --zip-file fileb://dist/extension.zip
```
medikoo: Missing EOL

go/packages/dev-mode/tests/unit/index.test.js
medikoo: Seems not addressed (?)

Danwakeem: The plan is to use the node tests while we work on rebuilding this testing framework.

medikoo: No JS code should be outside `node` folder. Can we put this as additional test to otel project, as discussed before?

Danwakeem: I don't know why that is would be a requirement though. These test are only related to the dev-mode extension so I don't know why I would put them in the otel folder, that feel really confusing. Especially since the plan is to replace the testing framework with something more go specific in the near future.

From my understanding the `node` folder was only created because we were creating runtime specific layers not because the code that the layer was written in was node.

medikoo: > I don't know why that is would be a requirement though.

1. All Node.js CI is configured against `node`, folder. For any JS code outside, there's no Prettier lint etc validation applied
2. All dev dependencies are configured in `node` context. Introducing any node projects outside doubles dev configuration for this repository, which should remain consistent across all work we introduce here.

So summarizing, introducing any JS packages outside makes maintenance of this repository really messy

Also, it's hard for me to understand why it's so problematic to configure this single test file in the context of otel project. 
This feels very straightforward to me, I'm happy to help if needed

go/packages/dev-mode/agent/forwarder.go
```diff
@@ -0,0 +1,110 @@
...
+			record := log.Record.(map[string]interface{})
+			requestId = record["requestId"].(string)
+			break
+		}
+	}
+	return requestId
+}
+
+func FormatLogs(logs []LogItem, requestId string) LogPayload {
+	var messages []LogMessage = make([]LogMessage, 0)
```
Mmarzex: This is really just a style thing, but in general it is more idiomatic when you are both defining and instantiating a variable to write it as  
```suggestion
 messages := make([]LogMessage, 0)
```

go/packages/dev-mode/agent/forwarder.go
```diff
@@ -0,0 +1,110 @@
...
+	logPayload := FormatLogs(logs, requestId)
+	if len(logPayload.Messages) == 0 {
+		return 0, nil
+	}
+	body, err := json.Marshal(logPayload)
+	if err != nil {
+		return 0, err
+	}
+
+	var _, internalLogsOnly = os.LookupEnv("SLS_TEST_EXTENSION_INTERNAL_LOG")
```
Mmarzex, line 90: Same as above, 
```suggestion
 _, internalLogsOnly := os.LookupEnv("SLS_TEST_EXTENSION_INTERNAL_LOG")
```

go/packages/dev-mode/agent/forwarder.go
```diff
@@ -0,0 +1,110 @@
...
+	if internalLogsOnly {
+		reportIncomingLogs(string(body))
+		return 200, nil
+	} else {
+		url := "https://core.serverless.com/dev-mode/log-socket/publish"
+		var _, isDev = os.LookupEnv("SERVERLESS_PLATFORM_STAGE")
```
Mmarzex, line 106:
```suggestion
	 _, isDev := os.LookupEnv("SERVERLESS_PLATFORM_STAGE")
```

go/packages/dev-mode/tests/unit/index.test.js
Danwakeem: > All Node.js CI is configured against node, folder. For any JS code outside, there's no Prettier lint etc validation applied

I could always add some linting steps here if that would help.

> All dev dependencies are configured in node context. Introducing any node projects outside doubles dev configuration for this repository, which should remain consistent across all work we introduce here.

This is not using any custom node packages from this project so I am not sure what maintenance savings we would get from moving these from this folder to the other especially since these tests will be so sort lived.

It is not difficult to move the tests from this folder to the other, that is a pretty trivial task. It is just strange to have tests for one monorepo project located in another monorepo project folder.

medikoo: > I could always add some linting steps here if that would help.

No need, CI is already setup, let's simply stick to rule that no JS code lands outside of `node` folder

> It is not difficult to move the tests from this folder to the other, that is a pretty trivial task. It is just strange to have tests for one monorepo project located in another monorepo project folder.

It's only because we want to test Go project, with tests which were written for JavaScript version. I'd say that this is a _strange_ thing we're dealing with here, but we agreed it's temporary

Nonetheless, let me know, I can setup needed test file in otel project shortly.

Other benefit of having it there is that we may confirm that both external extensions pass it (that also nicely validates tests on its own with extension that was already battle-tested in prod)

Danwakeem: > It's only because we want to test Go project, with tests which were written for JavaScript version. I'd say that this is a strange thing we're dealing with here, but we agreed it's temporary

Yeah agreed. I think this is the is real issue here, I am almost done with converting these tests to go anyway so I will just remove these javascript tests and replace them with go tests today 👍

Danwakeem:
Alright @medikoo I just pushed in the tests converted to go 🙌 I basically just copied the way you were doing the tests in node. Let me know if you think anything might be missing in the mock lambda API integration 🙏

go/packages/dev-mode/main_test.go
```diff
@@ -0,0 +1,279 @@
...
+	wg.Add(1)
+	go func() {
+		ExternalExtension()
+		wg.Done()
+	}()
+
+	// First invocation
+	requestId := uuid.New().String()
+	extensionReady()
+	extensionPlatformRuntimeDone(requestId)
```
medikoo: This is not a natural layout. As `platform. runtimeDone` is never emitted before `platform.start`.

To reflect a real scenario, it should rather be tested with the first invocation going normally and following being reported in a way that we get `platform.start`, `platform.runtimeDone` and `invoke` (in that order).

Also it'll be good to have these where order is `platform.start`, `invoke` and `platform.runtimeDone`

medikoo (commented):
Thanks Dan, looks very promising. I have just one suggestion. please see my coment

go/packages/dev-mode/main_test.go
Danwakeem: Ahh okay got it 🙏 I adjusted the order of these events in the test.

> To reflect a real scenario, it should rather be tested with the first invocation going normally and following being reported in a way that we get platform.start, platform.runtimeDone and invoke (in that order).

As for this comment, in the second test where I am testing the out of order runtime done event I added one normal invocation above to mimic what you mentioned 😎

go/packages/dev-mode/main_test.go
```diff
@@ -0,0 +1,287 @@
...
+	requestId := uuid.New().String()
+	extensionReady()
+	extensionPlatformStart(requestId)
+	extensionInvoke(requestId)
+	extensionPlatformRuntimeDone(requestId)
+	resetValidations()
+
+	// Second invocation
+	requestId2 := uuid.New().String()
+	extensionReady()
```
medikoo: This event is never repeated

go/packages/dev-mode/main_test.go
```diff
@@ -0,0 +1,287 @@
...
+		ExternalExtension()
+		wg.Done()
+	}()
+
+	// First invocation
+	requestId := uuid.New().String()
+	extensionReady()
+	extensionPlatformStart(requestId)
+	extensionInvoke(requestId)
+	extensionPlatformRuntimeDone(requestId)
```
medikoo, line 282: When doing this, we should confirm that it follows with `next` request being triggered

go/packages/dev-mode/main_test.go
```diff
@@ -0,0 +1,287 @@
...
+	wg.Add(1)
+	go func() {
+		ExternalExtension()
+		wg.Done()
+	}()
+
+	// First invocation
+	requestId := uuid.New().String()
+	extensionReady()
+	extensionPlatformStart(requestId)
```
medikoo: It'll be nice to have regular success case covered (where `platformStart` follows `invoke`)

medikoo (commented):
@Danwakeem what about integration tests?

Unit tests are nice to test those edge case scenarios which we cannot force from AWS infra, but otherwise, real confirmation comes from integration tests that confirm that extension works as expected in real world scenario

go/packages/dev-mode/main_test.go
Danwakeem: I could be misunderstanding what you are asking here but the next line is the invoke call right?

go/packages/dev-mode/main_test.go
Danwakeem: This is already being testing by the second invocation right? If the next request was never triggered again the second invocation data would never come through correct?

go/packages/dev-mode/main_test.go
Danwakeem: Ahh right my bad 👍 Removing

go/packages/dev-mode/main_test.go
medikoo: >  If the next request was never triggered again the second invocation data would never come through correct?

Yes, it's the reason, we should hold off with sending any further events until we record that `next` event was issued.

It's how it's done in JS tests: https://github.com/serverless/console/blob/e7e2cc667970bbc9a854fe47ef91157bfc6d1602/node/packages/aws-lambda-otel-extension/test/unit/external/index.test.js#L663

go/packages/dev-mode/main_test.go
medikoo: My point is, that ideally we should have separate tests for three scenarios:
- `invoke`, `start`, `runtimeEnd`
- `start`, `invoke`, `runtimeEnd`
- `start`, `runtimeEnd`, `invoke`

Danwakeem: Okay I created three tests for each of these scenarios. I updated the names to the tests so the names would match basically exactly what you posted above 👍

go/packages/dev-mode/main_test.go
Danwakeem: Ahh okay got it 👍 I did this a little differently since that won't exactly translate to these tests but I am making sure we are receiving the expected number of calls to next.

Danwakeem:
@medikoo Yeah I was working on these now 👍 I can drop these into this PR or a follow up. Either way works.

Danwakeem:
Alright @medikoo I have the integration tests set up with the same framework we are using for the other layers. Let me know if you think it looks alright 🙏 

One interesting note is that since go doesn't have a version number associated with the go module I figured we could just use the version number associated with the `package.json` file in the integration test folder I just set up. Seems a bit confusing but if you have any other ideas I am down to change this 👍

node/go/dev-mode/package.json
```diff
@@ -0,0 +1,53 @@
+{
```
medikoo: I don't think we need to create and publish any package for that.

Can't we simply have test files (with short doc) in `node/test/go/dev-mode` ?

node/go/dev-mode/test/lib/basename.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const testUid = require('./test-uid');
+
+module.exports = `test-sdk-${testUid}`;
```
medikoo: Basename should not be same as for SDK

.github/workflows/node-integrate.yml
```diff
@@ -34,6 +35,9 @@ jobs:
               - 'node/packages/aws-lambda-otel-extension/**'
             awsLambdaSdk:
               - 'node/packages/aws-lambda-sdk/**'
+            devModeExtension:
```
medikoo: Let's maybe name it `testGoDevMode` (name should follow package naming, and here it's good to add info that it's about _test_ of _go_

.github/workflows/node-integrate.yml
```diff
@@ -96,6 +100,15 @@ jobs:
           key: packages/aws-lambda-sdk/node-modules-v16-${{ runner.os }}-${{ github.ref }}-${{ hashFiles('node/packages/aws-lambda-sdk/package.json') }}
           restore-keys: packages/aws-lambda-sdk/node-modules-v16-${{ runner.os }}-${{ github.ref }}-
 
+      - name: Retrieve go/packages/dev-mode/node_modules from cache
+        if: steps.pathChanges.outputs.devModeExtension == 'true'
+        id: cacheDevModeExtensionNodeModules
+        uses: actions/cache@v2
+        with:
+          path: node/go/packages/dev-mode/node_modules
+          key: go/packages/dev-mode/node-modules-v16-${{ runner.os }}-${{ github.ref }}-${{ hashFiles('node/go/packages/dev-mode/package.json') }}
```
medikoo: Configuration of dev dependencies for all packages is kept at `node/package.json` so we don't need that.

We may just need to install extra deps for test fixtures, as we do other packages

.github/workflows/node-integrate.yml
```diff
@@ -496,6 +533,77 @@ jobs:
             git push --tags
           fi
 
+  integrateDevModeExtension:
+    name: Integrate Dev Mode Extension
+    runs-on: ubuntu-latest
+    needs: [linuxNode16, linuxNode14]
+    if: needs.linuxNode16.outputs.pathChangedDevModeExtension == 'true'
+    timeout-minutes: 30 # Default is 360
```
medikoo: Let's bring it down to `5` like in other jobs

node/go/dev-mode/CHANGELOG.md
```diff
@@ -0,0 +1,48 @@
+# Changelog
```
medikoo: Alll those files should not be commited in

node/go/dev-mode/package.json
Danwakeem: So I was planning on publishing the dev-mode extension using the same tag -> publish action we are using on the other layers. (See `dev-mode-layer-publish.yml` file) so I was thinking I should just use this package.json file to trigger that.

Is there some other way I should be creating the tags for the dev mode extension to publish the layer?

.github/workflows/node-integrate.yml
Danwakeem: Please confirm that i updated this correctly 👍

node/go/dev-mode/package.json
medikoo: But it's a Go package. I take that if it's to be published anywhere it's to some Go registry (not npm) and we should publish what's at `go` folder, and integration tests being at `node` should not be part of the publication (similarly as we do not publish node packages with tests)

Danwakeem: Okay I will remove this 👍

Danwakeem: Well actually I think this is another issue running the integration tests along with the other node tests.

We don't want to release the go extension unless integration tests pass so we need to trigger the publish via this workflow somehow and I don't think I can use the [workflow run trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) since this does more than just test the go extension.

Maybe I should just create a separate workflow file for the go integration tests?

medikoo: Usual configuration for integration CI is:
- Run integration tests
- If above passes and version of package has changed publish a tag

And its tag job that publishes the release.

So the trigger for publication is the actual version change in the package configuration file. What is the `package.json` equivalent in Go packages? Do we have one configured for `dev-mode` Go package here?

Danwakeem: From my understanding there is not really a `package.json` version equivalent for go. But I am fairly new to go so I could be wrong. I could just create a file that we add the version to and adjust the version manually 🤷 

Maybe @Mmarzex knows of a way to set a project version for go?

Danwakeem: The go docs just mention using a [git tag to manage the version](https://go.dev/doc/modules/publishing)

Danwakeem: Just confirmed with Roberto that versions are typically managed via git for go projects 😢

Danwakeem: Maybe I just create a file that we manually add the version number to so that we can use that to reference whether or not we should create a release tag? 🤷

medikoo: In such case maybe we simply introduce plain `VERSION` file that will have just version content as:

```
0.1.5
```

It'll allow us to rely on the same flow as with Node packages, which is quite convenient.

medikoo: It looks we both presented same idea at the same time :)

Danwakeem: Looking through all the tests you created has turned me into you 😎 Next thing you know I will have beautiful flowers outside the window I work next to 😆

Danwakeem: Alright I added a `version.json` file to the `go/packages/dev-mode` folder. Let me know what you think 🙏

.github/workflows/node-integrate.yml
```diff
@@ -143,6 +147,12 @@ jobs:
           cd node/packages/aws-lambda-sdk
           npm update --no-save
 
+      - name: Install test/go/dev-mode dependencies
```
medikoo: There shold be no dependencies installed in this folder (everything we need is installed at `node` and if something is missing let's simply add it there

.github/workflows/node-integrate.yml
```diff
@@ -496,6 +515,70 @@ jobs:
...
+          npm install
+
+      - name: Build Dev Mode Extension Layer
+        run: |
+          cd go/packages/dev-mode
+          make build
+
+      - name: Integration tests
+        run: |
+          cd node/packages/aws-lambda-sdk
```
medikoo: It's not that path :)

go/packages/dev-mode/version.json
```diff
@@ -0,0 +1,3 @@
+{
```
medikoo: I'd make it as plain text version (and name it `VERSION` in such case), it seems overkill to maintain it as JSON object

.github/workflows/node-integrate.yml
```diff
@@ -496,6 +515,70 @@ jobs:
...
+          TEST_INTERNAL_LAYER_FILENAME=../../../../go/packages/dev-mode/dist/extension.zip npm run test:integration
+
+      - name: Tag if new version
+        run: |
+          NEW_VERSION=`git diff -U0 ${{ github.event.before }} go/packages/dev-mode/version.json | grep '"version": "' | tail -n 1 | grep -oE "[0-9]+\.[0-9]+\.[0-9]+"` || :
```
medikoo: Keeping plain `VERSION` file, we just need to confirm on any changes in it (no need to check if given line changed and parse version out of it)

node/test/go/dev-mode/.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
```
medikoo: I think we do not need this file.  Let's just add `test/**` here: https://github.com/serverless/console/blob/ce3e992f6322458519ec9b53e4c1ba25b3f29778/node/.eslintrc.js#L10

node/test/go/dev-mode/.gitignore
```diff
@@ -0,0 +1 @@
+/dist
```
medikoo: This file I think is obsolete

node/test/go/dev-mode/.npmignore
```diff
@@ -0,0 +1,7 @@
+/dist
```
medikoo: Same that one

node/test/go/dev-mode/commitlint.config.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
```
medikoo: and that one, and some others here

node/test/go/dev-mode/test/benchmark/lib/resolve-common-benchmark-variants-config.js
```diff
@@ -0,0 +1,25 @@
...
+      {
+        configuration: {
+          MemorySize: memorySize,
+          Layers: [],
+          Environment: { Variables: {} },
+        },
+      },
+    ],
+    [
+      'internal',
```
medikoo: Many commited in files seem to either not be updated or should be committed in

.github/workflows/node-integrate.yml
Danwakeem: 🤦

Danwakeem:
@medikoo did I remove something that I shouldn't have in that last commit? I am wondering if it is because I removed the .eslintrc file or something? 🤔

Danwakeem:
Yeah adding back the `.eslintrc.js` file seems to do the trick 🤷

.github/workflows/node-integrate.yml
```diff
@@ -496,6 +509,65 @@ jobs:
...
+          make build
+
+      - name: Integration tests
+        run: |
+          cd node/test/go/dev-mode
+          TEST_INTERNAL_LAYER_FILENAME=../../../../go/packages/dev-mode/dist/extension.zip npm run test:integration
+
+      - name: Tag if new version
+        run: |
+          NEW_VERSION=`git diff -U0 ${{ github.event.before }} go/packages/dev-mode/version.txt` || :
```
Danwakeem: I am hoping I did this right 🤞

medikoo: It's best if you test it locally by introducing some temporary changes

.github/workflows/node-integrate.yml
```diff
@@ -304,6 +311,12 @@ jobs:
           cd node/packages/aws-lambda-sdk
           npm update --no-save
 
+      - name: Install test/go/dev-mode/dev-mode dependencies
```
medikoo: This should be removed. we shot define any dependencies at `node/test/go/dev-mode/dev-mode` (as all needed deps are defined at `node`)

.github/workflows/node-integrate.yml
medikoo: Please confirm locally that it'll work

go/packages/dev-mode/version.txt
```diff
@@ -0,0 +1 @@
+0.1.3
```
medikoo: Do we already have some publications?

If not, we should start with `0.0.0`

.github/workflows/node-integrate.yml
```diff
@@ -496,6 +509,65 @@ jobs:
...
+        run: |
+          cd node/test/go/dev-mode
+          TEST_INTERNAL_LAYER_FILENAME=../../../../go/packages/dev-mode/dist/extension.zip npm run test:integration
+
+      - name: Tag if new version
+        run: |
+          NEW_VERSION=`git diff -U0 ${{ github.event.before }} go/packages/dev-mode/version.txt` || :
+          if [ -n "$NEW_VERSION" ] && [ $NEW_VERSION != "0.0.0" ];
+          then
+            git tag @serverless/sls-external-extension@$NEW_VERSION
```
medikoo: `@serverless/..` tags are for packages publushed to npm exactly at this name.,

In case of Go I would follow some go specific convention e.g. `go/<package-name>@version`

node/test/go/dev-mode/.eslintrc.js
medikoo: Seems not addressed (in the sense that this file is still committed in)

node/test/go/dev-mode/package.json
```diff
@@ -0,0 +1,53 @@
+{
```
medikoo: This file should be removed

node/test/go/dev-mode/test/fixtures/lambdas/.eslintrc.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = {
+  extends: path.resolve(projectDir, '.eslintrc.js'),
+  rules: {
+    'import/no-extraneous-dependencies': [
+      'error',
+      {
+        packageDir: [
+          projectDir,
+          path.resolve(projectDir, 'packages/aws-lambda-sdk/test/fixtures/lambdas'),
```
medikoo: This configuration is not valid

node/test/go/dev-mode/test/fixtures/lambdas/250ms.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
medikoo: It'll be nice to centralize lambda fixtures for all the packages the rely on those, and not just copy them around. Still let's leave it for now.

However I would limit fixtures set, to only those that we need to test external extension, and I think we just need one simple fixture (as otherwise handling of different lambda variants it's in scope of internal extension, external extension should just confirm that it takes as expected all payloads that internal extension may generate).

So what exactly fixtures we need here (?)

 I think we can just live with regular `callback` test what do yo think? It'll produce all payloads (request, response and trace)

node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -0,0 +1,608 @@
...
+          ['v16', { configuration: { Runtime: 'nodejs16.x' } }],
+        ]),
+      },
+    ],
+    [
+      'callback',
+      {
+        variants: new Map([
+          ['v14', { configuration: { Runtime: 'nodejs14.x' } }],
+          ['v16', { configuration: { Runtime: 'nodejs16.x' } }],
```
medikoo: Let's limit test, to testing just this variant.

Unless we any valid reason to use few others? Let me know

node/test/go/dev-mode/test/lib/basename.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const testUid = require('./test-uid');
+
+module.exports = `test-go-dev-mode-extension-${testUid}`;
```
medikoo: Name cannot be too long as then it's possible it'll generate to long resource names (it's used as prefix)

So I'd limit it to `test-dev-mode-${testUid}`

node/test/go/dev-mode/test/lib/create-core-resources.js
```diff
@@ -0,0 +1,120 @@
...
+              filename: path.resolve(process.env.TEST_INTERNAL_LAYER_FILENAME),
+              skipBuild: true,
+            });
+            return;
+          }
+          config.layerInternalArn = await createLayer({
+            layerName: `${basename}-internal`,
+            filename: path.resolve(__dirname, '../../dist/extension.internal.zip'),
+          });
+          return;
```
medikoo, line 44: Where is external layer (we run test against) constructed here?

node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -0,0 +1,608 @@
...
+          expect(testConfig.name).to.have.string(
+            data.message.slice(data.message.lastIndexOf('\t') + 1).replace('\n', '')
+          );
+        }
+      }
+      if (testConfig.test) {
+        testConfig.test({ invocationsData, testConfig });
+      }
+    });
+  }
```
medikoo: This tests seems to test internal extension, while this is already well tested in its context.

Scope of this should be purely to confirm that external extension does it's job.

Same as in otel case, it'll be probably good to introduce mode where external extension payloads are written to CW logs and not send to server, and here were should read this logs and confirm they're as expected

go/packages/dev-mode/version.txt
Danwakeem: Yes, I have already published some versions of this extension. So I am just starting this file with the current version +1 to the patch number.

.github/workflows/node-integrate.yml
Danwakeem: That works for me 👍

node/test/go/dev-mode/package.json
Danwakeem: If I remove this file I won't have a nice `npm run test:integration` command to run.

Given the tests are written in node it might make sense to leave this file. I could always reference the path to mocha in a script or something but it would be a bit inconvenient 🤷

node/test/go/dev-mode/test/fixtures/lambdas/.eslintrc.js
Danwakeem: Ahh yeah good catch 👍

node/test/go/dev-mode/test/fixtures/lambdas/250ms.js
Danwakeem: I am thinking it might be best to just test the `thenable`, `callback`, and `250ms` tests. That way we cover the two different ways runtimeDone could be invokes with the first two tests and the `250ms` test will help test logs that come in after a delay 👍

node/test/go/dev-mode/test/integration/index.test.js
Danwakeem: I posted some more details in the comment above this one 😎

node/test/go/dev-mode/test/lib/create-core-resources.js
Danwakeem: Ahh I couldn't find this actually I was just using the `TEST_INTERNAL_LAYER_FILENAME` to run the tests 👍 I will change the path I was using in that environment variable to the default here.

node/test/go/dev-mode/test/integration/index.test.js
Danwakeem: Yeah this should be gone with the slimmed down testing set 👍

.github/workflows/node-integrate.yml
Danwakeem: Do you know how to recreate the `github.event.before` variable? 🤔

Danwakeem: Continuing the conversation on the below comment

node/test/go/dev-mode/package.json
medikoo: There's `package.json` at `node` folder, so if you want to have shortcut for running go integration tests I'll configure script there.

Here we neither introduce new dependencies scope nor new package, so having `package.json` is confusing.

node/test/go/dev-mode/test/fixtures/lambdas/250ms.js
medikoo: `thenable` and `callback` make a difference only for an internal extension. From an external extension point of view there's no real difference (that part is black box)

Actually, what I would introduce is a single lambda test that issues some logs over some time. e.g. runs for 5 seconds and issues 10 logs every half a second. That would well test that the extension actually gathers those logs on the go and e.g. in the next invocation.

node/test/go/dev-mode/package.json
Danwakeem: IMO it feels weird to have a dev-mode specific testing command in the root node `package.json` file

node/test/go/dev-mode/test/lib/create-core-resources.js
medikoo: `TEST_INTERNAL_LAYER_FILENAME` is for testing custom version if _internal_ layer, I'd keep it. Also if that env var is set, there's no default internal layer loaded.

So it means you tested just _external_ extension with no internal loaded (How it could pass?)

.github/workflows/node-integrate.yml
medikoo: it points to previous git commit hash.

node/test/go/dev-mode/test/lib/create-core-resources.js
Danwakeem: Lol this is a good question but it was running the tests alright before 🤷 Either way I updated this so it should be less confusing now

node/test/go/dev-mode/package.json
medikoo: I think it's ok, as we do not introduce any dev-mode Node.js package (as it's actually a Go package), we configure this script at global `package.json`

node/test/go/dev-mode/test/fixtures/lambdas/250ms.js
Danwakeem: This sounds like a good to me 👍 I went ahead an implemented almost exactly what you mentioned

.github/workflows/dev-mode-layer-publish.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Build layer
+        run: make build
+      - name: Publish layers (dev)
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.OPEN_DEV_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.OPEN_DEV_AWS_SECRET_ACCESS_KEY }}
+        run: |
+          TEMP_ARRAY=($(echo $GITHUB_REF | tr "@" "\n"))
+          VERSION=${TEMP_ARRAY[@]: -1}
+          ./../../../node/scripts/publish-extension-layers.js \
```
medikoo, line 52: Dependencies need to be installed for this script to run

node/test/go/dev-mode/.eslintrc.js
medikoo: Still, not addressed

node/test/go/dev-mode/test/fixtures/lambdas/callback.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
medikoo: Let's maybe name it `4s-logger` to distinguish from `callback` run in case of other packages (it'll also help when we will merge all fixtures)

node/test/go/dev-mode/test/lib/cleanup.js
```diff
@@ -0,0 +1,225 @@
...
+const deleteFunctions = async () => {
+  await Promise.all(
+    Array.from(await getAllFunctionNames(), async (functionName) => {
+      await awsRequest(Lambda, 'deleteFunction', { FunctionName: functionName });
+      log.notice('Deleted function %s', functionName);
+    })
+  );
+};
+
+const getAllRestApiIds = async (position = undefined) => {
```
medikoo: Let's remove unrelated cleanup logic

node/test/go/dev-mode/test/lib/create-core-resources.js
medikoo: This script seem to create external layer and create internal layer only if `TEST_INTERNAL_LAYER_FILENAME` var is set.

I think we should rather follow similar logic as in otel extension: https://github.com/serverless/console/blob/ce3e992f6322458519ec9b53e4c1ba25b3f29778/node/packages/aws-lambda-otel-extension/test/lib/create-core-resources.js#L49-L76

node/test/go/dev-mode/test/lib/create-core-resources.js
```diff
@@ -0,0 +1,123 @@
...
+            Resource: `arn:*:logs:*:*:log-group:/aws/lambda/${basename}*:*`,
+          },
+          {
+            Effect: 'Allow',
+            Action: ['logs:PutLogEvents'],
+            Resource: `arn:*:logs:*:*:log-group:/aws/lambda/${basename}*:*:*`,
+          },
+          {
+            Effect: 'Allow',
+            Action: ['sqs:*'],
```
medikoo: Do we need rights for `sqs` ?

node/test/go/dev-mode/test/lib/process-function.js
```diff
@@ -0,0 +1,298 @@
...
+        );
+        continue;
+      }
+      if (message.startsWith('⚡ DEV-MODE: Log###')) {
+        getCurrentInvocationData().logs.push(
+          JSON.parse(message.slice(message.lastIndexOf('###') + 3))
+        );
+        continue;
+      }
+      // if (startedMessage) {
```
medikoo: Let's not commit-in commented out code

node/test/go/dev-mode/test/lib/process-function.js
```diff
@@ -0,0 +1,298 @@
...
+      }
+      return currentInvocationData;
+    };
+    for (const { message } of events) {
+      if (message.startsWith('⚡ DEV-MODE: initialization')) {
+        processesData.push((currentProcessData = { extensionOverheadDurations: {}, logs: [] }));
+        continue;
+      }
+      if (message.startsWith('⚡ DEV-MODE: Overhead duration: External initialization')) {
+        currentProcessData.extensionOverheadDurations.internalInit = parseInt(
```
medikoo: It is not `internalInit` (which suggests it's about internal extension initialization)

node/test/go/dev-mode/test/lib/process-function.js
```diff
@@ -0,0 +1,298 @@
...
+          10
+        );
+        continue;
+      }
+      if (message.startsWith('START RequestId: ')) {
+        currentInvocationData = { extensionOverheadDurations: {}, logs: [] };
+        continue;
+      }
+      if (message.startsWith('⚡ DEV-MODE: Overhead duration: External request')) {
+        getCurrentInvocationData().extensionOverheadDurations.internalRequest = parseInt(
```
medikoo: Same here

node/test/go/dev-mode/.eslintrc.js
Danwakeem: @medikoo I had tagged you in a comment about this, [If I exclude this file the lint process fails.](https://github.com/serverless/console/pull/188/commits/dd2cb9c1217a63454175c1bee3be6d48e88f65d6)

.github/workflows/dev-mode-layer-publish.yml
Danwakeem: I added main project dependency installs. Let me know if I should add any others 👍

node/test/go/dev-mode/test/lib/create-core-resources.js
Danwakeem: Nope, removed 👍

node/test/go/dev-mode/test/lib/create-core-resources.js
Danwakeem: Ahh okay got it 👍 I updated all of these names to be `EXTERNAL` instead of `INTERNAL`

node/test/go/dev-mode/.eslintrc.js
medikoo: Please remove it, and let me see what the fail is.

If anything needs to be adjusted in lint configuration, we should do it in `node/.eslintrc.js` and not introduce a new one here

Danwakeem: [You can see the errors on the build when I removed it](https://github.com/serverless/console/actions/runs/3108502202/jobs/5037805345). I wasn’t sure what it needed.

medikoo: Back then errors were caused by `package.json` being in this folder (ESLint rules expected dev deps to be defined in its context), now it's no longer the case.

node/test/go/dev-mode/test/fixtures/lambdas/4sLogger.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
medikoo: Let's stick to hyphenated naming convention for filenames

node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -0,0 +1,78 @@
...
+        for (const [, logs] of invocationsData.map((data) => data.logs).entries()) {
+          const logPayload = logs[0];
+          expect(logs.length).to.equal(8);
+          expect(logPayload.name).to.equal(testConfig.configuration.FunctionName);
+          logs.forEach((logItem, index) => {
+            const message = (logItem.messages || [{ message: '' }])[0].message;
+            expect(`${testConfig.name.split('-')[0]} ${index + 1}`).to.have.string(
+              message.slice(message.lastIndexOf('\t') + 1).replace('\n', '')
+            );
+          });
```
medikoo: This test looks specific to `4s-logger` fixture. I would configure it specifically in its configuration (it's likely we will add another test cases here in future, and then those tests as configured now will stand in a way).

You can do it by adding `config.test` aside `variants` as here:  https://github.com/serverless/console/blob/d2f87d796f99ee35f38769a26fe8e0f5d19dc456/node/packages/aws-lambda-sdk/test/integration/index.test.js#L993-L1000

While in here, where we confirm on some general outcome from any test, I'd probably just stick to confirming that function resolved with expected values (which can be same for all). e.g as it's done here: https://github.com/serverless/console/blob/d2f87d796f99ee35f38769a26fe8e0f5d19dc456/node/packages/aws-lambda-sdk/test/integration/index.test.js#L1084-L1087

node/test/go/dev-mode/test/lib/create-core-resources.js
```diff
@@ -0,0 +1,118 @@
...
+        case 'nodeExternal':
+          if (process.env.TEST_EXTERNAL_LAYER_FILENAME) {
+            config.layerExternalArn = await createLayer({
+              layerName: `${basename}-external`,
+              filename: path.resolve(process.env.TEST_EXTERNAL_LAYER_FILENAME),
+              skipBuild: true,
+            });
+            return;
+          }
+          config.layerExternalArn = await createLayer({
```
medikoo, line 37: So at this point this test just loads external extension without internal, right?

I think it'll be good we also load internal, as it'll reflect real-world scenario, and we'll be sure we're covering the right way all eventual side effects that internal extension introduces

node/test/go/dev-mode/.eslintrc.js
Danwakeem: Ahh okay got it 👍 Seems to be working now.

node/test/go/dev-mode/test/integration/index.test.js
Danwakeem: Ahh okay yeah this makes sense. I kept the "ok" test in the main body and moved the 4s-logger test up to the config map 👍

node/test/go/dev-mode/test/lib/create-core-resources.js
Danwakeem: Yeah we might actually need one set of tests where we only have the dev mode extension and then another set of tests that have both internal and external extension loaded 🤔 

Would we want to add this test now even though we don't have the external and internal integration set up yet?

medikoo: If the intention is to include dev-mode extension independently out of our SDK, then let's skip it.

I somewhat assumed that our approach is to have them altogether from start, but that's not the case I take

Danwakeem: Yeah exactly 👍 The goal right now is to be able to use this external extension independently from the log subscription and sdk extension.

The next step for this extension is to integrate with the sdk in some fashion like I DMed you about last week. But even that integration will be optional so when we get to implementing that I will need to add tests with and without the SDK.

medikoo (dismissed):
Great work @Danwakeem ! 🚀

medikoo:
@Danwakeem just one thing: Can you merge it with squash merge (after you resolve the conflicts) ?

So we don't get the history noise into packages which we didn't alter after all

Danwakeem:
@medikoo yes for sure 💯 and dude thank you so much for your patience and thorough reviews while I stumbled my way through the project organization and testing ❤️ Lol I think I owe you one 🍻

.github/workflows/node-integrate.yml
```diff
@@ -496,6 +503,65 @@ jobs:
...
+        with:
+          path: ~/.npm
+          key: npm-v16-${{ runner.os }}-${{ github.ref }}-${{ hashFiles('node/**package*.json') }}
+
+      - name: Retrieve node_modules from cache
+        id: cacheNodeModules
+        uses: actions/cache@v2
+        with:
+          path: node/node_modules
+          key: node-modules-v16-${{ runner.os }}-${{ github.ref }}-${{ hashFiles('node/package.json') }}
```
medikoo, line 539: We need to also install those dependencies

node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -0,0 +1,78 @@
...
+  const coreConfig = {};
+
+  const useCasesConfig = new Map([
+    [
+      '4sLogger',
+      {
+        variants: new Map([
+          ['v14', { configuration: { Runtime: 'nodejs14.x', Timeout: 5 } }],
+          ['v16', { configuration: { Runtime: 'nodejs16.x', Timeout: 5 } }],
+        ]),
```
medikoo, line 30: Test seems to get back to common part

.github/workflows/node-integrate.yml
Danwakeem: Ahh I think I missed this altogether thanks for catching that 👍

node/test/go/dev-mode/test/integration/index.test.js
Danwakeem: Ahh yeah it looks like I removed that last commit I had made 🤦 I aded this back and fixed the `4sLogger` file name so it now reads `4s-logger` 👍

.github/workflows/node-integrate.yml
Danwakeem: I didn't see the step I just added on the other integrate tasks but I was assuming that might be because they are installing packages in sub-directories? 🤔

.github/workflows/node-integrate.yml
```diff
@@ -496,6 +503,70 @@ jobs:
...
+
+      - name: Install Node.js and npm
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
+
+      - name: Install node dependencies
+        run: |
+          cd node
+          npm install
```
medikoo: Let's use same install command as in case of other steps, so: `npm update --save-dev --no-save`

Danwakeem: Ohh my bad I didn't see the `Install main project dependencies` section listed above. This should be updated to the exact command and name 👍

medikoo (approved):
All looks good 👍

PR #217, opened 2022-09-26
Fixed layer publish workflow

Danwakeem:
## Description
I need to remove the default working directory from the dev mode layer publish workflow so that it is more consistent with other workflows and dependencies get installed correctly in the node folder 👍 

Here is [the bug](https://github.com/serverless/console/actions/runs/3129095593/jobs/5077753198) that I am seeing when running the workflow now


medikoo (approved):
👍

PR #218, opened 2022-09-26
Bump go extension version

Danwakeem:
## Description
I think either the second checkout or go installation was causing issues with the node modules cache/install 🤔 I moved these to the top of the workflow to hopefully fix the issue 🙏


.github/workflows/dev-mode-layer-publish.yml
```diff
@@ -68,5 +68,4 @@ jobs:
             --bucket-name sls-layers-registry \
             --layer-basename sls-external-extension \
             --version $VERSION \
-            --layer-filename go/packages/dev-mode/dist/extension.zip \
-            --github-tag $TAG
+            --layer-filename go/packages/dev-mode/dist/extension.zip
```
medikoo, line 71: Why `--github-tag` was removed?

Danwakeem: I could have this wrong but I thought that the `--github-tag` flag only needed when we wanted to create a release. Since we were not creating a package I didn't think we needed to do a release? 🤔

medikoo: True, it may not work if github release is not created. Ok let's skipt it

medikoo (approved):
👍

.github/workflows/dev-mode-layer-publish.yml
Danwakeem: Okay cool, I will double check once I get this working if the release process is actually working and if it isn't then I will add this back 😎

Danwakeem: I can confirm the s3 registry was updated 😎

PR #221, opened 2022-09-27
Upgrade dev mode processing endpoint

Danwakeem:
## Description
We are moving from the log socket endpoint to a log-ingestion endpoint so that we can follow the same enrichment process that we are following for log ingestion via kinesis.

Basically this just means that I am also moving from a JSON payload to a protobuf payload and updating the publish endpoint.


node/test/go/dev-mode/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "dev-mode-tests",
+  "version": "1.0.0",
+  "description": "Dev mode integration test dependencies",
+  "repository": "serverless/console",
+  "author": "Serverless, Inc.",
+  "license": "MIT",
+  "dependencies": {
+    "@serverless/sdk-schema": "^0.9.0"
```
medikoo: Let's remove this and related files

All _test_ dependencies are _dev_ dependencies, and for this repository we configure them in single place at `node/package.json`

Danwakeem: Ahh okay yeah that would definitely make things easier 😅

Danwakeem: Okay I moved the dependency 👍 I don't know why it didn't occur to me to do that before but this makes a lot more sense 😎

medikoo (commented):
👍

medikoo (approved):

PR #228, opened 2022-09-29
feat: Added req/res forwarding

Danwakeem:
## Description
Added req/res forwarding to the dev mode layer when the sdk layer is also attached to a function.

Any tips on organizing the integration tests would be appreciated 🙏 I just dropped in the internal layer when creating the resources and then added a flag in the `useCasesConfig` so that we could have tests that optionally used the internal and external layer together 👍


medikoo (approved):
Looks ok to me 👍

PR #232, opened 2022-10-03
SC-813: External extension with sdk endpoints

Danwakeem:
## Description
Added an internal server at `localhost:2772` with a `POST /spans` and `POST /reqres` route so our internal sdk extension can post spans and reqres data to our external server and we can forward this data long to our dev mode backend.

### DIFF
--- a/node/test/go/dev-mode/test/integration/index.test.js
+++ b/node/test/go/dev-mode/test/integration/index.test.js
@@ -61,16 +61,6 @@ describe('Integration', function () {
             const allReqResData = invocationsData.map((data) => data.reqRes);
             expect(allReqResData[0].length).to.equal(0);
             expect(allReqResData[1].length).to.equal(0);
-
-            // for (const [, reqRes] of invocationsData.map((data) => data.reqRes).entries()) {
-            //   expect(reqRes.length).to.equal(1);
-            //   for (const payload of reqRes) {
-            //     expect(payload).to.haveOwnProperty('region');
-            //     expect(payload).to.haveOwnProperty('accountId');
-            //     expect(payload).to.haveOwnProperty('timestamps');
-            //     expect(payload).to.haveOwnProperty('payloads');
-            //   }
-            // }
           },
         },
       },
--- a/node/test/go/dev-mode/test/integration/index.test.js
+++ b/node/test/go/dev-mode/test/integration/index.test.js
@@ -56,15 +56,21 @@ describe('Integration', function () {
         ]),
         config: {
           test: ({ invocationsData }) => {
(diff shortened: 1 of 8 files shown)


node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -56,15 +56,21 @@ describe('Integration', function () {
...
-            for (const [, reqRes] of invocationsData.map((data) => data.reqRes).entries()) {
-              expect(reqRes.length).to.equal(1);
-              for (const payload of reqRes) {
-                expect(payload).to.haveOwnProperty('region');
-                expect(payload).to.haveOwnProperty('accountId');
-                expect(payload).to.haveOwnProperty('timestamps');
-                expect(payload).to.haveOwnProperty('payloads');
-              }
-            }
+            // Replace with external + sdk integration results once the sdk is configured
```
medikoo, line 59: Is the expectation that once we send logs to an external extension, we stop writing payloads to logs?

Danwakeem: If I am understanding the question correctly, I don't think so. I believe we still want to write the existing sdk payloads to cloudwatch whether or not we are sending data to the external extension.

Danwakeem:
I guess I will mark this as ready for review since the updates to the external extension are done 🤷

node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -56,15 +56,21 @@ describe('Integration', function () {
...
+
+            // for (const [, reqRes] of invocationsData.map((data) => data.reqRes).entries()) {
+            //   expect(reqRes.length).to.equal(1);
+            //   for (const payload of reqRes) {
+            //     expect(payload).to.haveOwnProperty('region');
+            //     expect(payload).to.haveOwnProperty('accountId');
+            //     expect(payload).to.haveOwnProperty('timestamps');
+            //     expect(payload).to.haveOwnProperty('payloads');
+            //   }
+            // }
```
medikoo: Why this code was commented out? 

Also if given logic is no longer applicable, let's simply remove it instead of commenting it out

Danwakeem: Yeah no longer applicable I will remove the comments 👍

medikoo (commented):
One last suggestion: I think it'll be good to stick to naming convention we have in schema, and name endpoint for trace spans `/trace` and for request response `/request-response`

Danwakeem:
@medikoo Yeah good call man, I updated the endpoint names to `/trace` and `/request-response` 😎

medikoo (approved):
👍

PR #234, opened 2022-10-04
AWS Lambda SDK: Send payloads from internal to external extension in dev mode

medikoo:


Danwakeem:
@medikoo Just to make sure I am following here, this PR just setting up forwarding payloads to the external extension without including the req/res payload for instrumented libraries correct?

As you mentioned in [this PR](https://github.com/serverless/console/pull/235) the work to include the req/res payloads in instrumented libraries will happen in a follow up PR correct?

medikoo:
> @medikoo Just to make sure I am following here, this PR just setting up forwarding payloads to the external extension without including the req/res payload for instrumented libraries correct?

Yes, that's in scope of the next PR

node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -146,14 +153,15 @@ module.exports = (originalHandler, options = {}) => {
...
+          serverlessSdk._deferredTelemetryRequests.push(writeResponse(outcomeResult, context));
         }
       }
 
       const endTime = process.hrtime.bigint();
       awsLambdaInvocationSpan.close({ endTime });
       awsLambdaSpan.close({ endTime });
-      writeTrace();
+      serverlessSdk._deferredTelemetryRequests.push(writeTrace());
+      await Promise.all(serverlessSdk._deferredTelemetryRequests);
```
Danwakeem, line 164: Does this mean that we are deferring sending all of the trace payloads until the root lambda span is closed? 🤔

medikoo: In scope of the PR, I sent to dev mode same payloads as those written to the logs.

(This PR is purely about connecting internal with external and propagate things as they are now)

In next PR I'll update it spans are send right when they're closed

Danwakeem: Ahh okay awesome 👍

Danwakeem (approved):
This is looking like a good first step to me 👍

PR #235, opened 2022-10-04
SDK Schema: Release v0.10.1

medikoo:


Danwakeem (approved):
I am not sure what the process is to release the same proto schema package in golang 🤔  Does that get released along with this or maybe @Mmarzex would know?

medikoo:
> I am not sure what the process is to release the same proto schema package in golang 🤔 

I think the only person that touched Go here was you @Danwakeem :)

If anything needs to be made keep Go in sync I believe it needs to be added to SDK Schema publish job

Danwakeem:
@medikoo Well I think Max was kind of owning the protobuf portion of this project so I don't know much about this part of the project. There is a go package that is created based on the protobuf schema in this folder and I wasn't sure how that was getting created.

I just DMed Max and he said there is some manual process he is following right now to publish the go package 👍 I will see if I can help finish out the github action for this.

medikoo:
> There is a go package that is created based on the protobuf schema in this folder and I wasn't sure how that was getting created

@Danwakeem ah yes.. sorry forgot about it. Still, to fully clarify, this PR was purely about the release of NPM package of SDK schema (JS purposes only), and that even doesn't automatically update the Lambda SDK package to use it. That will need to happen in the next step in other PR.

I guess a similar flow should be followed in GO?

PR #236, opened 2022-10-04
Add Permissions for Cloudtrail and Eventbridge

Mmarzex:
This PR adds permissions for our AWS integration to react to CloudTrail events. The big permissions added are for,

1. Creating a CloudTrail if a Trail is not already created
2. Creating the Event Rule and API Destination in Eventbridge

All write permissions are scoped to the expected names for our resources only.

### DIFF
--- a/instrumentation/aws/iam-role-cfn-template.yaml
+++ b/instrumentation/aws/iam-role-cfn-template.yaml
@@ -24,20 +24,58 @@ Resources:
     Type: AWS::CloudFormation::CustomResource
     Version: "1.0"
     Properties:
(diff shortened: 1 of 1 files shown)


Danwakeem (commented):
When I try and drop this into IAM I seem to get the `Your policy character exceeds the non-whitespace character limit of 6,144.`. Is this something we are already accounting for?
Screen Shot 2022-10-04 at 1 14 40 PM

Danwakeem (dismissed):
Lol Ignore my comments I still had the full cloudformation template in my clipboard when I was checking the size 🤦 Should be good

Mmarzex:
No worries @Danwakeem , had to update the one S3 permission, picked the wrong name 🤦 . If you could give it another look now.

Danwakeem (approved):
Looking good 🚀

PR #237, opened 2022-10-04
Forward all dev mode data as a application/x-protobuf

Danwakeem:
## Description
Updated dev mode extension to forward all data as `application/x-protobuf`.

I also went ahead and updated the integration tests to check for traces and request-reponse data 😎

### DIFF
--- a/node/test/go/dev-mode/test/lib/process-function.js
+++ b/node/test/go/dev-mode/test/lib/process-function.js
@@ -222,15 +222,11 @@ const retrieveReports = async (testConfig) => {
         continue;
       }
       if (message.startsWith('⚡ DEV-MODE: ReqRes###')) {
-        getCurrentInvocationData().reqRes.push(
-          JSON.parse(message.slice(message.lastIndexOf('###') + 3))
-        );
+        getCurrentInvocationData().reqRes.push(message.slice(message.lastIndexOf('###') + 3));
         continue;
       }
       if (message.startsWith('⚡ DEV-MODE: Traces###')) {
-        getCurrentInvocationData().traces.push(
-          JSON.parse(message.slice(message.lastIndexOf('###') + 3))
-        );
+        getCurrentInvocationData().traces.push(message.slice(message.lastIndexOf('###') + 3));
         continue;
       }
       if (message.startsWith('⚡ DEV-MODE: Extension overhead duration: External shutdown')) {
--- a/node/test/go/dev-mode/test/lib/process-function.js
+++ b/node/test/go/dev-mode/test/lib/process-function.js
@@ -43,7 +43,7 @@ const create = async (testConfig, coreConfig) => {
             }
           : {}),
         SERVERLESS_PLATFORM_STAGE: 'dev',
-        SLS_ORG_ID: process.env.SLS_ORG_ID,
+        SLS_DEV_MODE_ORG_ID: process.env.SLS_ORG_ID,
         SLS_TEST_EXTENSION_LOG: '1',
       },
     },
@@ -183,7 +183,12 @@ const retrieveReports = async (testConfig) => {
     for (const { message } of events) {
       if (message.startsWith('⚡ DEV-MODE: initialization')) {
         processesData.push(
-          (currentProcessData = { extensionOverheadDurations: {}, logs: [], reqRes: [] })
+          (currentProcessData = {
+            extensionOverheadDurations: {},
+            logs: [],
+            reqRes: [],
+            traces: [],
+          })
         );
         continue;
       }
@@ -195,7 +200,12 @@ const retrieveReports = async (testConfig) => {
         continue;
       }
       if (message.startsWith('START RequestId: ')) {
-        currentInvocationData = { extensionOverheadDurations: {}, logs: [], reqRes: [] };
+        currentInvocationData = {
+          extensionOverheadDurations: {},
+          logs: [],
+          reqRes: [],
+          traces: [],
+        };
         continue;
       }
       if (message.startsWith('⚡ DEV-MODE: Overhead duration: External request')) {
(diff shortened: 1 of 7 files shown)


medikoo (commented):
Why exactly do we need to base64 encode?

This will create significantly larger payloads, also we didn't do that with otel version

Danwakeem:
@medikoo I need to pass a bit more information than just the raw proto payload since there is some information we were expecting to get from the kinesis stream event that we do not get inside the proto payload.

So I am base64 encoding so I can match the format we have on the existing proto processor.

medikoo:
> @medikoo I need to pass a bit more information than just the raw proto payload since there is some information we were expecting to get from the kinesis stream event that we do not get inside the proto payload.

Why not make it part of a payload?

This change may affect noticeably our performance

Danwakeem:
@medikoo I guess we could do that but we would need to change the proto format fairly significantly 🤷

Right now, for example, if you need to get the `accountId` or a `requestId` for a trace there is not a consistent place to retrieve this information from our current proto payloads across all spans. I know it is on the `aws.lambda` span but we can not rely on this as it will not always be included in each trace payload. Same issue goes for request/response but I don't think those include `accountId` at all.

I think our proto payloads are missing this information, because when were are ingesting the sdk proto data from kinesis that information is attached to the root event object so you don't have to dig into the payload for a value that may or may not be there.

Danwakeem:
> This change may affect noticeably our performance

Are you thinking it would affect performance on the API call that we are making?

medikoo:
> @medikoo I guess we could do that but we would need to change the proto format fairly significantly 🤷

Adding support for `accountId` and `requestId` as top properties of trace payload, does not feel as significant change (it won't be even breaking). It's simply questiion of adding it here: https://github.com/serverless/console/blob/8e74674335bcfeed76e3ad891dfb685c15299228/proto/serverless/instrumentation/v1/trace.proto#L14-L22 (and here for request respons payload: https://github.com/serverless/console/blob/main/proto/serverless/instrumentation/v1/request_response.proto#L12-L29)

Alternatively, we can also pass it with headers

> Are you thinking it would affect performance on the API call that we are making?

It'll make payloads larger, and we will totally diminish one of the advantages of protobuf which is minimal size due to how it packs data (in base64 format, size is similar to regular JSON)

Also even if not that, it feels as some hacky solution, that increases technical debt

Danwakeem:
Okay I can update the proto payloads to include the additional fields and then we can make the changes to the internal sdk  👍

Mmarzex:
Instead of changing the base trace/span payloads to have those top level tags, which I would rather we avoid changing the tagging convention we have, why don't we create a `DevModePayload` envelope structure. That way we can have tags that you need for every span without changing all our top level payloads. Thoughts @Danwakeem ?

Danwakeem:
@Mmarzex I guess the only issue I see with that is that we don't necessarily want to the dev mode payloads to be too significantly different from what we are normally processing and the data we are receiving through dev mode.

Maybe we should go down that route when/if we have bigger differences between dev mode and normal ingestion? What do you think about that idea?

We are really only missing three simple fields for dev mode

Traces
- AccountId
- RequestId
- Region (This one is there already just not normally filled out)

Req/Res
- Timestamp
- AccountId
- Region  (This one is there already just not normally filled out)

Mmarzex:
@Danwakeem I understand that, but we should not be surfacing AWS specific tags into the root tag payloads.

I could have something like this, 

```protobuf
message DevmodePayload {
  string account_id = 1;
  string region = 2;
   string request_id = 3;

   oneof payload {
     Span span = 1;
     ReqRes reqRes = 2;
  }
}
```

Then that way we adhere to our overall tagging standards, while meeting the devmode requirements.

Danwakeem:
@Mmarzex Okay I like this idea 🤔 I can create a PR for this now 👍

Danwakeem:
~~Alright @medikoo I updated the extension so it won't process the payloads from the internal extension at all it will just use the new dev mode proto format and forward that along for processing~~ 👍

Danwakeem:
Well actually sorry I need to do the same for logs 🤦

Danwakeem:
Okay @medikoo now it should be ready to go 👍

node/test/go/dev-mode/test/lib/process-function.js
```diff
@@ -212,9 +222,11 @@ const retrieveReports = async (testConfig) => {
         continue;
       }
       if (message.startsWith('⚡ DEV-MODE: ReqRes###')) {
-        getCurrentInvocationData().reqRes.push(
-          JSON.parse(message.slice(message.lastIndexOf('###') + 3))
-        );
+        getCurrentInvocationData().reqRes.push(message.slice(message.lastIndexOf('###') + 3));
```
medikoo, line 225: How it is written to logs now? Is this base64 encoded payload?

Danwakeem: Yeah exactly 👍 when the proto data is written into the logs it gets automatically base64 encoded

medikoo: > it gets automatically base64 encoded

How exactly it happens automatically?

Danwakeem: [Right here](https://github.com/serverless/console/blob/83891a2970c21f445596d146fb5719f772c3bc9c/go/packages/dev-mode/agent/forwarder.go#L107). I found that when you convert a set of `[]byte` to `string` it was defaulting to base64 encoding. 

I ran a test on this on AWS with this extension enabled and passing a standard string into the API payload also if I log out the results in the integration tests the data is base64 encoded.

Screen Shot 2022-10-06 at 8 47 58 AM

medikoo: Ok, thanks :) I wanted to be double sure. 

String you present in log is suspiciously short, but I take it's just some test string

medikoo (approved):

node/test/go/dev-mode/test/lib/process-function.js
Danwakeem: Yeah no problem it was good for me to double check 😎 

And you are right that string is suspiciously short because it is a fake proto payload I was using for testing with just some dummy text in it 😎

PR #238, opened 2022-10-04
chore: Updated sls tags to include additional identifying information

Danwakeem:
## Description
Per the discussion in [this PR](https://github.com/serverless/console/pull/237) I am adding account id and request id to the slsTags so that we can include accountId, requestId, and region on each trace and req/res proto payloads.

I put this in `slsTags` rather than directly on the root payload since this is specific to processing the payloads in our ingest pipeline 🤷


medikoo (approved):
👍

Danwakeem:
Lol I didn't mean to merge this 🤦 I saw a green button and clicked merge lol I will wait to hear Max's response in [this PR](https://github.com/serverless/console/pull/237) and revert if needed.

PR #239, opened 2022-10-04
Created dev mode proto file

Danwakeem:
## Description
Added a dev mode proto file so that we can add any unique attributes to the dev mode payloads without doing any additional data enrichment in the external extension.

Also reverted the changes from [this PR](https://github.com/serverless/console/pull/238) and this change is in response to the [conversation in this PR](https://github.com/serverless/console/pull/237).

### DIFF
--- a/proto/serverless/instrumentation/v1/dev_mode.proto
+++ b/proto/serverless/instrumentation/v1/dev_mode.proto
@@ -19,8 +19,8 @@ message DevModePayload {
 
   oneof payload {
     // The set of lambda traces that were generated via an internal extension
-    TracePayload traces = 4;
+    TracePayload trace = 4;
     // The req or response data from the instrumented lambda function
-    RequestResponse req_res = 5;
+    RequestResponse request_response = 5;
   }
 }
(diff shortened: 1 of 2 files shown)


proto/serverless/instrumentation/v1/dev_mode.proto
```diff
@@ -0,0 +1,26 @@
...
+  // The AWS Account ID where this payload originated from
+  string account_id = 1;
+  // The AWS Region where this payload originated from
+  string region = 2;
+  // The lambda request id where this payload originated from
+  string request_id = 3;
+
+  oneof payload {
+    // The set of lambda traces that were generated via an internal extension
+    TracePayload traces = 4;
```
medikoo: It's a single trace, so more accurate name would be `trace`

proto/serverless/instrumentation/v1/dev_mode.proto
```diff
@@ -0,0 +1,26 @@
...
+    // The req or response data from the instrumented lambda function
+    RequestResponse req_res = 5;
```
medikoo: I'd use full name `request_response`

Danwakeem: Yeah definitely keeps things consistent 👍

Mmarzex (approved):

PR #241, opened 2022-10-05
feat: Support telemetry API in us-east-1

Danwakeem:
## Description
Optionally use the `/telemetry` endpoint if the function is deployed in `us-east-1`. I am just getting ahead of that change in the current region that supports telemetry 🤷

### DIFF
--- a/go/packages/dev-mode/main_test.go
+++ b/go/packages/dev-mode/main_test.go
@@ -39,7 +39,7 @@ type ValidationResult struct {
 }
 
 var port = 9001
-var region = "us-east-1"
+var region = "us-east-2"
 var functionName = "test-function"
 
 type mockSTSClient struct {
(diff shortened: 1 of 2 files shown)


go/packages/dev-mode/main_test.go
```diff
@@ -39,7 +39,7 @@ type ValidationResult struct {
 }
 
 var port = 9001
-var region = "us-east-1"
+var region = "us-east-2"
```
Danwakeem, line 42: I will switch this back to us-east-1 once the telemetry API is rolled out in all regions 😎

medikoo (commented):
@Danwakeem maybe rebase with master to ensure CI checks go in

Danwakeem:
@medikoo Ahh yeah good call I must have opened this when github actions were down 😢

medikoo (approved):
👍

PR #244, opened 2022-10-06
AWS Lambda SDK: Release v0.8.0

medikoo:


Danwakeem (changes requested):
I was testing out some stuff this morning and I noticed that I was getting this error from the sdk 👇 

```
{
  "errorType": "TypeError",
  "errorMessage": "Cannot read properties of undefined (reading 'push')",
  "stack": [
    "TypeError: Cannot read properties of undefined (reading 'push')",
    "    at sendSpans (/opt/sls-sdk-node/wrapper.js:4821:48)",
    "    at processTicksAndRejections (node:internal/process/task_queues:78:11)"
  ]
}
```

For context, I built the internal sdk manually and loaded it on to a lambda function by replacing the published extension with the one I built manually. I am happy to post any details about the function config if you need it 🙏

medikoo:
> I was testing out some stuff this morning and I noticed that I was getting this error from the sdk 

@Danwakeem are you sure you've tested against version as it is on `main` branch now? Asking because stack trace doesn't seem to confirm on that.

e.g  wrapper.js:4821 looks like that on my side: https://gist.github.com/medikoo/5a2e7a91b977306802473424278a7bd4#file-wrapper-js-L4821 (it doesn't point `push` operation as in your stack)

Also what exactly you were doing, so such error surfaced? Can you add this case to our integration tests, so we have it well exposed (?) - That technically will make the best possible reproduction case to work with

Danwakeem:
@medikoo Yeah I just pulled main and rebuilt the extension to double check and I am seeing a similar error. Given the line number if very large i was assuming it referencing a minified/concatenated version of the extension code so the `push` was happening somewhere within the sdk 🤔 

After doing a bit more testing I am thinking it isn't covered by the tests because it seems to only happen when the `SLS_DEV_MODE_ORG_ID` flag is set. (It does the same with or without the dev mode extension included as a layer 🤔)

I can create a branch and add a test for this 👍

medikoo:
> i was assuming it referencing a  referencing a minified/concatenated version o

We're having concatenation but not minification, so source code is quite readable, and stack traces can be inspected against build easily

> After doing a bit more testing I am thinking it isn't covered by the tests because it seems to only happen when the SLS_DEV_MODE_ORG_ID flag is set.

In unit tests, we have some tests that cover external extension case. So maybe you can check if you can reproduce it over there (?)

Danwakeem:
@medikoo Oh yeah I just unzipped that file and it is fairly easy to read 😎 

It looks like the push error is coming from this part of the code 👇 

https://github.com/serverless/console/blob/ff719dfb6c31fc6dd2b284efffacac0d647b5811/node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js#L35-L37

 I am working on getting the tests set up to verify if things are not working it might take me a minute or two 😅

medikoo:
@Danwakeem thanks, I got it reproduced. It's weird that it's not exposed by unit tests we have. Anyway let me put this release to draft until it's fixed.

Great you've coined that!

Danwakeem:
Awesome 😎 Glad I could help!

Danwakeem (approved):
This is looking good 👍 I did one of the same tests on with the dev mode extension and the sdk extension this morning and I am seeing spans flow through the websocket 🎉

PR #245, opened 2022-10-06
Converted payloads to dev mode paylods

Danwakeem:
## Description
Updated payloads being sent to the API so that they follow the new dev mode proto format


go/packages/dev-mode/lib/logger.go
```diff
@@ -80,23 +81,23 @@ func ReportLog(logPayload string) {
...
+		BaseLogger.Printf("⚡ DEV-MODE: Log###%s", base64.StdEncoding.EncodeToString([]byte(logPayload)))
 	}
 }
 
 func ReportReqRes(logPayload string) {
 	_, ok := os.LookupEnv("SLS_DEBUG_EXTENSION")
 	_, toLogs := os.LookupEnv("SLS_TEST_EXTENSION_LOG")
 	if ok || toLogs {
-		BaseLogger.Printf("⚡ DEV-MODE: ReqRes###%s", logPayload)
+		BaseLogger.Printf("⚡ DEV-MODE: ReqRes###%s", base64.StdEncoding.EncodeToString([]byte(logPayload)))
```
Danwakeem, line 92: I needed to change these since I converted these to the dev mode proto format. Before I was simply forwarding the payloads so it changed `[]byte` to `string` conversion function that was producing the base64 string to the proto string method 👍

medikoo (approved):
👍

PR #247, opened 2022-10-07
AWS Lambda SDK: Fix dev mode support and improve its testing

medikoo:
Fix handling of telemetry requests with dev mode enabled

Additionally:
- Introduce dummy external extension to take payloads from internal, to enable better testing of dev mode.
- Update benchmarks to cover also external extension case and ensure performance check is run in CI

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -110,7 +110,6 @@ module.exports = (originalHandler, options = {}) => {
     let isResolved = false;
     let responseStartTime;
     const invocationId = ++currentInvocationId;
-    serverlessSdk._deferredTelemetryRequests = [];
     if (invocationId > 1) {
       // Reset root span ids and startTime with every next invocation
       delete awsLambdaSpan.traceId;
@@ -161,6 +160,7 @@ module.exports = (originalHandler, options = {}) => {
       reportTrace();
       flushSpans();
       await Promise.all(serverlessSdk._deferredTelemetryRequests);
+      serverlessSdk._deferredTelemetryRequests.length = 0;
       serverlessSdk._debugLog(
         'Overhead duration: Internal response:',
         `${Math.round(Number(process.hrtime.bigint() - responseStartTime) / 1000000)}ms`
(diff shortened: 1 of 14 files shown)


node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -161,6 +160,7 @@ module.exports = (originalHandler, options = {}) => {
       reportTrace();
       flushSpans();
       await Promise.all(serverlessSdk._deferredTelemetryRequests);
+      serverlessSdk._deferredTelemetryRequests.length = 0;
```
Danwakeem, line 163: TIL that you can set an array length to 0 to clear the array contents 🤯

node/packages/aws-lambda-sdk/test/fixtures/dev-mode-extension/extensions/dummy-dev-extension.js
```diff
@@ -0,0 +1,267 @@
+#!/usr/bin/env node
```
Danwakeem, line 1: I am curious why we wouldn't just use the actual dev mode extension?

Danwakeem (commented):
Looking good 👍 just one question about the mock external extension

node/packages/aws-lambda-sdk/test/fixtures/dev-mode-extension/extensions/dummy-dev-extension.js
medikoo: This is a no-op extension, through which we can isolate any side-effects as created by the internal extension. It's meant to expose issues in internal extension only. If I would introduce real external extension here its issues may create false positives.

Still, totally we should also have full integration test, that confirms that external extension sends expected payloads  but I think it's more in the scope of the external extension package (although it be nice to have such tests run also on any changes in this extension)

Danwakeem (approved):
This is looking good 👍

node/packages/aws-lambda-sdk/test/fixtures/dev-mode-extension/extensions/dummy-dev-extension.js
Danwakeem: Okay yeah this makes sense 👍

PR #240, opened 2022-10-05
AWS Lambda SDK: Report closed spans on the go and other improvements

medikoo:
- Report closed spans on the go
- In dev mode:
  - Report AWS SDK request and response bodies
  - Report HTTP(S) request and response bodies (added prevention to not report bodies that are larger than 10MB - configurable value)
- Do not write request and response bodies to CW logs

PR #248, opened 2022-10-10
AWS Lambda SDK: Ensure  opt-out for dev mode request and response monitoring

medikoo:
Should have been a part of #240 but was overlooked


Danwakeem (approved):
Looking good 👍

PR #250, opened 2022-10-10
Optional telemetry data on dev mode proto

Danwakeem:
## Description
Adding a place on the dev mode proto payload so we can add some supporting data from the lambda telemetry API.

### DIFF
--- a/proto/serverless/instrumentation/v1/dev_mode.proto
+++ b/proto/serverless/instrumentation/v1/dev_mode.proto
@@ -17,10 +17,24 @@ message DevModePayload {
   // The lambda request id where this payload originated from
   string request_id = 3;
 
+  // Extracted Lambda Telemetry API data
+  optional LambdaTelemetry telemetry = 6;
+
   oneof payload {
     // The set of lambda traces that were generated via an internal extension
     TracePayload trace = 4;
     // The req or response data from the instrumented lambda function
     RequestResponse request_response = 5;
   }
+}
+
+// Lambda Telemetry API data. This data is only available for lambda functions that
+// have access to the telemetry API so it will not be included in all regions.
+message LambdaTelemetry {
+  // Init duration as reported by the metrics on the platform.initReport event
+  optional uint32 init_duration = 1;
+  // Internal runtime duration as reported by the metrics on the platform.runtimeDone event
+  optional uint32 internal_duration = 2;
+  // Internal runtime duration as reported by the responseLatency span on the platform.runtimeDone event
+  optional uint32 internal_response_latency = 3;
 }
--- a/proto/serverless/instrumentation/v1/dev_mode.proto
+++ b/proto/serverless/instrumentation/v1/dev_mode.proto
@@ -31,10 +31,10 @@ message DevModePayload {
 // Lambda Telemetry API data. This data is only available for lambda functions that
 // have access to the telemetry API so it will not be included in all regions.
 message LambdaTelemetry {
-  // Init duration as reported by the metrics on the platform.initReport event
-  optional uint32 init_duration = 1;
-  // Internal runtime duration as reported by the metrics on the platform.runtimeDone event
-  optional uint32 internal_duration = 2;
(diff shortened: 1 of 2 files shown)


proto/serverless/instrumentation/v1/dev_mode.proto
```diff
@@ -17,10 +17,24 @@ message DevModePayload {
...
   }
 }
+
+// Lambda Telemetry API data. This data is only available for lambda functions that
+// have access to the telemetry API so it will not be included in all regions.
+message LambdaTelemetry {
+  // Init duration as reported by the metrics on the platform.initReport event
+  optional uint32 init_duration = 1;
+  // Internal runtime duration as reported by the metrics on the platform.runtimeDone event
+  optional uint32 internal_duration = 2;
```
medikoo: Is this reported by new telemetry API? (At least I don't see such duration reported in current version)

Danwakeem: Now that you are asking I am questioning it if I am right, but I thought that is what the `metrics.durationMs` was supposed to be 🤔 

Screen Shot 2022-10-10 at 3 53 19 PM

medikoo: Ok, I take it's available only with new telemetry API, at least my records of `platform.runtimeDone` look as:

```json
{
	"time": "2022-05-04T13:29:32.659Z",
	"type": "platform.runtimeDone",
	"record": {
		"requestId": "5f48a34d-ab8e-455c-aabd-818bc6fe1d31",
		"status": "success"
	}
}
```

Danwakeem: Ahh yes that is correct. This is telemetry API only. 👍 This PR is specifically meant to add a place in our `DevMode.proto` file where we can drop in some of the new telemetry API data.

medikoo: Thanks for clarification 👍

proto/serverless/instrumentation/v1/dev_mode.proto
```diff
@@ -17,10 +17,24 @@ message DevModePayload {
...
     // The req or response data from the instrumented lambda function
     RequestResponse request_response = 5;
   }
 }
+
+// Lambda Telemetry API data. This data is only available for lambda functions that
+// have access to the telemetry API so it will not be included in all regions.
+message LambdaTelemetry {
+  // Init duration as reported by the metrics on the platform.initReport event
+  optional uint32 init_duration = 1;
```
medikoo: Duration is in what units, ms?

It'll be good to document that, and maybe ensure the unit is reflected in name (e.g. `init_duration_ms`)

Danwakeem: Yeah I like this idea 👍 I updated both the description and the variable names

proto/serverless/instrumentation/v1/dev_mode.proto
```diff
@@ -17,10 +17,24 @@ message DevModePayload {
...
   }
 }
+
+// Lambda Telemetry API data. This data is only available for lambda functions that
+// have access to the telemetry API so it will not be included in all regions.
+message LambdaTelemetry {
+  // Init duration in milliseconds as reported by the metrics on the platform.initReport event
+  optional uint32 init_duration_ms = 1;
+  // Internal runtime duration in milliseconds as reported by the metrics on the platform.runtimeDone event
+  optional uint32 internal_duration_ms = 2;
```
medikoo: Wouldn't it be better to follow AWS naming, and name it `runtime_duration_ms` (`internal` suggests to me it's some AWS internal processing and not necessarily handler work)

Danwakeem: Yeah good point 👍 I was thinking internal since we were using that phrase in other places but I think changing to runtime could avoid some confusion.

medikoo (approved):
👍

PR #251, opened 2022-10-10
Added arm64 build and deploy

Danwakeem:
## Description
This PR adds a second `arm64` deploy target for this extension. I also changed the registry name to include the architecture so if they ever add other architectures we can include additional targets 👍

### DIFF
--- a/.github/workflows/dev-mode-layer-publish.yml
+++ b/.github/workflows/dev-mode-layer-publish.yml
@@ -5,7 +5,7 @@ on:
       - "go/sls-external-extension@[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
-  publishNewVersion_x86:
+  publishNewVersion:
     name: Publish new version
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/dev-mode-layer-publish.yml
```diff
@@ -66,6 +66,71 @@ jobs:
...
+          registry-url: https://registry.npmjs.org
+      - name: Install main project dependencies
+        if: steps.cacheNodeModules.outputs.cache-hit != 'true'
+        run: |
+          cd node
+          npm update --save-dev --no-save
+      - name: Build layer for arm64
+        run: |
+          cd go/packages/dev-mode
+          make build_arm64
```
medikoo, line 76: Can't we add arm64 publish step into previous jobs? If I see correctly most of the steps are repeated, and we could just add arm64 steps there on top (?)

Danwakeem: Yeah we could do that 👍

medikoo (approved):
👍

PR #255, opened 2022-10-11
Release v0.9.0

medikoo:
Additionally bumped restriction in performance test, as it crashed with `AssertionError: expected 301.64 to be below 300` for init time with external extension


Danwakeem (approved):
👍

PR #257, opened 2022-10-11
feat: Added multi-region endpoints

Danwakeem:
## Description
Refactoring how we get the publish URL in dev mode so that we are always using an endpoint as close as possible to the customers lambda function. [This will significantly reduce the total duration](https://docs.google.com/spreadsheets/d/1aW8Th0uDVi7uh_utCfSLO9FmO8rT8bV9StRfcDTpG5o/edit#gid=0) of the customers lambda function when using this extension.

### DIFF
--- a/go/packages/dev-mode/README.md
+++ b/go/packages/dev-mode/README.md
@@ -34,3 +34,8 @@ Integrations tests for this extension are using the integration testing setup/fr
 
 ## Releasing a version
 When you are ready to release a new version of this extension simply increment the version number in the `version.txt` file.
+
+## Regional Endpoints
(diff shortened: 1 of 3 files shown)


Danwakeem:
I will probably wait for https://github.com/serverlessinc/platform-core/pull/1135 to be merged before I merge this or whatever 😎

Mmarzex (dismissed):
This is awesome 🚀 !

If it isn't, let's make sure we document that supporting a new region means deploying the regional endpoint as well as the extension.

medikoo (dismissed):
👍

Danwakeem:
@Mmarzex Yeah that is a good call 👍 I will be sure to document that in the README for this project and in the backend README's as well

PR #261, opened 2022-10-14
AWS Lambda SDK: Do not trace DynamoDb request `expressionValues`

medikoo:
It may contain sensitive information which we do not expose in all cases.

While in dev mode users will still have it in request response data


Danwakeem (approved):
This is a good call 👍

PR #262, opened 2022-10-14
bug: Missing append logs

Danwakeem:
## Description
I had originally meant to do an append here instead of creating a new slice. Not a huge deal but I think it will be nice to just continue using the agent.ForwardLogs call on line 148 👍


medikoo (approved):
👍

PR #263, opened 2022-10-17
Release v0.9.1

medikoo:
Additionally, fixed test, which should have been updated with #261 and which caused recent CI failure


Danwakeem (approved):
🤦

PR #268, opened 2022-10-20
Fix cloudformation workflows & use validation specific IAM role for actions

Mmarzex:
Fixed some issues for the CFT sync stacks as well as introduced a second IAM role to assume for the validation workflow that can be safely used in PR context.


Danwakeem (approved):
😎

medikoo (approved):
👍

PR #269, opened 2022-10-20
SDK Schema: Move span `input` and `output` data to top level properties

medikoo:
BREAKING CHANGE:
- `http.request_body`, `http.response_body`, `aws.sdk.request_body` and `aws.sdk.response_body` tags are removed in favor of top level `input and `output` properties


Danwakeem (approved):
Looking good 👍

PR #272, opened 2022-10-20
AWS Lambda SDK: Adapt `span.input` and `span.output`

medikoo:


node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js
```diff
@@ -27,9 +26,8 @@ const sendSpans = () => {
     },
     spans: pendingSpans.map((span) => {
       const result = span.toProtobufObject();
-      for (const tagName of span.tags.keys()) {
-        if (devModeOnlyTags.has(tagName)) span.tags.delete(tagName);
-      }
+      span.input = null;
```
Danwakeem, line 29: Since you are returning `result` wouldnt we need to null the `result` object attributes? 🤔

Danwakeem (commented):
Looking good just one small question

node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js
medikoo: `result` is for dev mode. What's left on `span` goes to CW logs. So this logic, retrieves full data for dev mode, and clears data we do not want to be logged in CW

Danwakeem: Ohh okay I understand. Thank you for explaining 🙏

Danwakeem (approved):
:shipit:

PR #273, opened 2022-10-20
AWS Lambda SDK: Release v0.10.0

medikoo:


Danwakeem (approved):
:shipit:

Mmarzex:
@Danwakeem Should we get a new sdk-schema version out with your changes and then publish this so we don't need to publish twice today?

medikoo:
@Mmarzex @Danwakeem should I hold with merging this?

Danwakeem:
@Mmarzex @medikoo I don't think so 🤔 the changes to the schema only appear to be affect the ingestion pipeline when we are enriching the data coming from the SDK/cloudwatch/external extension.

Mmarzex:
Gotcha Sounds good, wasn't sure if we'd get out of sync.

PR #275, opened 2022-10-21
Update external

Danwakeem:
## Description
Updating the external extension to support the new protobuf schemas 😎


go/packages/dev-mode/agent/forwarder.go
```diff
@@ -100,6 +110,48 @@ func FindRuntimeDone(logs []LogItem) *LogItem {
...
+		}
+		levelNumber, ok := LogLevels[workingLevel]
+		if ok {
+			return *body, workingLevel, levelNumber
+		}
+	}
+	//TODO: try other log formats
+
+	return *body, level, LogLevels[level]
+}
```
Danwakeem, line 153: I don't think I actually have to do this here let me double check 🤔

Mmarzex (approved):

go/packages/dev-mode/agent/forwarder.go
Danwakeem: We don't necessarily need this but I am going to leave it incase we ever decide we want to support publishing to a customers system or something. This will prevent the payloads leaving this service from being incorrect. It may be worth publishing this logic to a go package or something we so can share it between our backend/extension but that can be done later I guess 🤷

PR #280, opened 2022-10-24
feat: Added timestamp to req/res payloads

Danwakeem:
## Description
Adding a timestamp value to the req/res payloads so we will need to know when these payloads were generated

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -8,6 +8,7 @@ const isPlainObject = require('type/plain-object/is');
 const coerceToString = require('type/string/coerce');
 const traceProto = require('@serverless/sdk-schema/dist/trace');
 const requestResponseProto = require('@serverless/sdk-schema/dist/request_response');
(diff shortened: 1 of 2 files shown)


node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -20,7 +21,17 @@ const { traceSpans } = serverlessSdk;
...
-const reportRequest = async (event, context) => {
+const resolveEpochTimestampString = (() => {
+  const diff = BigInt(Date.now()) * BigInt(1000000) - process.hrtime.bigint();
+  return (uptimeTimestamp) => String(uptimeTimestamp + diff);
+})();
+
+const toLong = (value) => {
+  const data = Long.fromString(String(value));
+  return new Long(data.low, data.high, true);
+};
```
Danwakeem, line 32: This is duplicated from the `trace-span/index.js` file. I figured we could refactor to consolidate the two later if we would like. I didn't combine them now because I didn't see a clear way we were sharing code across these files.

Mmarzex (approved):

PR #282, opened 2022-10-25
AWS Lambda SDK: Cleanup and tests improvements

medikoo:
- Cleanup support for `timestamp` on request and response payloads (so no public API is affected)
- Workaround occasional race condition on AWS side, where deployed endpoints are not available right away (remedy for https://github.com/serverless/console/actions/runs/3291396523/jobs/5425513617)


Danwakeem (approved):
This is looking great 👍

PR #285, opened 2022-10-31
AWS Lambda SDK: Fix capture of HTTP response body when it's a string

medikoo:
Fix to issue reported internally by @Danwakeem 

Apparently when issuing HTTP request via `http.request` API, the obtained response may emit data chunks in plain string format. So far implementation assumed it'll always be a buffer.

At this point this comes with no regression test, as it's unclear in what circumstances _string_ is propagated. Once we clarify that out, I'll prepare a corresponding test.


Danwakeem (approved):
This looks good and I tested it on the app that I noticed this issue with and everything seems to be working fine 😎

PR #288, opened 2022-10-31
SDK Schema: Support Node.js v12

medikoo:
We've learned that AWS Lambda until May 2023 will support Node.js v12 runtime.
 
To enable such support on our extension side. The generated SDK schemas need to work in Node.js v12


Danwakeem (approved):
💪

PR #297, opened 2022-11-01
gettng started improvements

hev:


hev:
Could be better still but reducing headers is a start.

PR #305, opened 2022-11-08
AWS Lambda SDK: Improve SDK error messaging

medikoo:
"Error" suggests lambda failure, while in those cases we do not influence the Lambda process. 

Replace "Error" term with "Warning" to not alarm users there's something wrong with their lambdas

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/lib/send-telemetry.js
+++ b/node/packages/aws-lambda-sdk/instrument/lib/send-telemetry.js
@@ -33,7 +33,7 @@ module.exports = async (name, body) => {
         (response) => {
           if (response.statusCode !== 200) {
             process.stderr.write(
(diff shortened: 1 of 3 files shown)


Danwakeem (commented):
Looking good 👍 
Would this one also need to change from Error to Warning? 👇 
https://github.com/serverless/console/blob/65e9bbab589e31026ff9768905a449e556b3836f/node/packages/aws-lambda-sdk/instrument/index.js#L103

medikoo:
> Would this one also need to change from Error to Warning? 👇

This one is an actual error we're throwing. It happens if someone instead of injecting via layer, attempts to require our SDK manually and doesn't provide the necessary arguments

Danwakeem:
Ahh okay that makes sense. Thanks for the clarification 😎

Danwakeem (approved):
✅

PR #308, opened 2022-11-16
bug: Response uses platform.runtimeDone timestamp

Danwakeem:
## Description
Updated the response payload to use the `platform.runtimeDone` timestamp instead of the one generated by the internal SDK.


medikoo (approved):
👍

PR #315, opened 2022-11-29
Add Event Payload Definition

Mmarzex:
See SC-142 in Linear.

### DIFF
--- a/proto/serverless/instrumentation/tags/v1/error.proto
+++ b/proto/serverless/instrumentation/tags/v1/error.proto
@@ -16,25 +16,11 @@ message ErrorTags {
         // No ErrorType was provided. This should never be the case and if it is received ingest will ignore it.
         ERROR_TYPE_UNSPECIFIED = 0;
 
-        // An Uncaught exception was raised that interupts execution unexpectedly. 
+        // An unexpected error that caused the application to fail
         ERROR_TYPE_UNCAUGHT = 1;
        
-        // An Uncaught exception, specifically a timeout of the compute environment,
-        // A timeout in Lambda, for example, is registered when ingest receives a log line
-        // stating that the task timed out.
-        ERROR_TYPE_UNCAUGHT_TIMEOUT = 2;
-
-        // An Uncaught exception, specifically the compute environment ran out of memory.
-        ERROR_TYPE_UNCAUGHT_OOM = 3;
-
-        // An uncaught exception that takes place during the AWS Lambda Initialization phase.
-        ERROR_TYPE_UNCAUGHT_AWS_LAMBDA_INIT = 4;
-
-        // An uncaught exception that takes place during the AWS Lambda Invocation phase.
-        ERROR_TYPE_UNCAUGHT_AWS_LAMBDA_RESOLVED = 5;
-
         // An error that was explicitly caught with the Serverless SDK
-        ERROR_TYPE_CAUGHT = 6;
+        ERROR_TYPE_CAUGHT = 2;
 
     }
 
--- a/proto/serverless/instrumentation/tags/v1/error.proto
+++ b/proto/serverless/instrumentation/tags/v1/error.proto
@@ -13,18 +13,27 @@ message ErrorTags {
     optional string stacktrace = 3;
 
     enum ErrorType {
+        // No ErrorType was provided. This should never be the case and if it is received ingest will ignore it.
         ERROR_TYPE_UNSPECIFIED = 0;
 
+        // An Uncaught exception was raised that interupts execution unexpectedly. 
         ERROR_TYPE_UNCAUGHT = 1;
-        
+       
+        // An Uncaught exception, specifically a timeout of the compute environment,
+        // A timeout in Lambda, for example, is registered when ingest receives a log line
+        // stating that the task timed out.
         ERROR_TYPE_UNCAUGHT_TIMEOUT = 2;
 
+        // An Uncaught exception, specifically the compute environment ran out of memory.
         ERROR_TYPE_UNCAUGHT_OOM = 3;
 
+        // An uncaught exception that takes place during the AWS Lambda Initialization phase.
         ERROR_TYPE_UNCAUGHT_AWS_LAMBDA_INIT = 4;
 
+        // An uncaught exception that takes place during the AWS Lambda Invocation phase.
         ERROR_TYPE_UNCAUGHT_AWS_LAMBDA_RESOLVED = 5;
 
+        // An error that was explicitly caught with the Serverless SDK
         ERROR_TYPE_CAUGHT = 6;
 
     }
--- a/proto/serverless/instrumentation/tags/v1/error.proto
+++ b/proto/serverless/instrumentation/tags/v1/error.proto
@@ -1,0 +1,33 @@
+syntax = "proto3";
+
+package serverless.instrumentation.tags.v1;
+
+option go_package = ".;protoc";
+
+message ErrorTags {
+    // The Error Name
+    string name = 1;
+    // The Error Message - Depending on runtime this is defined or not.
+    optional string message = 2;
+    // The Error stacktrace if applicable
+    optional string stacktrace = 3;
+
+    enum ErrorType {
+        ERROR_TYPE_UNSPECIFIED = 0;
+
+        ERROR_TYPE_UNCAUGHT = 1;
+        
+        ERROR_TYPE_UNCAUGHT_TIMEOUT = 2;
+
+        ERROR_TYPE_UNCAUGHT_OOM = 3;
+
+        ERROR_TYPE_UNCAUGHT_AWS_LAMBDA_INIT = 4;
+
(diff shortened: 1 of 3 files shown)


proto/serverless/instrumentation/tags/v1/error.proto
```diff
@@ -0,0 +1,33 @@
...
+message ErrorTags {
+    // The Error Name
+    string name = 1;
+    // The Error Message - Depending on runtime this is defined or not.
+    optional string message = 2;
+    // The Error stacktrace if applicable
+    optional string stacktrace = 3;
+
+    enum ErrorType {
+        ERROR_TYPE_UNSPECIFIED = 0;
```
austencollins, line 17: Let's put comments on what these mean in here. Unspecified looks new, and I haven't yet seen any docs on it.

Mmarzex: > Let's put comments on what these mean in here. Unspecified looks new, and I haven't yet seen any docs on it.

Yeah, my bad, I wrote it up but didn't push the commit out. Will get it updated. A note on `UNSPECIFIED` it is a requirement of enum types in protobuf. It functionally will never be seen in our data as our internal validation will reject that type, but it is needed so the type and client generation from the protobufs do not fail.

proto/serverless/instrumentation/tags/v1/error.proto
```diff
@@ -0,0 +1,42 @@
...
+       
+        // An Uncaught exception, specifically a timeout of the compute environment,
+        // A timeout in Lambda, for example, is registered when ingest receives a log line
+        // stating that the task timed out.
+        ERROR_TYPE_UNCAUGHT_TIMEOUT = 2;
+
+        // An Uncaught exception, specifically the compute environment ran out of memory.
+        ERROR_TYPE_UNCAUGHT_OOM = 3;
+
+        // An uncaught exception that takes place during the AWS Lambda Initialization phase.
```
medikoo: @Mmarzex Does it cover both (1) handler not found and (2) handler crashed at initialization?

Mmarzex: @medikoo Is that the intention we have for it? I'm fine with it covering both.

proto/serverless/instrumentation/tags/v1/error.proto
```diff
@@ -0,0 +1,42 @@
...
+    optional string message = 2;
+    // The Error stacktrace if applicable
+    optional string stacktrace = 3;
+
+    enum ErrorType {
+        // No ErrorType was provided. This should never be the case and if it is received ingest will ignore it.
+        ERROR_TYPE_UNSPECIFIED = 0;
+
+        // An Uncaught exception was raised that interupts execution unexpectedly. 
+        ERROR_TYPE_UNCAUGHT = 1;
```
medikoo, line 20: What exactly type of error this covers? This type code is very generic, and the description seems not specific

proto/serverless/instrumentation/tags/v1/error.proto
medikoo: Originaly intention was to handle just _handler not found_ errors here, and eventually split one responsible for uncaught exceptions or unhandled rejections to one that covers invocation and other that covers initialization

proto/serverless/instrumentation/tags/v1/error.proto
Mmarzex: I think you are thinking too much in lambda specific cases here. I don't see a scenario where a generic `uncaught` will happen too often in Lambda, but a generic `uncaught` will be needed for other compute environments.

proto/serverless/instrumentation/tags/v1/error.proto
Mmarzex: > Originaly intention was to handle just handler not found errors here, and eventually split one responsible for uncaught exceptions or unhandled rejections to one that covers invocation and other that covers initialization

Then propose those types please @medikoo. We talked about this in the meeting on monday and these were the types we all agreed upon. I'm fine with whatever types we want here, but I'm going off the spec so we need to update that if needed.

proto/serverless/instrumentation/tags/v1/error.proto
Mmarzex: I think in general too, we have made decisions on all of these types already from a product spec. I don't want us getting bogged down in this PR when we have already decided on these names.

medikoo: @Mmarzex product spec provides just a few loose examples that puts on us onto agreed direction. In my understanding, final spec details are being worked out here

`ERROR_TYPE_UNCAUGHT` looks to me as some _catch all_ type, to cover any errors that do not fit other 
`ERROR_TYPE_UNCAUGHT_*` categories. That feels too loose IMO. I'm afraid introduce such slot, may bite us back in a future.

I propose to go as follows:

```
        // No ErrorType was provided. This should never be the case and if it is received ingest will ignore it.
        ERROR_TYPE_UNSPECIFIED = 0;

        // An Uncaught runtime exception in a working app
        // In AWS Lambda case it covers uncaught exceptions and unhandled rejections
        // which happen either during the initialization phase or invocation phase
        ERROR_TYPE_UNCAUGHT_APP = 1;
       
        // An Uncaught exception, specifically a timeout of the compute environment,
        // A timeout in Lambda, for example, is registered when ingest receives a log line
        // stating that the task timed out.
        ERROR_TYPE_UNCAUGHT_TIMEOUT = 2;

        // An Uncaught exception, specifically the compute environment ran out of memory.
        ERROR_TYPE_UNCAUGHT_OOM = 3;

        // AWS Lambda Initialization startup error
        // Happens when either no configured module for handler is found
        // or no named function handler is found on exported lambda module
        ERROR_TYPE_UNCAUGHT_AWS_LAMBDA_INIT = 4;

        // AWS Lambda Invocation resolution error.
        // Error was either passed to handler callback or is a rejection resolution of async lambda handler
        ERROR_TYPE_UNCAUGHT_AWS_LAMBDA_RESOLVED = 5;

        // Intentionally captured and reported error through the SDK
        // Given error does not stop application from working and doesn't become its result
        ERROR_TYPE_CAUGHT = 6;
```

Mmarzex: > `ERROR_TYPE_UNCAUGHT` looks to me as some catch all type,

That is the intention. We need a catch all type for things that aren't defined. I do not see strong reason to not have that in general we want to classify an `uncaught` error using its name and/or stacktrace. These special types are really for cases like a timeout or a handler issue where we wouldn't have robust information like a stack trace. 

I'm not seeing a strong reason not to leave it as `ERROR_TYPE_UNCAUGHT`, so it is still the best phrase to use. `ERROR_TYPE_UNCAUGHT_APP`, assumes it is in fact an app and again seems too Lambda specific to me. We need to have a generic error type that we can use in all other platforms. 

I'm fine with the additional context you added in those comments @medikoo, except for `caught` I think that phrasing is a bit too Lambda specific, so I am going to update that.

Mmarzex: Extending further, if we did not have Lambda as a primary integration in our system we would only need an error type of `uncaught` or `caught` so while `ERROR_TYPE_UNCAUGHT` may seem ambiguous to you, I would argue the other error types are too specific, hence why we are adding Lambda as part of the name. When we talk about needing to support the Lambda errors that come from Cloudwatch I think it makes sense to add these types there, but for most other scenarios, I do not see a reason to try and create an error type for every possible uncaught exception.

Also it is important to keep in mind, these types need to be surface-able to users, surfacing `uncaught_app` is confusing and not intuitive where as surfacing `uncaught` vs. `caught` conveys that one is very bad while the other is potentially not.

proto/serverless/instrumentation/tags/v1/error.proto
```diff
@@ -0,0 +1,28 @@
...
+    optional string stacktrace = 3;
+
+    enum ErrorType {
+        // No ErrorType was provided. This should never be the case and if it is received ingest will ignore it.
+        ERROR_TYPE_UNSPECIFIED = 0;
+
+        // An unexpected error that caused the application to fail
+        ERROR_TYPE_UNCAUGHT = 1;
+       
+        // An error that was explicitly caught with the Serverless SDK
```
medikoo: I'd clarify that a bit more deeply:

```
An error that was reported via the Serverless SDK.
Error that doesn't explicitly fail the application.
Multiple errors of this type can be reported during a single application run
```

proto/serverless/instrumentation/tags/v1/error.proto
```diff
@@ -0,0 +1,28 @@
...
+        ERROR_TYPE_CAUGHT = 2;
+
+    }
+
+    ErrorType type = 4;
+}
```
medikoo: Let's add EOL at EOF

medikoo (approved):
👍 Looks great to me

PR #316, opened 2022-12-02
chore: Release v0.13.2

Mmarzex:


medikoo (commented):
I think it should be v0.13.2, as there were no breaking changes

Mmarzex:
@medikoo I'm fine with that. The only reason I incremented to v0.14 is since it introduces a new payload type it seems like a significant enough change. This might be something worth creating a guideline for.

medikoo:
@Mmarzex we're following semver and there are strict rules. In experimental phase (v0.x), minor is increased only in case of breaking changes, and patch in all other cases.

Once we switch to v1, it's _major_ that will be increased only in case of breaking change, _minor_ in case of added feature and _patch_ in case of maintenance or bug fix.

If you want, we may switch to v1 now, consider it stable and rely on semver fully

Mmarzex:
> we're following semver and there are strict rules. In experimental phase (v0.x), minor is increased only in case of breaking changes

I'll switch it to 0.13.2 for now. I'd argue any addition of a new schema warrants more than a patch version change. Even in the case of v0

Mmarzex:
> If you want, we may switch to v1 now, consider it stable and rely on semver fully

I don't think we are at a place to say `v1` just yet. I think we need to have the full schema for the next few features that are in the roadmap (errors and alerts specifically) and then we should set everything to `v1`. I'd consider that point would put as in a place of being pretty confident none of the current structures will have major changes to them.

medikoo (approved):

PR #317, opened 2022-12-02
Add Events as a property of TracePayload

Mmarzex:


.github/workflows/node-validate.yml
```diff
@@ -252,11 +252,14 @@ jobs:
       - name: Buf Setup
         if: steps.pathChanges.outputs.sdkSchema == 'true' || steps.pathChanges.outputs.proto == 'true'
         uses: bufbuild/buf-setup-action@v1
+        with:
+          github_token: ${{ github.token }}
```
medikoo, line 256: Why these changes are needed?

Mmarzex: `buf` is rate limited. So if you pass in the github_token it uses that to authenticate/reset the rate limit. This is the safer method then using one of their authentication tokens in a PR workflow as that token would allow pushes into our buf repo which we do not want.

medikoo (approved):
👍

PR #318, opened 2022-12-07
chore: Release sdk-schema v0.14.0

Mmarzex:


medikoo (commented):
I was just preparing the same PR :)

I think, in this case, we need to do a _major_ bump (so v0.14)

Problem is that introducing _a repeated_ field breaks SDK. As the way `ts-proto` works it now requires passing an _array_ unconditionally for this field (it's not treated as optional).
This is an issue on the `ts-proto` side (I've commented about that at https://github.com/stephenh/ts-proto/issues/225#issuecomment-1341088228)

Theoretically, we can relax that by adding the `useOptionals=all` option to `ts-proto` but while it'll prevent the application code from breaking, our tests will still fail (as we compare input and output, and output for not provided _repeated_ fields will list empty array). So we can skip on that, and just treat this as _breaking change_.

Mmarzex:
That is frustrating. I'm fine to just bump it to v0.14. I'll make the change now.

> Theoretically, we can relax that by adding the useOptionals=all option to ts-proto but while it'll prevent the application code from breaking, our tests will still fail 

I agree @medikoo , I don't think we should go this route.

node/packages/sdk-schema/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [0.14.0](https://github.com/serverless/console/compare/@serverless/sdk-schema@0.13.2...@serverless/sdk-schema@0.14.0) (2022-12-07)
+
+### Features
```
medikoo, line 11: Let's add also `BREAKING CHANGES` note, stating that _repeated_ field was added to Trace payload

medikoo (approved):
Looks good.

Just best if it lands as one `chore: ..` commit (`fix` should be strictly about code fixes and it's added to changelog)

PR #320, opened 2022-12-07
doc: Update manual instrumentation example for AWS SDK and Express

Mmarzex:
I noticed yesterday while setting up the AWS SDK v3 instrumentation, the docs didn't match what we expose in the SDK. Updated docs to match.


medikoo (commented):
Indeed great catch. Forgot to update it after internal renames.

CI crashes, I believe due to `doc` scope being used instead of `docs`

Mmarzex:
> CI crashes, I believe due to doc scope being used instead of docs

That was inevitable I mis-typed that eventually 🤦

medikoo (approved):
👍

PR #325, opened 2022-12-14
SDK: Initial version

medikoo:
Environment agnostic `@serverless/sdk` (excluded from `@serverless/aws-lambda-sdk`)

### DIFF
--- a/node/packages/sdk/package.json
+++ b/node/packages/sdk/package.json
@@ -1,0 +1,55 @@
+{
+  "name": "@serverless/sdk",
+  "repository": "serverless/console",
+  "version": "0.0.0",
+  "author": "Serverless, Inc.",
+  "dependencies": {
+    "d": "^1.0.1",
+    "ext": "^1.7.0",
+    "long": "^5.2.1",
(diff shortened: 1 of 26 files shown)


node/packages/sdk/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@serverless/sdk",
```
Mmarzex, line 2: @medikoo Looks like we used the `@serverless/sdk` name already years ago, https://www.npmjs.com/package/@serverless/sdk. 

We will need to change this to something else in order to publish it.

Mmarzex (changes requested):

node/packages/sdk/package.json
medikoo: @Mmarzex yes, I'm aware of that. It's an old project that's not used at all and left abandoned  5 years ago.

It's also our package, and we have rights to publish new versions. I plan to prepare a v0.1.0 release after this PR is taken

Mmarzex (approved):

PR #330, opened 2022-12-19
SDK: `captureError` interface

medikoo:
- Generalize internals that are common for traces and events
- Remove a small bit of AWS Lambda SDK-specific logic that sneaked with the initial release
- Implement `CapturedEvent` interface
- Expose public `serverless.captureError` method


node/packages/sdk/docs/sdk.md
```diff
@@ -23,3 +23,13 @@ Create custom trade span
...
 
 Returns instance of [TraceSpan](trace-span.md)
+
+### `.captureError(error[, options])`
+
+Record captured error
+
+- `error` - Captured error
+- `options`:
+  - `tags` _(bigInt)_ - User tags object. Tag names can contain only lowercase alphanumeric tokens separated with dot. Values can be _string_, _boolean_, _number_, Date or Array containing any values of prior listed types
```
Mmarzex: What's the `_(bigInt)_` referring to here?

node/packages/sdk/lib/tags.js
```diff
@@ -0,0 +1,130 @@
...
+})();
+
+module.exports = class Tags extends Map {
+  set(inputName, value) {
+    const name = ensureTagName(inputName);
+    value = ensureTagValue(value, name);
+    if (this.has(name)) {
+      const currentValue = this.get(name);
+      if (Array.isArray(value)) {
+        if (Array.isArray(currentValue)) {
```
Mmarzex, line 99: Should we maybe throw a different error in this case? I know it is still technically a duplicate tag error but it is also a type mismatch if it throws when it fails the `Array.isArray(currentValue)` check here.

Mmarzex (changes requested):

node/packages/sdk/docs/sdk.md
medikoo: It's supposed to list type of value, and it should be `object` fixed that

node/packages/sdk/lib/tags.js
medikoo: This logic is taken as is from the current version of AWS Lambda SDK (it's not a new addition).

The idea is that we allow to set tag once and throw if there's an attempt to set the same tag again.
Still, to support some race condition scenarios where the same tag might be set with the same value by different parts of logic, I've added handling that, instead of throwing, ignores attempts of setting tag with the same value.

Now, I'm not sure what would be the extra value on producing different errors when the type and value is different. It definitely will increase the complexity of this check. 
Just throwing on different value feels good enough to me. We're addressing a rare scenario here, and the user is well-informed of what happened.

Mmarzex (approved):

PR #333, opened 2022-12-20
feat: Add Warning Tags to Event Schema

Mmarzex:


proto/serverless/instrumentation/tags/v1/warning.proto
```diff
@@ -0,0 +1,12 @@
...
+package serverless.instrumentation.tags.v1;
+
+option go_package = ".;protoc";
+
+message WarningTags {
+    // The Warning Name
+    string name = 1;
+    // The Warning Message
+    optional string message = 2;
+}
```
medikoo: Let's ensure EOL at EOF

proto/serverless/instrumentation/tags/v1/warning.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package serverless.instrumentation.tags.v1;
+
+option go_package = ".;protoc";
+
+message WarningTags {
+    // The Warning Name
+    string name = 1;
```
medikoo: What is expected to be provided with `name` ?

Typical warning registration will look as:

```javascript
serverlessSdk.captureWarning("Alarming behavior");
```

medikoo (commented):
Thanks @Mmarzex

proto/serverless/instrumentation/tags/v1/warning.proto
Mmarzex: I think it comes down to what we are saying is the `name` vs. a potential `message`. In my mind, what you are passing to `captureWarning` is the `name` as that is going to be what you use to filter your warnings in the UI. 

We should keep the naming relatively consistent between errors and warnings, so `name` is the filter used for errors, then it should be the same for warnings.

medikoo: > We should keep the naming relatively consistent between errors and warnings

I though the same. In case of errors, it's like that:

```javascript
serverlessSdk.captureError(new UserError("Cannot do that"));
// { name: "UserError", message: "Cannot do that", stacktrace: ... }

serverlessSdk.captureError("Cannot do that");
// { message: "Cannot do that" }
```

Now in case of warnings, there's no object interface, we accept string, and in my eyes it should translate to _message_ in the same way as we translate _string_ that is provided to `captureError`.
I don't see here a room for `name` as there's no native `Warning` interface, and we also do not plan to introduce one

```javascript
serverlessSdk.captureWarning("Cannot do that");
// { message: "Cannot do that" }
```

Mmarzex: We can change it to optional for now, I could see it being a useful feature to maybe add, it would help facilitate grouping if we want to add that for groupings.

medikoo: Maybe let's add it, once we find a use case for it. At this point I don't see how it can be used

Mmarzex: I disagree, I fully understand the unlikeliness of it being used in the SDK any time soon, but internally we will most likely need a `name` property internally to correctly implement the fingerprinting in a standard way. We want errors and warnings to match relatively similarly so let's keep the spec as close together as possible even if the name isn't used in the user facing part of this.

medikoo: I think adding properties because _maybe_ we will decide to use them is not the right direction. It's a recipe for a bloated schema that hosts confusing, never used properties, of which purpose for new coming developers would be very unclear.

It feels logical to me that we should add here only things we need (when the design of APIs is concerned it's way harder to remove than add, so everything we add should have a clearly defined purpose).

So if you strongly feel it should be added, can you clearly specify in the documentation the purpose of it, and how it'll be actually used (?) If there's a problem with that, let's skip it for now (it can always be added later)

Mmarzex: I don't think I made myself clear. We will need the name property on warnings. We are creating a filter experience around names. To add a second property to filter the message does not make sense in any scenario. 

This schema is a user facing schema but it also must support all of our internal systems. We cannot have normalization in the backend if we do not have the property on the schema, hence the `name` property here.

I will remove the `name` property for now to unblock you, but in the future let's make sure to read the internal tickets related to both the ingest requirements and the SDK requirements so that we both have the correct context when discussing in these PRs.

medikoo (approved):
It feels that problem is the lack of a proper single source of specification. I've taken the further discussion to slack

PR #335, opened 2022-12-21
doc: Use correct URL for `trace-span.md` doc

alexdelorenzo:
This document currently points to `trace-spans.md`, but the filename is `trace-span.md`. This fixes the 404 error given when following the prior link.


medikoo (approved):
Thank you @alexdelorenzo ! Good find

PR #139, opened 2022-07-28
Instrumentation v2 Protobufs

Mmarzex:

PR #340, opened 2023-01-02
ci: Add Publish Go SDK Schema Workflow

Mmarzex:
For right now, coupling this to the publishing of the node sdk-schema. As we add in the Python project we should come up with a way to better couple all the different sdk-schema versions together.

### DIFF
--- a/.github/workflows/go-sdk-schema-publish.yml
+++ b/.github/workflows/go-sdk-schema-publish.yml
@@ -2,8 +2,9 @@ name: "Go: Publish buf.build/serverless/sdk-schema"
 
 on:
   push:
-    tags:
-      - "@serverless/sdk-schema@[0-9]+.[0-9]+.[0-9]+"
+    branches: [main]
+    paths:
+      - proto/**
 
 defaults:
   run:
--- a/.github/workflows/go-sdk-schema-publish.yml
+++ b/.github/workflows/go-sdk-schema-publish.yml
@@ -1,0 +1,33 @@
+name: "Go: Publish buf.build/serverless/sdk-schema"
+
+on:
+  push:
+    tags:
+      - "@serverless/sdk-schema@[0-9]+.[0-9]+.[0-9]+"
+
+defaults:
+  run:
+    working-directory: proto
+jobs:
+  publishNewGoSdkSchemaVersion:
+    name: Publish new version
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.USER_GITHUB_TOKEN }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Setup Buf
+        uses: bufbuild/buf-setup-action@v1
+        with:
+          buf_token: ${{ secrets.BUF_TOKEN }}
+      - name: Buf Lint
+        uses: bufbuild/buf-lint-action@v1
+        with:
+          input: "proto"
+          buf_token: ${{ secrets.BUF_TOKEN }}
+      - name: Buf Push
+        uses: bufbuild/buf-push-action@v1
+        with:
+          buf_token: ${{ secrets.BUF_TOKEN }}
+          input: "proto"


.github/workflows/go-sdk-schema-publish.yml
```diff
@@ -0,0 +1,33 @@
+name: "Go: Publish buf.build/serverless/sdk-schema"
+
+on:
+  push:
+    tags:
+      - "@serverless/sdk-schema@[0-9]+.[0-9]+.[0-9]+"
```
medikoo: This tag relates strongly to `@serverless/sdk-schema` package, which may have a new release not necessarily caused by schema changes.

Also, I'm not sure if it's a good idea to try to enforce the same SDK versions across runtimes. Each runtime has its own specificity and may require interim releases and breaking changes at certain periods (e.g. in Node.js it's the case when want to drop support for the EOL version In such cases bumping major shouldn't force bumping major of e.g. SDK for Python)

Summarizing wouldn't it be better to push to BSR whenever there are changes in proto files pushed to `main`?

Mmarzex: > Also, I'm not sure if it's a good idea to try to enforce the same SDK versions across runtimes.

I strongly disagree on this. If I am using the proto schemas in python, node, and go I should be able to safely assume that `v1` of all three map to the same version of the protobufs. This could be accomplished by publishing versions that change the minor or major version based off of tags of the protobufs and then letting maintenance releases of the individual runtime packages increment the minor or point release. 

> Summarizing wouldn't it be better to push to BSR whenever there are changes in proto files pushed to main?

I am fine to do this, but as soon as we start doing this we are going to have to adhere to stricter protobuf rules. There can be no breaking changes as pushes will fail.

medikoo: > I strongly disagree on this. If I am using the proto schemas in python, node, and go I should be able to safely assume that v1 of all three map to the same version of the protobufs. 

If you refer to `v1` of the protobuf package path.  It is a different versioning area, which, if I remember correctly, agreed as obsolete and kept only due to Go protobuf specificity (https://github.com/serverless/console/pull/139#discussion_r934562811) It doesn't correspond in any way to the versioning of `@serverlress/sdk`, which is at v0.14.

If however, you suggest that if `@serverlress/sdk` is at v0.14, the Python SDK release should also start at v0.14. It'll quickly become very confusing. As in all cases, we will lose semver versioning, as breaking change in any runtime will require a major bump in every runtime.

e.g. in the case of JS adding a _repeated_ field to an already used schema creates a _breaking_ change. While on Protobuf level where _repeated_ fields are optional, it's nothing breaking, and I expect it to work like that for most of the other runtimes.

Also for comparison, AWS SDK, even though it exposes the same API interface doesn't maintain version parity across runtimes. Versioning seems very specific to a given runtime and its ecosystem

> I am fine to do this, but as soon as we start doing this, we are going to have to adhere to stricter protobuf rules. 
> There can be no breaking changes as pushes will fail.

I think we agreed from the beginning not to introduce any breaking changes at all. I've introduced some in the first months, but only because the product was not really used, we can agree it won't happen anymore

Mmarzex: > If however, you suggest that if @serverlress/sdk is at v0.14, the Python SDK release should also start at v0.14.

I don't mean the sdk packages, I'm talking solely the SDK Schema packages, those are what need to maintain the same versions across runtimes. 

> e.g. in the case of JS adding a repeated field to an already used schema creates a breaking change. While on Protobuf level where repeated fields are optional, it's nothing breaking, and I expect it to work like that for most of the other runtimes.


This would be a breaking change in `@serverless/sdk` not `@serverless/sdk-schema`.

Either way this is a discussion we can have as we start to expand our runtime support for the generated schema packages.

I've updated the PR to just push on all mainline changes, let's go forward with that change.

medikoo (approved):
👍

PR #334, opened 2022-12-20
feat: Setup sdk-schema python

Mmarzex:

PR #342, opened 2023-01-04
SDK: SC-256 Initial Python SDK, CI, tests

alexdelorenzo:
This PR includes:
 - SC-265: Create basic SDK container with exports
 - SC-269: Create basic unit tests
 - SC: 266: Create CI validation for `sdk` package

The unit test suite for this PR passes, as does the CI code validation.

This also includes the patch from this PR: https://github.com/serverless/console/pull/334

[You can browse through this branch here](https://github.com/alexdelorenzo/console/tree/sc-256/python/packages/sdk)

[SC-265]: https://serverlessinc.atlassian.net/browse/SC-265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alexdelorenzo:
@medikoo This should be good to go for review :smile:

.github/workflows/python-validate.yml
```diff
@@ -0,0 +1,76 @@
...
+          cache-dependency-path: |
+            **/pyproject.toml
+
+      - name: Install SDK project and dependencies
+        run: |
+          cd python/packages/sdk
+
+          python3 -m pip install --editable .
+
+      - name: Run unit tests
```
medikoo: I'd establish order as: 

1. Formatting
2. Lint
3. Unit tests

So e.g. we crash fast if there are formatting issues which will anyway fail the build, and lint issues may announce test fails, so if that's not green there's probably no point to confirm on unit tests

python/packages/sdk-schema/pyproject.toml
```diff
@@ -0,0 +1,14 @@
...
+[project]
+name = "serverless-sdk-schema"
+version = "0.0.1"
+description = "The protobuf generated Serverless SDK Schema"
+authors = [{ name = "serverlessinc" }]
+dependencies = [
+    "protobuf >= 4.21.12"
+]
+
+requires-python = ">=3.8"
```
medikoo: Let's ensure EOL at EOF

python/packages/sdk/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[build-system]
+requires = ["setuptools>=65.6.3", "wheel>=0.38.4"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "serverless-sdk"
+version = "0.0.1"
```
medikoo: Let's use `0.0.0` (then changing to `0.0.1` should trigger first publication)

.github/workflows/python-validate.yml
```diff
@@ -0,0 +1,76 @@
...
+
+          # ensure project dependencies are cached
+          # When using only `pyproject.toml` for dependencies, see:
+          #  https://github.com/actions/setup-python/issues/529#issuecomment-1367029699
+          cache: 'pip'
+          cache-dependency-path: |
+            **/pyproject.toml
+
+      - name: Install SDK project and dependencies
+        run: |
```
medikoo, line 53: Let's cover SDK (with deps install and unit tests) only if there are changes to `python/packages/sdk` (e.g. in Node.js we guard that with: https://github.com/serverless/console/blob/main/.github/workflows/node-validate.yml#L286

It'll help us to have CI more to the point.

medikoo (commented):
Thanks, @alexdelorenzo, that looks very good 👍 

I see it also introduces `sdk-schema` base, which was not in the scope of this task. Ideally, if we stick precisely to the scope of the task, but it's not a big issue, you've added it here.

I have just a few very minor suggestions, and we should be good to go 💯

alexdelorenzo:
@medikoo Cool, thanks for the feedback. I'll make the required changes and ping you here when done, assuming that's the preferred process 🙂

alexdelorenzo:
@medikoo made the changes and looks good for review 👍

medikoo:
> @medikoo Cool, thanks for the feedback. I'll make the required changes and ping you here when done, assuming that's the preferred process 🙂

Yes 👍 Still in that case you may simply re-request review by clicking a button in reviewers box (I'll get the notification then)

python/packages/sdk-schema/pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[build-system]
+requires = ["setuptools>=65.6.3", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "serverless-sdk-schema"
+version = "0.0.1"
```
medikoo: Same here, let's ensure `0.0.0`

medikoo (commented):
@alexdelorenzo one last thing, and I believe we're ready to go with it.

medikoo (approved):
Thank you @alexdelorenzo !

PR #343, opened 2023-01-04
AWS Lambda SDK: Send captured events to dev mode

medikoo:


node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js
```diff
@@ -15,37 +15,47 @@ const sendTelemetry = require('./send-telemetry');
...
 const pendingSpans = [];
+const pendingCapturedEvents = [];
 let isScheduled = false;
 let timeoutId = null;
-const sendSpans = () => {
+const sendData = () => {
   isScheduled = false;
   clearTimeout(timeoutId);
-  if (!pendingSpans.length) return;
+  if (!pendingSpans.length && !pendingCapturedEvents.length) return;
```
Danwakeem, line 24: Is it possible to have 0 pending spans and 1 or more pending captured events? 🤔

Danwakeem (approved):
This looks good I just had a logistical question 👍

node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js
medikoo: Definitely, in the on going long span, there can be multiple captured events reported. And when they happen, there will be no new spans to report

Danwakeem: @medikoo Ahh okay got it so these would be captured events for a parent span. Maybe it is the long span or some other one 👍 Thanks!

PR #344, opened 2023-01-04
SC-152 Updated initialization error handling

Danwakeem:
## Description
This will attempt for send logs on an initialization error. This will ensure logs are forwarded even if we never receive a requestId or platform event.

There appears to be a bug in the lambda APIs where the following is not happening

- The `platform.runtimeDone` event is not always sent. Sometimes it is, sometimes it isn't. I think this is because the lambda function is shutting down too early.
- The lambda function is shutting down before the extension sends a second `next` event. I think this is the real bug here 😢


medikoo (approved):
👍

PR #345, opened 2023-01-05
AWS Lambda SDK: Initialize original handler at AWS intended processing point

medikoo:
When testing the `console.*` instrumentation, I realized that console patching doesn't pick up as expected. It appears it's because we initialize the SDK and the original handler _before_ the AWS Lambda initialization logic is run. This logic is responsible for patching of some Node.js core modules, including `console` (so messages are prefixed with timestamp and request-id) - so AWS Lambda logic overrides our patching and makes it ineffective.

The reason why we initialize handler _before_  the AWS Lambda initialization is to be able to prevent any instrumentation of the handler in case of its misconfiguration (e.g., not a function exported as a handler), so in such case, we leave the function untouched for AWS Lambda, and the surfaced error is the same whether lambdas are instrumented or not)

However, it introduces more concerning issues. Due to not being instrumented by AWS Lambda `console.*` methods work differently at the initialization phase when the function is instrumented (they're not patched by the AWS and hence not prefixed with the timestamp). Also, some other Node.js modules remain not instrumented at this point which introduces risks of other bugs if the user's code relies on those modules at the initialization phase.

Therefore, the right decision is to ensure that the original function is initialized in the same processing point as when it is not instrumented, which is what this PR fixes.
Thanks to that, all AWS Lambda patching is effective on initialization, and automatically also, our `console.` patching works without the issues.

Concerning invalid handler errors, to mitigate that problem, I've replicated the errors that AWS Lambda throws in such cases, so still user experience should remain the same with this approach.

_I've confirmed the change with all unit and integration tests passing_

### DIFF
--- a/node/packages/aws-lambda-sdk/internal-extension/index.js
+++ b/node/packages/aws-lambda-sdk/internal-extension/index.js
@@ -31,27 +31,6 @@ try {
   const handlerModuleName = path.resolve(handlerModuleDirname, handlerModuleBasename);
 
   const nodeVersionMajor = Number(process.version.split('.')[0].slice(1));
-  const isAsyncLoader = nodeVersionMajor >= 16;
-
-  const importEsm = (() => {
-    if (isAsyncLoader) {
-      return async (p) => {
-        return await import(p);
-      };
-    }
-    try {
-      // eslint-disable-next-line import/no-unresolved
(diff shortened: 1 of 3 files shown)


Mmarzex (approved):

Danwakeem (approved):
👍

PR #352, opened 2023-01-05
Bumped proto version

Danwakeem:
## Description
I didn't want the proto version to start falling behind so I went ahead and bumped it here 👍


medikoo (approved):
👍

PR #353, opened 2023-01-06
fix: Added _origin to captured event before emit

Danwakeem:
## Description
We need to attach the `_origin` attribute to the captured event before we emit the event or else filtering out the event does not work.

> I was also trying to add this to the dev mode integration tests but I think we need to publish a new sdk version first before this will work? 🤔 TBH I am not sure what the flow is so if there is a way to update this and the dev mode integration tests at the same time let me know 😎


medikoo (approved):
Thanks @Danwakeem ! Indeed I totally overlooked that. I'll update the tests on AWS SDK side to pick that

Danwakeem:
Awesome thanks @medikoo 😎

PR #354, opened 2023-01-06
fix: Added v3 document client support

Danwakeem:
## Description
It is kind of funny we didn't notice this until recently but if you use document client in the v3 sdk the `dynamodb` tags were not showing up in the `aws.sdk` tags section because it was called `dynamodbdocument` instead of `dynamodb`.

This addresses that issue by adding a common dynamodb config between document client and the usual dynamo client 👍


medikoo (approved):
@Danwakeem thanks for covering that! 🎉 

I assume we're 100% sure that exactly same mapping has to be used for DocumentClient (?)

Danwakeem:
Yes @medikoo I had verified the fields between document client and the main dynamo class and they shared the same fields we had listed here 👍

PR #355, opened 2023-01-08
SDK: SC-257 `TraceSpan` implementation, unit tests

alexdelorenzo:
This PR is for SC-257 and it contains:
 - SC-292: Implement `TraceSpan` with exposed properties and methods
 - SC-293: Handle root `TraceSpan` and `parentSpan` with context variables
 - SC-294: Implement name, value and tags validation
 - SC-295: Expose `createTraceSpan()` on SDK object
 - SC-296: Implement unit tests
 - SC-297: Implement generation of protobuf intermediate object

It also includes exceptions that are raised that match the same errors that are thrown in the JS SDK, as well as a `Tags` implementation with unit tests.

Tests, validation, linting, formatting and CI automation all pass on my end for this PR.

[You can browse through this branch here](https://github.com/alexdelorenzo/console/tree/sc-257/python/packages/sdk)


alexdelorenzo:
@medikoo Looks like this PR passed CI validation and is good to go for review :+1:

medikoo (commented):
@alexdelorenzo after merging #342 this needs to be rebased against `main`

alexdelorenzo:
> @alexdelorenzo after merging #342 this needs to be rebased against `main`

@medikoo `sc-257` was rebased with `main` 👍

python/packages/sdk/pyproject.toml
```diff
@@ -1,5 +1,8 @@
 [build-system]
-requires = ["setuptools>=65.6.3", "wheel>=0.38.4"]
+requires = [
+    "setuptools>=65.6.3",
+    "wheel>=0.38.4",
+]
```
medikoo, line 6: Ideally if whitespace changes are not introduced, as they blur the substance of the PR

Can we cover this file with some form of _prettier_ which will ensure unified style in all cases?

python/packages/sdk/sdk/span/trace.py
```diff
@@ -0,0 +1,184 @@
...
+        global root_span
+        default: Nanoseconds = time_ns()
+
+        if self.end_time is not None:
+            raise ClosureOnClosedSpan("TraceSpan already closed.")
+
+        self.end_time = default if end_time is None else end_time
+
+        # if self is root_span:
+        #     ctx.set(None)
```
medikoo: Let's not commit commented out code

python/packages/sdk/sdk/sdk/base.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
```
medikoo, line 1: Why do we end with `sdk/sdk/sdk/base.py` path  (`sdk` nested three times)? Is this some Python specificity?

medikoo (commented):
Thank you, @alexdelorenzo! In general, it looks very promising. I have just a few style & organization concerns. Please see my comments

python/packages/sdk/pyproject.toml
alexdelorenzo: As `pyproject.toml` is a relatively new convention, it doesn't look like the standard formatting tools like `black`, `yapf`, etc support formatting package metadata, although [this project](https://pyproject-fmt.readthedocs.io/en/latest/) exists to unify style. It only gives you knobs to set the number of spaces per indent, though (default is 2 with the tool, `black` insists on 4). I can integrate it into the project easily.

python/packages/sdk/sdk/span/trace.py
alexdelorenzo: Good catch, will do

python/packages/sdk/sdk/sdk/base.py
alexdelorenzo: The first `sdk` is the bare folder that holds the package, it can be named anything, but the spec from SC-220 said to keep it as `python/packages/sdk`. The second `sdk` is the module name. This name will need to be changed for publication, as [`setuptools` demands that module names are the same as package names](https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-package-dir). 

In SC-259, I've changed the second `sdk` name, the module name, to `serverless_sdk` to match the package name that will appear on PyPI as per the `setuptools` docs, so the new full path for the packaged module will be `console/python/packages/sdk/serverless_sdk`.

The final `sdk` directory is a submodule that holds the logic for the `serverlessSdk` object itself. After conforming to the naming conventions required for publication, the final path will look like `python/packages/sdk/serverless_sdk/sdk`, where the final `sdk`submodule is aptly named for what will end up being distributed simply as `serverless_sdk/sdk` without any other nested `sdk` directories.

alexdelorenzo:
@medikoo Made changes as per your comments, sans the final one about module names. Happy to change those if necessary, though 🙂

python/packages/sdk/sdk/sdk/base.py
medikoo: Great thanks for explanations. Makes sense

medikoo (approved):
Thank you @alexdelorenzo !

PR #356, opened 2023-01-09
SDK: SC-258 Protobuf generation & interfaces, unit tests, CI

alexdelorenzo:
This PR is for SC-258 and it contains:
 - Protobuf interface generation
 - Export payload encoders
 - Unit test payload parsing and validation
 - Add CI validation for `sdk-schema`

CI code validation passes on my end :+1: 

[You can browse through this branch here](https://github.com/alexdelorenzo/console/tree/sc-258/python/packages/sdk-schema)

NOTE: I saw [the draft Max made](https://github.com/serverless/console/pull/334) that uses [Buf.build](https://buf.build) to automatically generate Python classes from protobuf interfaces, but found that it does not output importable Python modules. 

[It looks like Python support from Buf is experimental and not as well supported as Go and JavaScript are with Buf](https://docs.buf.build/bsr/remote-packages/overview#supported-languages), and to use the files it generated would mean writing scripts to convert them into modules, which can be pretty hacky.

The code Buf outputs for Python is also non-idiomatic, lacks type annotations, lacks `async` support which is necessary for FastAPI support along with other modern web frameworks, and [there are more issues with Buf's backend generator for Python documented here](https://github.com/danielgtaylor/python-betterproto#motivation).

For those reasons, I chose to generate Python code for the protobuf interfaces using [`betterproto`](https://github.com/danielgtaylor/python-betterproto). It addresses the issues I mentioned above, and generates idiomatic Python code with type annotations, has `async` support and the code it generates is generally more pleasant to work with for both developers and consumers of this library.

Code generation via `betterproto` can be automated for packaging via CI like Buf currently does for the Node code it generates.


alexdelorenzo:
@medikoo CI validation passes and this PR is good to go for review :+1:

medikoo (commented):
@alexdelorenzo after merging #342 this needs to be rebased against `main`

Mmarzex (changes requested):
@alexdelorenzo That's fine if the remote plugin from buf doesn't work for Python. Can we do something similar to what we do for node then?

We use `ts-proto` to generate our node `sdk-schema` and buf just leverages the locally installed version to generate typescript classes.

https://github.com/serverless/console/blob/main/proto/buf.gen.yaml#L3

alexdelorenzo:
> @alexdelorenzo That's fine if the remote plugin from buf doesn't work for Python. Can we do something similar to what we do for node then?
> 
> We use `ts-proto` to generate our node `sdk-schema` and buf just leverages the locally installed version to generate typescript classes.
> 
> https://github.com/serverless/console/blob/main/proto/buf.gen.yaml#L3

@medikoo Sure that looks doable, I think `betterproto` acts as a `protoc` plugin so it can fit into Buf's local plugin model.

alexdelorenzo:
@Mmarzex Looks like I've hooked `betterproto` up to Buf successfully according to tests, at least. I had to move the Python configuration to `buf.gen.python.yaml`, as leaving it in `buf.gen.yaml` would trigger Buf to build the Python schema when validating/building the Node module, and would build the Node schema when running the Python CI.

alexdelorenzo:
@Mmarzex @medikoo Not sure who wants to review this, but I believe it reflects the requested changes  👍

medikoo:
> @Mmarzex @medikoo Not sure who wants to review this, but I believe it reflects the requested changes 👍

@alexdelorenzo be sure to re-request review in top right box. e.g. after last review of mine it was not re-requested, so I assumed this work is still pending

.github/workflows/python-validate.yml
```diff
@@ -77,3 +77,53 @@ jobs:
...
+
+          python3 -m pip install poethepoet
+          python3 -m poethepoet install
+
+      - name: Validate SDK Schema code formatting
+        run: |
+          cd python/packages/sdk-schema
+
+          python3 -m pip install black
+          python3 -m black --check .
```
medikoo: As commented in other PR. It'll be probably better to have one `black` call for whole `python` directory

.github/workflows/python-validate.yml
```diff
@@ -77,3 +77,53 @@ jobs:
...
+
+      - name: Check SDK Schema code with linter
+        run: |
+          cd python/packages/sdk-schema
+
+          python3 -m pip install ruff
+          python3 -m ruff .
```
medikoo: Similarly linting probably could work (be done against `python`)

.github/workflows/python-validate.yml
```diff
@@ -77,3 +77,53 @@ jobs:
...
+
+      - name: Run SDK Schema unit tests
```
medikoo: Shouldn't those tests be run after we have proto changes linted and build?

alexdelorenzo: Good point 👍

alexdelorenzo:
@medikoo I believe I've addressed the requested changes you've made 👍

.github/workflows/python-validate.yml
```diff
@@ -56,19 +58,33 @@ jobs:
...
+        if: steps.pathChanges.outputs.packages == 'true'
         run: |
-          cd python/packages/sdk
+          cd python/packages
 
           python3 -m pip install ruff
-          python3 -m ruff sdk
+          python3 -m ruff .
 
       - name: Run unit tests
```
medikoo: Let's name this step `Run SDK unit tests`

medikoo (commented):
Thanks @alexdelorenzo looks good to me. I have just one minor not necessarily related to this changes suggestion.

I let  @Mmarzex approve before merging

Mmarzex (dismissed):

alexdelorenzo:
@medikoo Should be all good to go

medikoo (approved):
Thank you @alexdelorenzo !

PR #360, opened 2023-01-09
fix: Adjusted typescript implementation for sdk

Danwakeem:
## Description
Changed the `index.d.ts` file to import the main implementation file via the `index.ts` file as well as export the `Sdk` type. This should give typescript projects typical intellisense and the ability to import the sdk like typical typescript packages.

### DIFF
--- a/node/packages/sdk/.ts-types/index.d.ts
+++ b/node/packages/sdk/.ts-types/index.d.ts
@@ -19,7 +19,7 @@ export interface Sdk {
     options?: {
       startTime?: bigint;
       immediateDescendants?: string[];
-      tags?: Record<string, boolean | number | string | Date | Array<any> | null>;
+      tags?: Record<string, boolean | number | string | Date | Array<unknown> | null>;
       input?: string;
       output?: string;
       onCloseByRoot?: Function;
@@ -28,13 +28,13 @@ export interface Sdk {
   captureError(
     error: Error,
     options?: {
-      tags?: Record<string, boolean | number | string | Date | Array<any> | null>;
+      tags?: Record<string, boolean | number | string | Date | Array<unknown> | null>;
     }
   ): undefined;
   captureWarning(
     message: string,
     options?: {
-      tags?: Record<string, boolean | number | string | Date | Array<any> | null>;
+      tags?: Record<string, boolean | number | string | Date | Array<unknown> | null>;
     }
   ): undefined;
 }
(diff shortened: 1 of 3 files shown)


node/packages/sdk/.ts-types/index.d.ts
```diff
@@ -18,7 +19,7 @@ interface Sdk {
     options?: {
       startTime?: bigint;
       immediateDescendants?: string[];
-      tags?: Record<string, boolean | number | string | Date | Array | Null>;
+      tags?: Record<string, boolean | number | string | Date | Array<any> | null>;
```
Mmarzex: Probably would be better to use `Array<unknown>` here so it doesn't get caught by `tsc` that is configed for no any's.

Mmarzex (changes requested):

node/packages/sdk/.ts-types/index.d.ts
Danwakeem: Yeah that is a good call, made this change 😎

Mmarzex (approved):

medikoo:
@Danwakeem this PR broke our extension builds.. all integration tests now fail. Do you know what could be the cause?

Danwakeem:
@medikoo I pulled down the `main` branch, reinstalled dependencies, ran the build in the `aws-lambda-sdk` directory, and I am seeing the failed integration tests but I am not sure what exactly the error is? 🤔 

When I look a the content of the zip file that was created and look at the content of the `extension.internal/nodejs/node_modules/@serverless/sdk/index.js` I see some references to a seemingly local path and I am wondering if that may have something to do with it? 👇 

Screenshot 2023-01-11 at 11 40 03 AM

Does that look unusual to you? I am not sure what this looked like before but seems strange I guess. I also don't how or why this PR would have effected the builds. The build file seems to only reference the `index.js` file which is was unchanged here 🤔

Danwakeem:
Yeah when I revert this PR those local paths seem to go back to normal 🤔 Weird. I am still playing around with this a bit.

medikoo:
@Danwakeem I'm not sure as well what's going on. It appears ESbuild is trying to be smart with TS and something breaks.

Can we revert this PR, until this is figured out? I wanted to push new SDK release now, so we have it ready for tomorrow with recent changes

Danwakeem:
@medikoo Yeah that is alright with me 👍 I can work on this in the background while you get the release out.

PR #364, opened 2023-01-10
SDK: Relax tag name format restriction

medikoo:


Danwakeem (dismissed):
This looks good but I do think it would be nice if we did not resolve with an error if a tag does not follow this format.

It would be better if we stripped the tag from the map and published a warning about the tag name instead.

medikoo:
> It would be better if we stripped the tag from the map and published a warning about the tag name instead.

I'm not sure about that, as we're speaking about explicit API usage here, and common behavior among all popular JS APIs is to throw in case of unexpected usage and not log warnings. 

Also logging in this case is likely to hide issue in confusing way. User may not notice our warning, and at the same time assume that configured warning is never triggered, as searching for given tag UI will always show zero results.

If we feel that pattern is still too restrictive I'd simply further relax it.

Danwakeem:
I don't think we should be following typical API usage for our tags given this is for reporting errors/warnings. It doesn't make sense that we would throw an error for an invalid format when reporting an error (or IMO a warning).

For example, I have this code. (This is for real something I tried today and crashed my lambda when I was testing 😅 I did not read the docs I just assumed the sdk would not crash when reporting my custom tag lol naive I know but this is most likely what customers will be doing)

``` javascript
try {
  if (!req?.body?.title) throw new Error('Missing title for form');
  
  const Item = {
    pk: 'form',
    sk: `form:${uuid()}`,
    title: req.body.title,
    description: req.body.description,
  };
  await ddb.send(new PutCommand({
    TableName,
    Item,
  }));
  return res.send({
    forum: Item
  });
} catch(error) {
  sdk.captureError(error, {
    tags: { body: req.body },
  });
  return res.status(400).send({
    message: error.message,
  });  
}
```

In this case, it would be really strange to have a try/catch around the `sdk.captureError` function because I am trying to log an error so there would be no way for me to safely find all of the edge cases of our tags set up without throughly reading the docs which most people, myself included, are not always going to do or remember what the docs said. 

It would be one thing if the `sdk` functions were safely wrapped in a try/catch where the customer might handle things more gracefully but more than likely customers will be using this outside of a safe try catch block which might end up causing lambda crashes that trigger new instances of the lambda function to be booted up increasing cold starts, etc.

This is what happened when I tried this code today 👇 
Screenshot 2023-01-10 at 12 29 01 PM

Given I am not the decision maker here I think @Mmarzex should weigh in 🙏 I am going to mark this as requested changes as I think we need to do more there.

Danwakeem (changes requested):
Just marking this as requested changes because I am finding a few more reasons why we should prioritize not crashing when we run into format issues.

Relaxing the standard is alright but given the nature of the functions we are exposing I think the real solution here is not crashing when tags are incorrect.

medikoo:
> This is what happened when I tried this code today 👇

@Danwakeem I think the real issue here is not that we throw that way but that users may want to store JSON objects as tag values, and we do not allow that.

The solution to that could be to _stringify_ JSON values, and we can even go forward and use Node.js `inspect` to normalize any other non-JSON values. @Mmarzex, what do you think about that?

Ideally, this API should fail only on obvious misusage and otherwise accept and process input the best it can.
In my eyes, this is what we're really after here.

Mmarzex:
Our general approach should be to never crash a user's function due to our SDK. The errors we see with tags here strikes me as a good usecase for sending a Console internal error for users to see the error in Console. That way their function doesn't fail but they can still be notified/alerted to incorrect usage.

medikoo:
> The errors we see with tags here strikes me as a good usecase for sending a Console internal error for users to see the error in Console. That way their function doesn't fail but they can still be notified/alerted to incorrect usage.

@Mmarzex what do you think is better:
1. Supporting objects as tag values (to be stored in stringified format)
2. Not supporting objects as tag values, but instead reporting such usage as an error in our console

?

Mmarzex:
@medikoo We should not have object types as values for custom tags. I don't think there is any usefulness to filtering by a stringified object. 

So let's go with your second option.

medikoo:
> So let's go with your second option.

To avoid back and forth let me also confirm with @skierkowski and @ac360  on Slack

medikoo:
Ok, I've pushed the update, so errors on user input validation are logged into the console instead of being thrown

Danwakeem (approved):
Looking good 😎

PR #365, opened 2023-01-10
SDK: SC-259 Publication of new versions

alexdelorenzo:
This PR is for SC-259 and it contains:
 - Integration logic that sets tags upon new versions
 - Publication logic that publishes upon new versions `>=0.0.1`

The publication depends on setting the `PYPI_AUTH_TOKEN` secret, which can be retrieved here: https://pypi.org/help/#apitoken


alexdelorenzo:
@medikoo Looks like this one is good to go without merge conflicts

.github/workflows/python-sdk-publish.yaml
```diff
@@ -0,0 +1,43 @@
+name: "Python: Publish @python/serverless-sdk"
+
+on:
+  push:
+    tags:
+      - "@python/serverless-sdk@[0-9]+.[0-9]+.[0-9]+"
```
medikoo: I think we don't need `@` prefix. It can be just `python/serverless-sdk@[0-9]+.[0-9]+.[0-9]+`

.github/workflows/python-sdk-schema-publish.yaml
```diff
@@ -0,0 +1,65 @@
...
+jobs:
+  publishNewSdkVersion:
+    name: Publish new version
+    runs-on: ubuntu-latest
+    env:
+      # It'll work with secrets.GITHUB_TOKEN (which is provided by GitHub unconditionally)
+      # Still then release author would be "github-actions". It's better if it's dedicated repo bot
+      GITHUB_TOKEN: ${{ secrets.USER_GITHUB_TOKEN }}
+    steps:
+      - name: Resolve path filters
```
medikoo: What's the purpose of this step?

.github/workflows/python-validate.yml
```diff
@@ -60,15 +60,15 @@ jobs:
           cd python/packages/sdk
 
           python3 -m pip install black pyproject-fmt
-          python3 -m black --check sdk
+          python3 -m black --check .
```
medikoo: In JS, we have prettier check for whole `node` directory. Could it work that way here as well?

.github/workflows/python-validate.yml
```diff
@@ -60,15 +60,15 @@ jobs:
...
           python3 -m pyproject_fmt --indent 4 pyproject.toml
 
       - name: Check code with linter
         run: |
           cd python/packages/sdk
 
           python3 -m pip install ruff
-          python3 -m ruff sdk
+          python3 -m ruff .
```
medikoo: Same here

.github/workflows/python-validate.yml
```diff
@@ -77,3 +77,63 @@ jobs:
 
           python3 -m pip install pytest
           python3 -m pytest
+
+      - name: Tag if new version
```
medikoo: Tagging should happen only in _integrate_ builds that are run on all commits to `main` (we should not attempt taggin on PRs).

Also we should tag only when all tests (including integration if there are any) pass (see e.g. https://github.com/serverless/console/blob/4ba7158b9e0db8bfd06db031eb30af76cea8de5d/.github/workflows/node-integrate.yml#L514 and `needs` section)

python/packages/sdk-schema/pyproject.toml
```diff
@@ -1,14 +1,40 @@
 [build-system]
-requires = ["setuptools>=65.6.3", "wheel"]
 build-backend = "setuptools.build_meta"
+requires = [
```
medikoo: This changes seem unrelated to this PR.

Let's keep this PR only about CI publication setup, and not mixin other work here

medikoo (changes requested):
Thanks @alexdelorenzo, please see my comments

.github/workflows/python-sdk-schema-publish.yaml
alexdelorenzo: I believe that's required for detecting changes to paths and so you can do things like:
```
if: steps.pathChanges.outputs.sdkSchema == 'true' || steps.pathChanges.outputs.proto == 'true'
```

.github/workflows/python-validate.yml
alexdelorenzo: Just tested it out and `black` allows you to do that

python/packages/sdk-schema/pyproject.toml
alexdelorenzo: I believe this change was necessary to pass `pyproject-fmt`'s validation after adding the build deps necessary to publish builds of this module, it seems to insist on an order to how the metadata is listed.

alexdelorenzo: Yeah, I just ran it again, it's a change made by `pyproject-fmt`

alexdelorenzo: Happy to remove it, just don't think this PR will pass CI validation without it

.github/workflows/python-sdk-schema-publish.yaml
alexdelorenzo: Just realized I'm looking at the publishing logic, good find 👍

.github/workflows/python-validate.yml
alexdelorenzo: Added a commit that does this with `ruff`

alexdelorenzo:
@medikoo Added changes per your review 👍

python/packages/sdk-schema/pyproject.toml
medikoo: If it's needed in context of this PR, let's keep it. I blindly assumed this change relates to SDK Schema generation which is not in context here

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,172 @@
...
+
+name: "Python: Integrate"
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - node/**
+      - go/**
+      - proto/**
```
medikoo: I believe should be just `python` and `proto`

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,172 @@
...
+
+env:
+  FORCE_COLOR: 1
+  FORMAT_INSPECT_DEPTH: 20
```
medikoo: This are env vars that improve output from Node.js test runners, Are they needed in Python case?

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,172 @@
...
+
+jobs:
+  linuxPython37:
+    name: "[Linux] Python 3.7: Unit Tests"
+    runs-on: ubuntu-latest
+    outputs:
+      pathChangedLambdaSdk: ${{ steps.pathChanges.outputs.lambdaSdk }}
```
medikoo: We don't have this package yet, so let's not add this at this point

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,172 @@
...
+        uses: dorny/paths-filter@v2
+        id: pathChanges
+        with:
+          filters: |
+            sdk:
+              - 'python/packages/sdk/**'
+            sdkSchema:
+              - 'python/packages/sdk-schema/**'
+            lambdaSdk:
+              - 'python/packages/aws-lambda-sdk/**'
```
medikoo: Same here

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,172 @@
...
+          # Ensure to have complete history of commits pushed with given push operation
+          # It's loose and imperfect assumption that no more than 30 commits will be pushed at once
+          fetch-depth: 30
+          # Tag needs to be pushed with real user token, otherwise pushed tag won't trigger the actions workflow
+          # Hence we're passing 'serverless-ci' user authentication token
+          token: ${{ secrets.USER_GITHUB_TOKEN }}
+
+      - name: Tag if new version
+        run: |
+          NEW_VERSION=`git diff -U0 ${{ github.event.before }} python/packages/aws-lambda-sdk/pyproject.toml | grep 'version = ' | grep -oE "[0-9]+\.[0-9]+\.[0-9]+"` || :
```
medikoo: I think package path should be `/sdk`

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,172 @@
...
+          if [ -n "$NEW_VERSION" ] && [ $NEW_VERSION != "0.0.0" ];
+          then
+            git tag @python/serverless-sdk@$NEW_VERSION
```
medikoo: Let's strip leading `@`

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,172 @@
...
+          # Tag needs to be pushed with real user token, otherwise pushed tag won't trigger the actions workflow
+          # Hence we're passing 'serverless-ci' user authentication token
+          token: ${{ secrets.USER_GITHUB_TOKEN }}
+
+      - name: Tag if new version
+        run: |
+          NEW_VERSION=`git diff -U0 ${{ github.event.before }} python/packages/sdk-schema/pyproject.toml | grep 'version = ' | grep -oE "[0-9]+\.[0-9]+\.[0-9]+"` || :
+          if [ -n "$NEW_VERSION" ] && [ $NEW_VERSION != "0.0.0" ];
+          then
+            git tag @python/serverless-sdk@$NEW_VERSION
```
medikoo: Let's strip leading `@`

.github/workflows/python-validate.yml
```diff
@@ -77,3 +93,31 @@ jobs:
...
+
+      - name: Install SDK Schema
+        if: steps.pathChanges.outputs.sdkSchema == 'true'
+        run: |
+          cd python/packages/sdk-schema
+
+          python3 -m pip install poethepoet
+          python3 -m poethepoet install
+
+      - name: Run SDK Schema unit tests
```
medikoo: I believe this should be run after the build.

Also, let's ensure that PR's just address the scope of the task. I see this change being proposed in two different PR's which should not be the case.

Either it should be proposed separately or included just in one where SDK Schema build is configured

medikoo (changes requested):
Thanks @alexdelorenzo we're getting closer!

medikoo:
@alexdelorenzo, when working on a feature and you find a need for a small improvement to the existing codebase, which is not strictly to the feature. It's fine, and I think better to propose it with a separate dedicated PR.
It's also better for our processing flow and for project history, as the given part, when excluded, doesn't blur the substance of other PRs and also will appear in `main` commit history as a distinct commit.
Every PR ends as a single commit, and when this single commit contains many unrelated things, it may turn into a real maintenance issue, especially if, for some reason, we need to revert from something

.github/workflows/python-integrate.yaml
alexdelorenzo: Good catch, beginning to think I might need to increase the text size in my IDE

.github/workflows/python-integrate.yaml
alexdelorenzo: Nope, I had assumed they were set for another purpose in the CI stack. I'll remove them

alexdelorenzo:
> @alexdelorenzo, when working on a feature and you find a need for a small improvement to the existing codebase, which is not strictly to the feature. It's fine, and I think better to propose it with a separate dedicated PR. It's also better for our processing flow and for project history, as the given part, when excluded, doesn't blur the substance of other PRs and also will appear in `main` commit history as a distinct commit. Every PR ends as a single commit, and when this single commit contains many unrelated things, it may turn into a real maintenance issue, especially if, for some reason, we need to revert from something

Thanks for the heads up and explanation about PR etiquette/processes, I'll keep that in mind for the future 🙂

.github/workflows/python-validate.yml
alexdelorenzo: I moved `Run SDK Schema unit tests` after the build, but were you also referencing `Install SDK project and dependencies` as well?

And will do, I think this might have been the result of rebasing this branch with `sc-25*`, I'll avoid doing that before they're merged with `main`.

medikoo (changes requested):
@alexdelorenzo this PR shows conflicts to be resolved

.github/workflows/python-integrate.yaml
medikoo: Seems not addressed

alexdelorenzo:
@medikoo Cleaned up this PR, fixed merge conflicts and addressed the references to the Lambda SDK by removing them 👍

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,136 @@
...
+          # Tag needs to be pushed with real user token, otherwise pushed tag won't trigger the actions workflow
+          # Hence we're passing 'serverless-ci' user authentication token
+          token: ${{ secrets.USER_GITHUB_TOKEN }}
+
+      - name: Tag if new version
+        run: |
+          NEW_VERSION=`git diff -U0 ${{ github.event.before }} python/packages/sdk/pyproject.toml | grep 'version = ' | grep -oE "[0-9]+\.[0-9]+\.[0-9]+"` || :
+          if [ -n "$NEW_VERSION" ] && [ $NEW_VERSION != "0.0.0" ];
+          then
+            git tag python/serverless-sdk@$NEW_VERSION
```
medikoo: If package name on PIP side is `serverless_sdk` (not `serverless-sdk`) then let's tag it as `python/serverless_sdk`

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,136 @@
...
+          # Tag needs to be pushed with real user token, otherwise pushed tag won't trigger the actions workflow
+          # Hence we're passing 'serverless-ci' user authentication token
+          token: ${{ secrets.USER_GITHUB_TOKEN }}
+
+      - name: Tag if new version
+        run: |
+          NEW_VERSION=`git diff -U0 ${{ github.event.before }} python/packages/sdk-schema/pyproject.toml | grep 'version = ' | grep -oE "[0-9]+\.[0-9]+\.[0-9]+"` || :
+          if [ -n "$NEW_VERSION" ] && [ $NEW_VERSION != "0.0.0" ];
+          then
+            git tag python/serverless-sdk@$NEW_VERSION
```
medikoo: `python/serverless_sdk_schema`

.github/workflows/python-integrate.yaml
```diff
@@ -0,0 +1,136 @@
...
+        run: |
+          cd python/packages/sdk
+
+          python3 -m pip install --editable .
+
+      - name: Unit tests /sdk
+        if: steps.pathChanges.outputs.sdk == 'true'
+        run: |
+          cd node/packages/sdk
+          npm test
```
medikoo: I think it's supposed to test Python SDK package here

.github/workflows/python-sdk-publish.yaml
```diff
@@ -0,0 +1,39 @@
...
+          # When using only `pyproject.toml` for dependencies, see:
+          #  https://github.com/actions/setup-python/issues/529#issuecomment-1367029699
+          cache: 'pip'
+          cache-dependency-path: |
+            **/pyproject.toml
+
+      - name: Install main project dependencies
+        run: |
+          cd python/packages/sdk
+          python3 -m pip install .
```
medikoo, line 28: Does for publication package dependencies need to be installed? (Asking because in Node.js it's not the case - dependencies are only needed if we want to run package code)

medikoo (changes requested):
Thanks @alexdelorenzo, please see my comments

.github/workflows/python-sdk-publish.yaml
alexdelorenzo: This will ensure that build dependencies are installed before the package is built and published. Otherwise the build dependencies will need to be explicitly installed in the publication step, and the two can go out of sync if new build deps are added to the package.

alexdelorenzo:
@medikoo Added the requested changes 👍

.github/workflows/python-sdk-publish.yaml
medikoo: Ok, so to publish the package to pip, extra dependencies are needed, and this is what ensures that's the case (on Node.js side, npm is good enough, which naturally comes with Node.js).

Thanks for explanations

.github/workflows/python-sdk-publish.yaml
```diff
@@ -0,0 +1,39 @@
+name: "Python: Publish python/serverless-sdk"
+
+on:
+  push:
+    tags:
+      - "python/serverless-sdk@[0-9]+.[0-9]+.[0-9]+"
```
medikoo: `serverless_sdk`

.github/workflows/python-sdk-publish.yaml
```diff
@@ -0,0 +1,39 @@
+name: "Python: Publish python/serverless-sdk"
```
medikoo: Let's maybe name it: "Python: Publish serverless_sdk"

.github/workflows/python-sdk-schema-publish.yaml
```diff
@@ -0,0 +1,49 @@
+name: "Python: Publish python/serverless-schema-sdk"
```
medikoo: Same here, I'll revise the name

.github/workflows/python-sdk-schema-publish.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  push:
+    tags:
+      - "python/serverless-schema-sdk@[0-9]+.[0-9]+.[0-9]+"
```
medikoo: Here tag also needs to be updated

medikoo (changes requested):
Thanks @alexdelorenzo, it's near great :) Please see my comments

alexdelorenzo:
@medikoo Made the requested changes, think we're there :)

medikoo (approved):
Thank you @alexdelorenzo !

PR #368, opened 2023-01-11
AWS Lambda SDK: Revert problematic TS addition

medikoo:
Revert https://github.com/serverless/console/pull/360

1. Our extension stopped building properly
2. We should not introduce TS files. I believe as in the case of other packages that take a similar approach, just typing should be good enough to have ensured proper IntelliSense. If there's an issue with that I'll look into improvement to typings on their own


Danwakeem (approved):
👍

PR #369, opened 2023-01-11
fix: Fixed typings file

Danwakeem:
## Description
Updated the `index.d.ts` file so we are exporting the correct public functions and do not need a silly typescript index file 😄

### DIFF
--- a/node/packages/sdk/.ts-types/index.d.ts
+++ b/node/packages/sdk/.ts-types/index.d.ts
@@ -7,40 +7,44 @@ export interface Instrumentation {
   expressApp: ExpressAppInstrument;
 }
 
-interface Sdk {
+export function createTraceSpan(
+  name: string,
+  options?: {
+    startTime?: bigint;
+    immediateDescendants?: string[];
+    tags?: Record<string, boolean | number | string | Date | Array<unknown> | null>;
+    input?: string;
+    output?: string;
+    onCloseByRoot?: Function;
+  }
+): TraceSpan;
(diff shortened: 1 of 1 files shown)


node/packages/sdk/.ts-types/index.d.ts
```diff
@@ -7,41 +7,45 @@ export interface Instrumentation {
...
+    tags?: Record<string, boolean | number | string | Date | Array<unknown> | null>;
+  }
+): undefined;
+
+export interface Sdk {
   name: string;
   version: string;
   orgId: string;
   traceSpans: TraceSpans;
   instrumentation: Instrumentation;
```
Mmarzex, line 43: Wouldn't the `Sdk` interface still need those method definitions on them @Danwakeem ?

Mmarzex (changes requested):

node/packages/sdk/.ts-types/index.d.ts
Danwakeem: @Mmarzex yes, I believe that is what the next three lines should do 👇 

https://github.com/serverless/console/blob/0444c35962983cfb69056c977e3ebf5bab672d53/node/packages/sdk/.ts-types/index.d.ts#L44-L46

Mmarzex: 🤦 Totally missed that.

Mmarzex (approved):

PR #370, opened 2023-01-11
chore: Release v0.2.5 of the sdk

Danwakeem:
## Description
Releasing this version of the sdk so that we can include it in the `aws-lambda-sdk` release 😎


Danwakeem:
Actually sorry I think I found one more improvement we could make to the types 😅 Closing this for now 😎

PR #371, opened 2023-01-11
chore: Improved types

Danwakeem:
## Description
Actually after reading a bit more about the `declare` keyword I believe this is much more accurate as to what exactly this SDK is exporting 👍 

This will give us the full interface with the non function attributes. Probably not super important but definitely more accurate 😎

### DIFF
--- a/node/packages/sdk/.ts-types/index.d.ts
+++ b/node/packages/sdk/.ts-types/index.d.ts
@@ -7,43 +7,37 @@ export interface Instrumentation {
   expressApp: ExpressAppInstrument;
 }
 
-export function createTraceSpan(
-  name: string,
-  options?: {
-    startTime?: bigint;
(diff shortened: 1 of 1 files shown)


Mmarzex (approved):

medikoo (commented):
@Danwakeem yes, this version is definitely better, as we do not export disconnected values, but the SDK instance and this update (contrary to the previous one) reflect that.

If I understand correctly the issue was that we were exporting the interface, which per semantics, is not functional on its own, and declaring and exporting a value that supposedly implements the interface fixed the issue.

I'm glad it got sorted out properly this time :)

Danwakeem:
Yes that is exactly right @medikoo 👍 I have never written typescript types fora non typescript project before so I was unaware of this difference. 😅 also glad we got this sorted out properly.

PR #374, opened 2023-01-12
SDK: Make `createTraceSpan` private

medikoo:
`sdk.createTraceSpan` is removed. The functionality of creating custom spans is not considered part of public API at the time being (a new method for that will be introduced in one of the following releases)


Danwakeem (approved):
🙈

PR #375, opened 2023-01-12
SDK: Dedicated handling for structured warnings issued by the SDK

medikoo:
We want to report warning as generated by SDK to the Console and label them as coming from this SDK.
Additional requirement was to rely on structure logging, so also in CloudWatch those logs are discoverable

This patch implements handing for SDK originated structured warnings as such:

```javascript
console.warn({ source: "serverlessSdk", message: "Something is off"})
```

Those are caught as captured warning events and marked as of SDK type


Danwakeem (approved):
👍

PR #378, opened 2023-01-12
AWS Lambda SDK: Report internal warnings with structured logs

medikoo:
In response to #375

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/lib/send-telemetry.js
+++ b/node/packages/aws-lambda-sdk/instrument/lib/send-telemetry.js
@@ -33,10 +33,14 @@ module.exports = limit(10, async (name, body) => {
         },
         (response) => {
           if (response.statusCode !== 200) {
-            process.stderr.write(
-              'Serverless SDK Warning: Cannot propagate telemetry, ' +
-                `server responded with "${response.statusCode}" status code\n`
-            );
+            console.warn({
+              source: 'serverlessSdk',
+              message:
+                'Cannot propagate telemetry, ' +
+                `server responded with "${response.statusCode}" status code\n`,
+              code: 'DEV_MODE_SERVER_REJECTION',
+              httpStatusCode: response.statusCode,
+            });
           }
           response.on('data', () => {});
           response.on('end', resolve);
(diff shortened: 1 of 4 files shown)


node/packages/aws-lambda-sdk/instrument/lib/send-telemetry.js
```diff
@@ -33,10 +33,12 @@ module.exports = limit(10, async (name, body) => {
...
-            process.stderr.write(
-              'Serverless SDK Warning: Cannot propagate telemetry, ' +
-                `server responded with "${response.statusCode}" status code\n`
-            );
+            console.warn({
+              source: 'serverlessSdk',
+              message:
+                'Cannot propagate telemetry, ' +
+                `server responded with "${response.statusCode}" status code\n`,
+            });
```
Danwakeem: These warnings might be nice to search for in Cloudwatch so maybe it also make sense to put these variables in the string into the warning object? Maybe this one under `statusCode`?

node/packages/aws-lambda-sdk/trace-spans/aws-lambda.js
```diff
@@ -14,9 +14,10 @@ const arch = (() => {
     case 'arm64':
       return 'arm64';
     default:
-      process.stderr.write(
-        `Serverless SDK Warning: Unrecognized architecture: "${process.arch}"\n`
-      );
+      console.warn({
+        source: 'serverlessSdk',
+        message: `Unrecognized architecture: "${process.arch}"\n`,
```
Danwakeem, line 19: And maybe this one under `arch`?

Danwakeem (dismissed):
Looks good just thinking out loud about adding params to the warning objects if we are going that route 👍

node/packages/aws-lambda-sdk/trace-spans/aws-lambda.js
medikoo: This one is actually controversial, as I don't think it can ever happen. Either environment would have to change so our detection will stop working, or we will have to use e.g. arm arch build in some new arch as introduced by AWS

Danwakeem: Yeah true as far as I know I don't think there is a way to provide a custom arch for lambda functions

node/packages/aws-lambda-sdk/instrument/lib/send-telemetry.js
medikoo: That's a good hint. Although I'm not sure how useful it would be to search here on specific status codes.
What would definitely be nice is to narrow the search to all warnings of the same type, and for that, I've added `code` property (similar concept as there's with error).

Nonetheless I've also added `httpStatusCode` in this case.

Danwakeem: Oh yeah I like what you have added. That would make it easy to find these if they ever came up. Thanks man 👍

Danwakeem (approved):
:shipit:

PR #380, opened 2023-01-12
feat: Added increasing string as sequenceId

Danwakeem:
## Description
Added a sequenceId that increases in value so that we can use this as a fallback if two logs have the same timestamp


medikoo (dismissed):
👍 

_Side comment: I think commit scope should rather be "feat" or "refactor", "chore" is more dedicated to meta data updates (e.g. package.json etc.)_

Danwakeem:
@medikoo Ahh okay I am glad you mentioned this because I wasn't sure how to mark this one. It isn't a required feature so I wasn't sure if `feat` was appropriate here. I will mark it as `feat`.

medikoo (approved):
👍

PR #382, opened 2023-01-13
SDK: SC-260 Package with a handler wrapper

alexdelorenzo:
See comment on SC-260. Needs tests so it's not ready for merge review.


alexdelorenzo:
@medikoo Added tests and fixed some logic, this seems like it's ready for review. I believe something was said about integration testing with Node, as well

.github/workflows/python-sdk-publish.yaml
```diff
@@ -0,0 +1,43 @@
+name: "Python: Publish @python/serverless-sdk"
```
medikoo: @alexdelorenzo this PR should not introduce those changes.

Can you ensure it contains only changes related to SC-260?

medikoo (changes requested):
@alexdelorenzo can you please cleanup this PR so it purely focuses on SC-260.

It seems to contain a lot of unrelated work

medikoo:
> @medikoo Added tests and fixed some logic, this seems like it's ready for review. I believe something was said about integration testing with Node, as well

@alexdelorenzo yes, it's good that you mention it. Tomorrow I'll take a look on how to best tackle that.

.github/workflows/python-sdk-publish.yaml
alexdelorenzo: @medikoo Good catch, I've cleaned up this PR to reflect only the changes for SC-260

alexdelorenzo:
@medikoo I've cleaned up this PR to reflect only the changes for SC-260

.github/workflows/python-validate.yml
```diff
@@ -33,6 +33,8 @@ jobs:
               - 'python/packages/sdk/**'
             sdkSchema:
               - 'python/packages/sdk-schema/**'
+            lambdaSdkSchema:
```
medikoo: It should be `awsLambdaSdk`

.github/workflows/python-validate.yml
```diff
@@ -129,3 +131,14 @@ jobs:
 
           python3 -m pip install poethepoet pytest
           python3 -m poethepoet test
+
+      - name: Run AWS Lambda SDK unit tests
+        if: steps.pathChanges.outputs.lambdaSdkSchema == 'true'
+        run: |
+          cd python/packages/sdk
+          python3 -m pip install .
+          cd -
```
medikoo: I think this doesn't belong here.

This CI setup, I assume should introduce `Install AWS Lambda SDK project and dependencies` (which installs `aws-lambda-sdk` dependencies, also in this case ideally `sdk` should be installed from pip and not referenced  from side folder - it's how it works in Node.sj) and  `Run AWS Lambda SDK unit tests` which runs tests in it

python/packages/aws-lambda-sdk/pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+    "setuptools>=65.6.3",
+    "wheel>=0.38.4",
+]
+
+[project]
+name = "serverless_lambda_sdk"
```
medikoo: Let's name package `serverless_aws_lambda_sdk`

python/packages/aws-lambda-sdk/pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+version = "0.0.0"
+description = "Serverless Lambda SDK for Python"
```
medikoo: `AWS Lambda SDK`

python/packages/aws-lambda-sdk/serverless_lambda_sdk/internal_extension/base.py
```diff
@@ -0,0 +1,126 @@
+from __future__ import annotations
```
medikoo: @alexdelorenzo can you describe how Lambdas are expected to be set up to have that wrapping working?

After we package the Python SDK into a layer and attach it to the function, what environment variables with what values need to be ensured to have auto-wrapping work?

python/packages/sdk-schema/.gitignore
```diff
@@ -1 +1 @@
-sdk_schema/schema
+serverless_sdk_schema/schema
```
medikoo: Those changes seem unrelated to this work, can they be proposed with different PR?

medikoo (changes requested):
Thanks @alexdelorenzo. Please re-request the review after you feel that PR is ready for re-review.

I'm reviewing it now because I assume from your comment that it is ready. Still, it is not perfectly clear to me (and I still see unrelated work being included in this PR)

.github/workflows/python-validate.yml
alexdelorenzo: I agree, I was under the impression that steps ran independently and didn't share state, but that's not the case.

python/packages/aws-lambda-sdk/serverless_lambda_sdk/internal_extension/base.py
alexdelorenzo: Ah, didn't know that was in scope for this, but I've added functionality that addresses this. Users need to set `AWS_LAMBDA_EXEC_WRAPPER` to `/opt/serverless_aws_lambda_sdk/internal_extension/exec_wrapper`, assuming that path is how this SDK will be packaged for the Lambda extension layer.

python/packages/sdk-schema/.gitignore
alexdelorenzo: Removed

alexdelorenzo:
@medikoo Cleaned this up and made the requested changes. To answer your question about env vars and entrypoints: 
> Users need to set `AWS_LAMBDA_EXEC_WRAPPER` to `/opt/serverless_aws_lambda_sdk/internal_extension/exec_wrapper`, assuming that path is how this SDK will be packaged for the Lambda extension layer.

python/packages/aws-lambda-sdk/serverless_lambda_sdk/internal_extension/base.py
medikoo: > Ah, didn't know that was in scope for this

Without that, wrapper cannot really be utilized, so yes it needs to be complete in a sense that after merging it we should be able to wrap any Python function with it (otherwise how can we test that it works?)

> `AWS_LAMBDA_EXEC_WRAPPER` to `/opt/serverless_aws_lambda_sdk/internal_extension/exec_wrapper`, 

In Node.js it's `/opt/sls-sdk-node/exec-wrapper.sh`. It'll be nice if for Python we follow sam naming convention. So put all extension logic into `/opt/sls-sdk-python`  and have internal extension script accessible at `/opt/sls-sdk-python/exec-wrapper.sh`.

.github/workflows/python-validate.yml
```diff
@@ -51,7 +53,7 @@ jobs:
             **/pyproject.toml
 
       - name: Install SDK project and dependencies
-        if: steps.pathChanges.outputs.sdk == 'true'
+        if: steps.pathChanges.outputs.sdk == 'true' || steps.pathChanges.outputs.awsLambdaSdk == 'true'
```
medikoo: Do we need SDK dependency at this stage? I think just wrapping should not depend on base SDK in any way.

Also, ideally, we should depend on `serverless_sdk` as published to PiP and not local in the repository (going another way we risk publishing `serverless_aws_lambda_sdk`, which won't work with the PiP version, as it's tested against one in the repository)

.github/workflows/python-validate.yml
```diff
@@ -129,3 +131,19 @@ jobs:
...
+        if: steps.pathChanges.outputs.awsLambdaSdk == 'true'
+        run: |
+          cd python/packages/aws-lambda-sdk
+          python3 -m pip install .
+
+      - name: Run AWS Lambda SDK unit tests
+        if: steps.pathChanges.outputs.awsLambdaSdk == 'true'
+        env:
+          _HANDLER: "serverless_aws_lambda_sdk.tests.example"
+          LAMBDA_TASK_ROOT: "./python/packages/aws-lambda-sdk"
```
medikoo: I believe those should be hardcoded into test file (so we don't need to set those fixed env vars when running tests locally).

e.g. it's the case in Node.js: 
- https://github.com/serverless/console/blob/fa1801ac87943becaadb6d7b6fb2ee9342124f68/node/packages/aws-lambda-sdk/test/unit/internal-extension/index.test.js#L130-L135
- https://github.com/serverless/console/blob/fa1801ac87943becaadb6d7b6fb2ee9342124f68/node/packages/aws-lambda-sdk/test/unit/internal-extension/index.test.js#L15-L19

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/README.md
```diff
@@ -0,0 +1,5 @@
+# Serverless AWS Lambda SDK
+
+## Setup
+
+Set `AWS_LAMBDA_EXEC_WRAPPER` to `/opt/serverless_aws_lambda_sdk/internal_extension/exec_wrapper` in your Lambda environment.
```
medikoo: We will not expect users do that, so that should not be part of documentation.

I was asking for that details to understand how it's expected to be setup by our integration service later

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/base.py
```diff
@@ -0,0 +1,33 @@
+from __future__ import annotations
```
medikoo, line 1: `serverless_aws_lambda_sdk` in main module should return SDK that's exported by `serverless_sdk` but enriched with AWS Lambda specific things.

Node.js counterpart can be found here: https://github.com/serverless/console/blob/fa1801ac87943becaadb6d7b6fb2ee9342124f68/node/packages/aws-lambda-sdk/index.js

As scope of this PR is just wrapping, we should not focus on adding any AWS specific instrumentation, so it should export `serverless_sdk` as is, with just meta props as `main` and `version` updated.

While instrumentation that wraps lambda and gets initialized via internal extension ideally should be isolated and organized in different folder. In Node.js it's in `internal-extension` -> https://github.com/serverless/console/tree/main/node/packages/aws-lambda-sdk/internal-extension
We have there:
- `exec-wrapper.sh` to which `AWS_LAMBDA_EXEC_WRAPPER` will point
- `index.js` which is executed by instruction put in `exec-wrapper.sh`, and which overrides `_HANDLER` with new location of handler (so AWS runtime doesn't pick original handler but our wrapper instead)
- `wrapper.js` our Lambda handler, which initializes original handler and returns it wrapped with our decorator (this decorator is not baked into wrapper module, but imported from outer `instrument` module - reasoning is to allow to instrument handlers manually (not via internal extension) - although at this point we do not support such setup)

In Python I believe we should do similar, and I see we're close to that, just wanted to lay out it cleanly, so any spotted differences can be coined and discussed

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/internal_extension/base.py
```diff
@@ -0,0 +1,132 @@
...
+        if module_file.exists():
+            return module_file
+
+    if handler_module.exists():
+        return handler_module
+
+    return None
+
+
+def get_handler_via_module_import(handler: str = HANDLER) -> Handler:
```
medikoo: I think this function should be part of a wrapper.

Initialization logic just resolves the full path to the handler and confirms that the given file exists. If it approaches errors, it should refrain from instrumenting anything and just let AWS report the error.

Wrapping logic should receive from initialization logic final handler location and should initialize it (if initialization crashes, it should ensure that the error is exposed in the same manner as if Lambda was not instrumented).

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/tests/test_internal_extension.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
```
medikoo, line 1: Ideally we should have here tests like Node.js has here: https://github.com/serverless/console/blob/fa1801ac87943becaadb6d7b6fb2ee9342124f68/node/packages/aws-lambda-sdk/test/unit/internal-extension/index.test.js

Those tests set up all needed AWS Lambda environment variables, run internal extension with handler path pointing to pre-prepared lambda fixtures, then run obtained handler and confirm that they operate as expected.

At this stage, we should just prepare a few basic fixtures (success and error case?) and run them through machinery and confirm we have the expected outcome.

In the next task, when we will address trace spans configuration and the trace report will be written, we will also read that payload and confirm it contains the expected information.

Aside of that, as I announced we will also have integration tests which will be configured in Node.js and through real AWS lambdas will confirm the same.
As I think of it, in the scope of this task (as there's no instrumentation) it's probably too early for integration tests (which, in this case, should skip on confirming trace payloads, while it's their main point), so let's agree we will add them once tracing is configured (it'll save us back and forth work). For now just please manually confirm that in the AWS environment published basic cases of lambda work the same with the wrapping as they do without.

medikoo (commented):
Thanks @alexdelorenzo, please see my comments

.github/workflows/python-validate.yml
alexdelorenzo: Part of the wrapping is initializing the SDK in the Node wrapper: https://github.com/serverless/console/blob/main/node/packages/aws-lambda-sdk/internal-extension/wrapper.js#L15

I can remove it if you don't think it's needed right now, though

alexdelorenzo: I removed the local building of deps for the Lambda SDK validation. It will now look for its dependencies on PyPI

.github/workflows/python-validate.yml
alexdelorenzo: Fixed 👍

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/README.md
alexdelorenzo: Okay, I'll remove that. Right now it [depends on these variables being set](https://github.com/alexdelorenzo/console/blob/sc-260/python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/tests/__init__.py#L10)

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/base.py
alexdelorenzo: Thanks, appreciate the detail you went into :) I'll make the request changes

.github/workflows/python-validate.yml
medikoo: > Part of the wrapping is initializing the SDK in the Node wrapper: 

Yes, I forgot to edit that comment after reviewing further. Yes it's good to initialize SDK at this stage

alexdelorenzo: > > Part of the wrapping is initializing the SDK in the Node wrapper:
> 
> Yes, I forgot to edit that comment after reviewing further. Yes it's good to initialize SDK at this stage

Reverted the change that got rid of it 👍

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/internal_extension/base.py
alexdelorenzo: Fixed

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/tests/test_internal_extension.py
alexdelorenzo: >  and run them through machinery and confirm we have the expected outcome.

> For now just please manually confirm that in the AWS environment published basic cases of lambda work the same with the wrapping as they do without.

When you say this, do you mean via tests or literally going into AWS Lambda and running it?

medikoo: > When you say this, do you mean via tests or literally going into AWS Lambda and running it?
 
 I suggested that you test if this wrapping works without issues in AWS environment, manually, as at this stage we won't have integration test configured for it.

medikoo:
Closing as this was addressed with https://github.com/serverless/console/pull/464

PR #384, opened 2023-01-13
SDK:  `serverless.setTag` implementation

medikoo:
_Relies on #383_

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -57,7 +57,7 @@
     "log-node": "^8.0.3",
     "memoizee": "^0.4.15",
     "mocha": "^9.2.2",
-    "ncjsm": "^4.3.1",
+    "ncjsm": "^4.3.2",
     "node-fetch": "^2.6.7",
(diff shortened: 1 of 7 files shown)


Danwakeem (commented):
This would also need to be include in the typescript file and documented in the readme, or is the coming in a follow up PR?

medikoo:
> This would also need to be include in the typescript file and documented in the readme, or is the coming in a follow up PR?

Good call, forgot about that! Added

Danwakeem (commented):
The types look good 😎 Would we also want to add a description for this function in the [sdk docs](https://github.com/serverless/console/blob/main/node/packages/sdk/docs/sdk.md)?

medikoo:
> The types look good 😎 Would we also want to add a description for this function in the sdk docs?

Not sure where my head was :) great you've pointed that. Added

Danwakeem (approved):
Looks great 🚀

PR #385, opened 2023-01-13
AWS Lambda SDK: Print traces in case of unhandled exceptions at invocation phase

medikoo:
I've found a way on how to hack the AWS setup, so we can observe uncaught exceptions and unhandled rejections as a _man in the middle_, and through that, we can print traces to CW logs when they happen.

One side effect is that AWS normally in such situation immediately stops the Lambda, and our injection logic will postpone it by a few ticks, and in the case of dev mode, it'll take longer as we will wait for requests to drain. This characteristic may introduce situations where Lambda crashes differently whether SDK is attached or not (as during a delay introduced by our SDK the second crash may happen in lambda logic and that will crash Lambda immediately). Still, I consider it a rare edge case.

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -113,6 +113,8 @@ const resolveOutcomeEnumValue = (value) => {
       return 1;
     case 'error:handled':
       return 5;
+    case 'error:unhandled':
+      return 3;
     default:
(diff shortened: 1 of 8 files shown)


Danwakeem (approved):
While it will change the lambda crash behavior I think the end result should be the same so seems worth it to me 🤷

Mmarzex (approved):

PR #390, opened 2023-01-16
AWS Lambda SDK: Handle gently non JSON compliant AWS SDK input and output

medikoo:
/cc @garethmcc 

Addresses issue reported by user attempting to use dev mode

```
TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at /opt/sls-sdk-node/index.js:2919:58
    at /var/runtime/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:6:28"
```

It appears user is passing _bigint_ values with AWS SDK payload.

This fix ensures we normalize such values as strings and do not crash on JSON stringification.

With no integration test at this point, as it's not clear to me which AWS service and method supports bigint as input.


Danwakeem (approved):
That is interesting. I am also not aware of which AWS SDK call would support BigInt 🤔

PR #392, opened 2023-01-16
feat: Add fingerprints array to span schema

Mmarzex:
We need to be able to filter spans/traces by the errors/warnings that they generate. This PR Adds a repeated field on the `Span` schema that we can populate on ingest when we have the spans and events together.

### DIFF
--- a/proto/serverless/instrumentation/v1/trace.proto
+++ b/proto/serverless/instrumentation/v1/trace.proto
@@ -74,4 +74,9 @@ message Span {
     // The optional custom tags to be set by the user
     // This is expected to be a JSON object in string format.
     optional string custom_tags = 13;
+
+    // Optional list of event fingerprints related to the span.
+    // This will most likely always be populated by ingest when looking at the events
+    // attached to the TracePayload
+    repeated string fingerprints = 14;
 }
--- a/node/packages/sdk-schema/test/unit/span.test.js
+++ b/node/packages/sdk-schema/test/unit/span.test.js
@@ -33,6 +33,7 @@ const testTracePayload = {
       name: 'test',
       startTimeUnixNano: longValue,
       endTimeUnixNano: longValue,
+      fingerprints: [],
       tags: {
         aws: {
           lambda: {
--- a/node/packages/sdk/lib/trace-span.js
+++ b/node/packages/sdk/lib/trace-span.js
@@ -160,6 +160,7 @@ class TraceSpan {
       output: this.output || undefined,
       tags: this.tags,
       customTags: objHasOwnProperty.call(this, 'customTags') ? this.customTags : undefined,
+      fingerprints: [],
     };
   }
   toProtobufObject() {
@@ -176,6 +177,7 @@ class TraceSpan {
       customTags: objHasOwnProperty.call(this, 'customTags')
         ? JSON.stringify(this.customTags)
         : undefined,
+      fingerprints: [],
     };
   }
   get spans() {
--- a/node/packages/sdk/test/unit/lib/trace-span.test.js
+++ b/node/packages/sdk/test/unit/lib/trace-span.test.js
@@ -143,6 +143,7 @@ describe('lib/trace-span.test.js', () => {
       endTime: jsonValue.endTime,
       tags: { foo: 12 },
       customTags: { elo: 'marko' },
+      fingerprints: [],
     });
   });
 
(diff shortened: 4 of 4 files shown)


proto/serverless/instrumentation/v1/trace.proto
```diff
@@ -74,4 +74,9 @@ message Span {
     // The optional custom tags to be set by the user
     // This is expected to be a JSON object in string format.
     optional string custom_tags = 13;
+
+    // Optional list of event fingerprints related to the span.
+    // This will most likely always be populated by ingest when looking at the events
+    // attached to the TracePayload
+    repeated string fingerprints = 14;
```
Danwakeem, line 81: Just so I am clear on what the plan is does this sound correct 👇


If we receive a trace payload with a set of custom events then ingest will attach the finger prints from each custom event to the root span (or any other span in the list)? Or will we only add specific fingerprints to specific spans?

Danwakeem (dismissed):
Looks good just had a question about the implementation details we had in mind 👍

proto/serverless/instrumentation/v1/trace.proto
Mmarzex: For the immediate term it will be all fingerprints to the root span. Longer term what we will probably do is attach all fingerprints to the root span and then specific fingerprints to the specific span in which the error occured.

Danwakeem: Ahh okay sounds good 👍

Danwakeem: @Mmarzex do we need to mark this as optional? That might be why the unit tests are failing

Mmarzex: @Danwakeem Repeated fields are inherently optional, in the sense that not setting them will encode/decode to an empty array, the tests are failing because of how `ts-proto` works. See [issue](https://github.com/stephenh/ts-proto/issues/225) on their repo. For now we will just do the required updates when we add fields like this. This is also as far as I can see only an issue for the node SDK, it will not be a problem for Python or go.

medikoo (commented):
As I see in the comments, it’s only needed for ingest internals. Do we really have to add it to schema?
Other issue is that it adds a field for information that's already provided with other fields (fingerprints are already present in `trace.events[]`).  So design-wise this addition doesn't look compelling

proto/serverless/instrumentation/v1/trace.proto
Danwakeem: Ahh okay sad times 😢

Mmarzex:
@medikoo It is also used in the user facing portions of the UI. This is a field that is need by `ingest` and then by the user facing APIs for retrieving/filtering data. 

I understand the concern but we made a decision early on to use the same fields across all three areas, so since this is needed in two of them we have to add it. It would require a change to the entire ingest pipeline to change it from sending just `span` typed data to sending a `span & { fingerprints: [] }` type, which I do not want to do. It is a slippery slope to add fields in one part that do not exist everywhere.

Mmarzex:
@medikoo Also we don't currently expose a fingerprint property on the event schema itself, just the `custom_fingerprint` which is not expected to be set all the time. I think this property provides the best solution to minimize query overhead in the whole system.

Danwakeem (approved):
Given we need this when we pass the protobuf around at some point I am cool with adding this. We may wanna wait for @medikoo to reply before sending it though.

medikoo (commented):
> I understand the concern but we made a decision early on to use the same fields across all three areas, so since this is needed in two of them we have to add it. It would require a change to the entire ingest pipeline to change it from sending just span typed data to sending a span & { fingerprints: [] } type, which I do not want to do. It is a slippery slope to add fields in one part that do not exist everywhere.

This signals to me that we're using same schema for different requests, where the same data is expected to be passed. in a different form.

This doesn't look as the right direction to me (ideally those requests should be defined separately). It seems we're introducing a technical debt by building a schema that is hard to reason about.

Is there really no better way to do that?

Mmarzex:
> This doesn't look as the right direction to me (ideally those requests should be defined separately). It seems we're introducing a technical debt by building a schema that is hard to reason about.

How could they be defined separately. If we pull say all `events` from Opensearch that have fingerprint `abc123`, that may return to us 1,000,000 traceIds, we can't actually then query for all traces like that. We need to be able to query the `spans` index in opensearch for fingerprints. This is what this achieves.

medikoo:
> How could they be defined separately. If we pull say all events from Opensearch that have fingerprint abc123, that may return to us 1,000,000 traceIds, we can't actually then query for all traces like that. We need to be able to query the spans index in opensearch for fingerprints. This is what this achieves.

@Mmarzex will moving `events` from `TracePayload` to `Span` improve that? Having that we would have fingerprints at span level, and across TracePayload no data will be duplicated

PR #393, opened 2023-01-17
AWS Lambda SDK: Ensure to not instrument S3 requests signing

medikoo:
Addresses issue reported internally by @ac360 Where relying on `s3.getSignedUrlPromise()` was creating unresolved AWS SDK trace span which produced warning as:

```
Serverless SDK Warning: Following trace spans didn't end before end of lambda invocation
```

It's caused by the fact that AWS SDK for request signing (resolution of preconfigured request URLs) relies on internal `Request` interface. In this case, no real HTTP requests are issued yet our instrumentation is attempting to track these as real requests which leads to errors.


Danwakeem (approved):
Nice find 😎

PR #396, opened 2023-01-18
AWS Lambda SDK: Propagate all captured events to `dev mode`

medikoo:
As all captured events are expected to be shown in a special way, we agreed to propagate all of them, and instead filter related `console.error` and `console.warn` CW logs on dev mode side.

PR #398, opened 2023-01-18
fix: Log console messages as events

Danwakeem:
## Description
According to SC-360 we wanted to log `console.error` and `console.warn` messages as events in dev mode rather than normal logs. This PR combined with [the PR @medikoo submitted](https://github.com/serverless/console/pull/396) this should forward the events rather than `console.warn` and `console.error` logs.

> Note: when I test this manually things appear to work fine but when I am running the integration tests I am getting a proto parsing error on the trace payloads 🤔 I commented out the part of the integration tests but maybe @medikoo would know what I am doing wrong? 🤔

### DIFF
--- a/go/packages/dev-mode/agent/forwarder.go
+++ b/go/packages/dev-mode/agent/forwarder.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"regexp"
 	"serverless/dev-mode-extension/lib"
 	"strings"
 	"time"
@@ -53,6 +54,9 @@ type RuntimeDoneRecord struct {
 	Spans     []TelemetrySpans `json:"spans"`
 	Metrics   MetricsObject    `json:"metrics"`
 }
+
+var wrapper = os.Getenv("AWS_LAMBDA_EXEC_WRAPPER")
+var hasInternalExtension = wrapper == "/opt/sls-sdk-node/exec-wrapper.sh"
 
 func FindRequestId(logs []LogItem) string {
 	var requestId string = ""
@@ -143,6 +147,36 @@ func FindReqData(logs []LogItem) *LogItem {
 	return nil
 }
 
+func IsDefaultConsoleMessage(record interface{}) bool {
+	message := fmt.Sprintf("%v", record)
+	arr := strings.Split(message, "\t")
+	// Not enough items in the list
+	// The first item is not a date timestamp
+	// The second item is not a request id
+	// The third item is not WARN or ERROR
+	if !hasInternalExtension || len(arr) < 4 || !IsValidDate(arr[0]) || !IsValidUUID(arr[1]) || !IsWarnOrError(arr[2]) {
+		return false
+	}
+	return true
+}
+
+func IsWarnOrError(logLevel string) bool {
+	if logLevel == "WARN" || logLevel == "ERROR" {
+		return true
+	}
+	return false
+}
+
+func IsValidDate(date string) bool {
+	_, isValid := time.Parse("2006-01-02T15:04:05.000Z", date)
+	return isValid == nil
+}
+
+func IsValidUUID(uuid string) bool {
+	r := regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
+	return r.MatchString(uuid)
+}
+
 func FormatLogs(logs []LogItem, requestId string, accountId string, traceId string) schema.LogPayload {
 	messages := make([]*schema.LogEvent, 0)
 	platform := "aws"
@@ -158,7 +192,7 @@ func FormatLogs(logs []LogItem, requestId string, accountId string, traceId stri
 		LogEvents: messages,
 	}
 	for i, log := range logs {
-		if log.LogType == "function" {
+		if log.LogType == "function" && !IsDefaultConsoleMessage(log.Record) {
 			t, _ := time.Parse(time.RFC3339, log.Time)
 			if !strings.Contains(log.Record.(string), "SERVERLESS_TELEMETRY.") {
 				// Apparently these environment variables don't
@@ -199,8 +233,6 @@ func CollectRequestResponseData(logs []LogItem, requestId string, accountId stri
 	metadata := make([]*LogItem, 0)
 	hasPlatformStart := FindPlatformStart(logs)
 	hasRuntimeDone := FindRuntimeDone(logs)
-	wrapper := os.Getenv("AWS_LAMBDA_EXEC_WRAPPER")
-	hasInternalExtension := wrapper == "/opt/sls-sdk-node/exec-wrapper.sh"
 	foundReqRes := false
 	for _, log := range logs {
 		if log.LogType == "reqRes" {
--- a/go/packages/dev-mode/main.go
+++ b/go/packages/dev-mode/main.go
@@ -49,6 +49,9 @@ func (e *Extension) ExternalExtension() {
 	printPrefix := fmt.Sprintf("[%s]", extensionName)
 	extensionClient := extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
 
+	// Print environment variables
+	lib.Info(os.Environ())
+
 	// Get account id from sts
 	getAccountId(e.Client)
 
--- a/go/packages/dev-mode/main.go
+++ b/go/packages/dev-mode/main.go
@@ -50,7 +50,6 @@ func (e *Extension) ExternalExtension() {
 	extensionClient := extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
 
 	// Print environment variables
-	lib.Info(os.Environ())
 
 	// Get account id from sts
 	getAccountId(e.Client)
--- a/go/packages/dev-mode/main.go
+++ b/go/packages/dev-mode/main.go
@@ -49,8 +49,6 @@ func (e *Extension) ExternalExtension() {
 	printPrefix := fmt.Sprintf("[%s]", extensionName)
 	extensionClient := extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
 
-	// Print environment variables
-
 	// Get account id from sts
 	getAccountId(e.Client)
 
--- a/node/test/go/dev-mode/test/fixtures/lambdas/with-internal.js
+++ b/node/test/go/dev-mode/test/fixtures/lambdas/with-internal.js
@@ -2,5 +2,7 @@
 
 module.exports.handler = (event, context, callback) => {
   console.log('with-internal 1');
+  console.warn('with-internal warning 1');
+  console.error(new Error('with-internal warning 1'));
   callback(null, 'ok');
 };
--- a/node/test/go/dev-mode/test/fixtures/lambdas/with-internal.js
+++ b/node/test/go/dev-mode/test/fixtures/lambdas/with-internal.js
@@ -3,6 +3,6 @@
 module.exports.handler = (event, context, callback) => {
   console.log('with-internal 1');
   console.warn('with-internal warning 1');
-  console.error(new Error('with-internal warning 1'));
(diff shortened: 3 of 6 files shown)


Danwakeem:
Here is the error I am seeing in the integration test 👇 When I removed the custom events it still results in a similar error 🤔 

Screenshot 2023-01-18 at 12 48 14 PM

medikoo:
> Here is the error I am seeing in the integration test 👇 When I removed the custom events it still results in a similar error 🤔

@Danwakeem have you rebuilt the SDK schema? Maybe there were additions there in a meantime, and that can influence outcome of integration tests

Danwakeem:
@medikoo Yeah I had run through the build steps in the [README in the proto folder](https://github.com/serverless/console/blob/main/proto/README.md) but I just re-ran them just to be sure and I am still seeing it generate a valid trace when I run the internal extension manually but running the integration tests I still get the same error 🤔

medikoo:
Just tested and for current state of `main` integration tests pass for me locally.

There was however crash in the last CI integration build, but as I checked it's due to timeout error on AWS SDK v3 test, so most likely some interim issue on AWS side

Danwakeem:
@medikoo Yeah main is working for me but these tests were not trying to parse the trace payload 🤔 When I try and parse the trace payload from the tests results is where I am running into issues. It looks like the node, aws-lambda-sdk, and dev-mode folder all have the same version of the proto package. So weird.

node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -81,6 +82,10 @@ describe('Integration', function () {
               // Since we are sending spans as they are complete I need to change this to a variable length
               // but we should always have at least 1
               expect(traces.length >= 1).to.equal(true);
+              // For whatever reason the traces are being written in a bad wire format? 🤔
+              // const traceData = Buffer.from(traces[0], 'base64');
+              // const tracePayload = traceProto.TracePayload.decode(traceData);
+              // expect(tracePayload.events.length).to.equal(2);
```
Danwakeem: @medikoo If you pull down this branch, uncomment lines `8` and `86:88`, and run these integration tests locally do you also get the wire format error I am seeing?

I know that is kind of a lot to ask but I am just trying to eliminate a local testing issue here 😅

node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -81,6 +82,10 @@ describe('Integration', function () {
               // Since we are sending spans as they are complete I need to change this to a variable length
               // but we should always have at least 1
               expect(traces.length >= 1).to.equal(true);
+              // For whatever reason the traces are being written in a bad wire format? 🤔
+              // const traceData = Buffer.from(traces[0], 'base64');
+              // const tracePayload = traceProto.TracePayload.decode(traceData);
```
medikoo: Not sure exactly what's `traces` here, but in counterpart in Node.js testing, here you receive already decoded Trace, which is decoded in `process-function` util, in this way: https://github.com/serverless/console/blob/8008eb90bd7c8af375bd830b448666043f484643/node/packages/aws-lambda-sdk/test/lib/process-function.js#L254

Danwakeem: In this context, `traces` is an array of base64 trace strings. I believe what I am doing here is equivalent to what is happening above it just doesn't what to translate the base64 string to a valid proto object for some reason. I tried adding the `.trim()` but that doesn't seem to be the issue either.

medikoo: I plan to refactor out `process-function` from `aws-lambda-sdk` to top `node` folder (so we can also rely on it for Python SDK testing) hopefully tomorrow. I think then it'll be good to update this test to that as well.

If it's not issue in test utils config, it would mean that written report is for some reason broken

Danwakeem: Yeah something is up with this test or build or something because I just [tried forcing the integration tests to run on github in a separate branch](https://github.com/serverless/console/actions/runs/3952735645/jobs/6768192872) and those failed with a similar wire format error 😢 

I guess I am still technically covered on this update without decoding the trace payload and explicitly checking since we are getting the correct number of logs. (If this wasn't working there would be 3 instead of 1) It is of course just less explicit.

Danwakeem: Ahh I am dumb 🤦 I just realized the issue. We are sending trace payloads to the dev mode extension wrapped as a [DevMode Payload](https://github.com/serverless/console/blob/main/proto/serverless/instrumentation/v1/dev_mode.proto) to add additional context. So this is not a `TracePayload` it is `dev_mode` payload.

Totally forgot we were doing that lol I will update these test to reflect this 👍

node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -81,6 +82,10 @@ describe('Integration', function () {
               // Since we are sending spans as they are complete I need to change this to a variable length
               // but we should always have at least 1
               expect(traces.length >= 1).to.equal(true);
+              // For whatever reason the traces are being written in a bad wire format? 🤔
```
Danwakeem:
```suggestion
```

Danwakeem:
Alright @medikoo this should be ready for review now 😎 Thanks for humoring me 😅

go/packages/dev-mode/agent/forwarder.go
```diff
@@ -143,6 +147,36 @@ func FindReqData(logs []LogItem) *LogItem {
...
+	// The third item is not WARN or ERROR
+	if !hasInternalExtension || len(arr) < 4 || !IsValidDate(arr[0]) || !IsValidUUID(arr[1]) || !IsWarnOrError(arr[2]) {
+		return false
+	}
+	return true
+}
+
+func IsWarnOrError(logLevel string) bool {
+	if logLevel == "WARN" || logLevel == "ERROR" {
+		return true
```
medikoo, line 165: Will it filter all `console.error` calls? If so, it can be problematic, as we issue captured events only for those which host Error instance.  So if users log strings with `console.error` they won't appear in dev mode

go/packages/dev-mode/main.go
```diff
@@ -49,6 +49,9 @@ func (e *Extension) ExternalExtension() {
 	printPrefix := fmt.Sprintf("[%s]", extensionName)
 	extensionClient := extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
 
+	// Print environment variables
+	lib.Info(os.Environ())
```
medikoo: Is it printing to CW logs?

node/test/go/dev-mode/test/fixtures/lambdas/with-internal.js
```diff
@@ -2,5 +2,7 @@
 
 module.exports.handler = (event, context, callback) => {
   console.log('with-internal 1');
+  console.warn('with-internal warning 1');
+  console.error(new Error('with-internal warning 1'));
```
medikoo: just a message confusion, as it should probably be `with-internal error..`

go/packages/dev-mode/main.go
Danwakeem:
```suggestion
```
Ahh yeah sorry this should be removed 👍

go/packages/dev-mode/main.go
```diff
@@ -49,6 +49,8 @@ func (e *Extension) ExternalExtension() {
 	printPrefix := fmt.Sprintf("[%s]", extensionName)
 	extensionClient := extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
 
+	// Print environment variables
+
```
Danwakeem:
```suggestion
```

node/test/go/dev-mode/test/fixtures/lambdas/with-internal.js
Danwakeem:
```suggestion
  console.error(new Error('with-internal error 1'));
```

Ahh yes good call 👍

go/packages/dev-mode/agent/forwarder.go
Danwakeem: Ahh yes this is a good point 🤔 

I will do a bit more research for errors

Danwakeem: @medikoo This is actually kind of a problem then. There doesn't seem to be a straight forward way to pull out the error from `console.error` from the message alone. Especially if you pass multiple messages into the error function 👇 

```javascript
console.error('string 1', new Error('This is a bug'), 'string 2');
```

I could log the messages up to when the error object was printed but it would be difficult, maybe impossible to log the messages behind the error 🤔 I'll post this question in our linear ticket.

go/packages/dev-mode/agent/forwarder.go
```diff
@@ -143,6 +147,52 @@ func FindReqData(logs []LogItem) *LogItem {
...
+}
+
+func HasErrorObject(message string) bool {
+	foundErrorDeclaration := strings.Index(message, "Error: ")
+	foundTracePath := strings.Index(message, "/var/task")
+	if foundErrorDeclaration == -1 && foundTracePath == -1 {
+		return false
+	}
+	return true
+}
```
Danwakeem: This is where I am detecting if there is an error object in the error message string 👍

Danwakeem:
Alright should be ready for rereveiw @medikoo 🙏 What do you think about how I am detecting if there is an error object? 🤔

go/packages/dev-mode/agent/forwarder.go
medikoo: @Danwakeem It looks good to me, it's probably best we can do. I have just two concerns:

1. This looks as Node.js specific, while dev mode should be runtime agnostic. Still I think it's fine to assume that at this point we just want to filter those error logs in case of Node.js runtime. Having that assumption, maybe it'll be worth to add an extra constraint which will ensure that this filter works only if we're in context of Node.js Lambda?
2. If there's a scenario where dev-mode works without SDK instrumentation (so we end with Lambda that has dev-mode layer attached but not SDK internal extension), then I think we need to ensure that this filtering is not applied if no SDK extension is present (as no events are sent to dev mode). Is that the case?

medikoo: @Danwakeem I've actually thought a bit more on that, and I think requiring `/var/task` will not pick many error cases. Let's follow with a discussion on Linear

go/packages/dev-mode/agent/forwarder.go
```diff
@@ -143,6 +147,52 @@ func FindReqData(logs []LogItem) *LogItem {
 	return nil
 }
 
+func IsDefaultConsoleMessage(record interface{}) bool {
+	message := fmt.Sprintf("%v", record)
+	arr := strings.Split(message, "\t")
+	// Not enough items in the list
+	// The first item is not a date timestamp
+	// The second item is not a request id
+	if !hasInternalExtension || len(arr) < 4 || !IsValidDate(arr[0]) || !IsValidUUID(arr[1]) {
```
Danwakeem: @medikoo the `hasInternalExtension` check that we have the node internal extension before filtering any logs 👍

go/packages/dev-mode/agent/forwarder.go
Danwakeem: Yeah I agree on both points 👍 both of these points should already be applied in the function that calls this function 😎

Danwakeem:
@medikoo should we go ahead and also push out this version of dev mode so we don't get duplicate events in dev? 🤔 We probably still need Maciej's feedback but this should satisfy the current requirements 🤷

medikoo (commented):
> @medikoo should we go ahead and also push out this version of dev mode so we don't get duplicate events in dev? 🤔 We probably still need Maciej's feedback but this should satisfy the current requirements 🤷

Maybe let's for now do as follows:
- Filter all Console warning logs
- Filter all Error events

So we will address console warning case, but for error we'll wait for feedback and just ensure no duplication for now. WDYT?

Danwakeem:
@medikoo Yeah that sounds like a good plan to me. I will remove the extra error filters. I kind of think the result of that conversation will end with us filtering all errors/warnings anyway 👍

medikoo (approved):
👍

PR #400, opened 2023-01-18
docs: Adding docs on the @serverless/sdk package

skierkowski:
Adding documentation on `@serverless/sdk` package.

This adds documentation on the package, capturing errors/warnings, and tagging.

I intentionally left out fingerprinting for now as there is no UI yet that groups by the fingerprint.

### DIFF
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -50,35 +50,6 @@ Serverless Console in order for the data to be ingested.
 
 [Enable Tracing, Logging, and Dev Mode](/console/docs/integrations/enable-monitoring-features)
 
-**Using a bundler**
-
-If you use a bundler, like esbuild, the AWS Lambda Layer for Serverless Console
-will not be able to auto-instrument traces and spans on your handler. To enable
-auto-instrumentation of spans and traces, you will need to manually add the
-AWS-specific auto-instrumentation library and initiate auto-instrumentation.
-
-First, install the `@serverless/aws-lambda-sdk` package locally. This replaces
-the need for the `@serverless/sdk` package, so you do not need both.
-
-```javascript
-npm install @serverless/aws-lambda-sdk --save
-# or
-yarn add @serverless/aws-lambda-sdk
-```
-
-```javascript
-const serverlessSdk = require("@serverless/aws-lambda-sdk");
-
-// Instrument AWS SDK v2
-serverlessSdk.instrumentation.awsSdkV2.install(AWS)
-
-// Instrument AWS SDK v3 client
-serverlessSdk.instrumentation.awsSdkV3Client.install(client)
-
-// instruments Express.js
-serverlessSdk.instrumentation.expressApp.install(express)
-```
-
 ## Usage
 
 The package does not require any configuration as the credentials are
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -22,6 +22,13 @@ Serverless Console, much like a Captured Error.
 Captured Error or Captured Warning, and sent to Serverless Console. Tags can be
 viewed on the Trace Explorer Details and Dev Mode.
 
+## Compatibility
+
+While Serverless Console is developed by the makers of the Serverless Framework,
+the entire Serverless Console product and this SDK are 100% agnostic of the
+deployment tool you use. Serverless Console and this SDK work just as well with
+Terraform, CDK, SAM, Pulumi, etc, as as they do with Serverless Framework.
+
 ## Installation
 
 **Install the package**
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -36,10 +36,41 @@ If you will run this on AWS Lambda without Serverless Console Tracing enabled,
 or running it on a different runtime, like locally, then you'll need to add the
 `@serverless/sdk` package.
 
-```javascript
+```
 npm install @serverless/sdk --save
 # or
 yarn add @serverless/sdk
+```
+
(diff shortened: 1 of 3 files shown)


skierkowski:
I made this a draft PR as it shouldn't be merged until the SDK, Dev Mode, and Trace Details are released.

docs/nodejs.md
```diff
@@ -0,0 +1,161 @@
...
+The package does not require any configuration as the credentials are
+automatically set on the AWS Lambda function environment variables when Tracing
+is enabled in Serverless Console.
+
+To use the Serverless SDK you must require the `@serverless/sdk` method in your
+AWS Lambda function handler.
+
+```javascript
+const serverlessSdk = require("@serverless/sdk");
+```
```
medikoo, line 105: It's worth here to put some light on difference between `@serverless/sdk` and `@serverless/aws-lambda-sdk`

In Lambda environment importing either will export exactly same SDK. However when working locally, it's only `@serverless/aws-lambda-sdk` which will export AWS Lambda specific instrumentation methods properties, which are documented here: https://github.com/serverless/console/blob/main/node/packages/aws-lambda-sdk/docs/sdk.md (also in VS Code the intellisense for those methods will only be exposed if `@serverless/aws-lambda-sdk` is imported)

So possibly best approach is to rely on `@serverless/sdk` in environment agnostic logic, while for AWS Lambda specific code it's safer to rely on `@serverless/aws-lambda-sdk`

docs/nodejs.md
```diff
@@ -0,0 +1,161 @@
...
+We recommend avoiding using unique instance values for the strings. For example,
+if you need to include a userId, email, request ID, or any ID that may be unique
+to the individual invocation, we recommend using Tagging instead.
+
+### Tagging
+
+**Setting Tags on the Trace**
+
+```javascript
+serverlessSdk.setTag("userId", "bd86489cf036")
```
medikoo, line 174: It would be nice to document here what we support as tag names and values

medikoo (commented):
@skierkowski looks great to me 👍  I have just few suggestions

docs/nodejs.md
skierkowski: I added additional docs on the validation rules on both `setTag` as well as `captureError` and `captureWarning`.

docs/nodejs.md
skierkowski: I added content on using the manual instrumentation; however, we couldn't get the fix in for the manual instrumentation, so I've removed it for now. Next week, when we release the fix I'll publish the updated docs with reference to `@serverless/aws-lambda-sdk` and the manual instrumentation. The proposed changes are in the Linear issue.

skierkowski:
Going to hold off merging this until the SDK is released.

docs/nodejs.md
```diff
@@ -0,0 +1,250 @@
...
+```
+npm install @serverless/sdk --save
+# or
+yarn add @serverless/sdk
+```
+
+**Using a bundler**
+
+If you use a bundler, like esbuild, the AWS Lambda Layer for Serverless Console
+will not be able to auto-instrument traces and spans on your handler. To enable
```
medikoo: The above message may confuse users that instrumentation won't work at all.

Handler and native Node.js APIs (like `http` or `console`) remain instrumented. It's just dependencies that are bundled that will not. So at this point, it applies only to `express` and AWS SDK (assuming the user relies on the bundled version).

docs/nodejs.md
```diff
@@ -0,0 +1,250 @@
...
+
+[Enable Tracing, Logging, and Dev Mode](/console/docs/integrations/enable-monitoring-features)
+
+## Usage
+
+The package does not require any configuration as the credentials are
+automatically set on the AWS Lambda function environment variables when Tracing
+is enabled in Serverless Console.
+
+To use the Serverless SDK you must require the `@serverless/sdk` method in your
```
medikoo, line 100: We may add note, that this package doesn't have to be installed and bundled for Lambda, that the extension will ensure it in the Lambda environment (but also bundling won't break it)

docs/nodejs.md
```diff
@@ -0,0 +1,250 @@
...
+methods will also output a structured-log JSON string, like the one shown
+below.
+
+This string is easier to read, and can also be used with other tools like
+CloudWatch Log Insights to parse and search.
+
+```javascript
+{
+  "source": "serverlessSdk",
+  "type": "ERROR_TYPE_CAUGHT",
```
medikoo: Precisely it'll be `ERROR_TYPE_CAUGHT_USER`

medikoo (commented):
@skierkowski that looks very solid. I'll adapt some of that writing into package documentation (which is visible on npm). I've picked just on few things, please see my comments

docs/nodejs.md
skierkowski: In the **Install the package** section I already had a note about how bundling the package is optional. I added a note in there that bundling won't break it.

Mmarzex (approved):

docs/nodejs.md
medikoo: I've just realized that not bundling the SDK is dangerous due to how we instrument lambda.

It's possible to have integrated lambda published without our extension layer (e.g. if some does redeployment manually and alter the layer section it will remove our extension), then if code attempts to require our SDK it'll crash, and we have no means to mitigate that.

Not sure though what's the best way to communicate that, and how likely such situations are

PR #404, opened 2023-01-19
SDK: Enable customization of event data via private options

medikoo:
Needed to support registering uncaught error events which is needed in context of AWS Lambda SDK


Danwakeem (commented):
I see the options added to the SDK, are we going to use these options in the `aws-lambda-sdk` once this version is published?

medikoo:
> I see the options added to the SDK, are we going to use these options in the aws-lambda-sdk once this version is published?

Yes, exactly, and I have already changes in AWS Lambda SDK ready and tested

Danwakeem (approved):
👍

PR #408, opened 2023-01-19
AWS Lambda SDK: Expose custom tags directly on trace instead of root span

medikoo:
Global custom tags are exposed on TracePayload instead of root span. It's configured both for CW trace report writing and dev mode

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -16,6 +16,8 @@ const pkgJson = require('../package');
 
 const serverlessSdk = require('./lib/sdk');
 
+const objHasOwnProperty = Object.prototype.hasOwnProperty;
+
 const capturedEvents = [];
 serverlessSdk._eventEmitter.on('captured-event', (capturedEvent) =>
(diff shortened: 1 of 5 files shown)


Danwakeem (approved):
This looks good 👍 

But I just realized, do we also need to remove custom tags from the trace in the SDK?

Specifically in these three places 👇 
https://github.com/serverless/console/blob/e13699146d731bc31efe49e0a9f13291becf30d7/node/packages/sdk/lib/trace-span.js#L162

https://github.com/serverless/console/blob/e13699146d731bc31efe49e0a9f13291becf30d7/node/packages/sdk/lib/trace-span.js#L176-L178

https://github.com/serverless/console/blob/e13699146d731bc31efe49e0a9f13291becf30d7/node/packages/sdk/lib/trace-span.js#L214

medikoo:
> But I just realized, do we also need to remove custom tags from the trace in the SDK?

@Danwakeem No there's no need to, as they are never set, and this functionality will be important once we'd allow custom spans with custom tags

Danwakeem:
@medikoo Ahh okay got it 👍

PR #409, opened 2023-01-19
AWS Lambda SDK: Report outcome error data with event instead of root span tags

medikoo:


node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -138,27 +136,16 @@ const closeTrace = async (outcome, outcomeResult) => {
   };
 
   try {
+    const endTime = process.hrtime.bigint();
```
Danwakeem, line 139: Just curious why the `responseStartTime` and `endTime` became one? 🤔

Danwakeem (dismissed):
Looking good 👍

node/packages/aws-lambda-sdk/instrument/index.js
medikoo: Good question, I should have probably split this into another commit.

`responseStartTime` was used to measure the overhead of the extension, and was measured around the same moment as `endTime`, so this is a simplification to rely on single timestamp (below now we write summary by subtracting `endTime` instead of `responseStartTime`)

PR #411, opened 2023-01-19
chore: Release v0.5.6 of dev mode

Danwakeem:
## Description
Ignore `console.error` and `console.warn` logs if we have an internal extension enabled


medikoo (approved):
👍

PR #412, opened 2023-01-20
Test: Generalize integration tests engine

medikoo:
Move and generalize all integration engine modules from `aws-lambda-sdk` package to the root `node` folder.

Having that, itcan be used to test layers in other runtimes aside of Node.js (cc @alexdelorenzo)


Danwakeem (approved):
Nice 👍 I will work on moving the dev mode integration tests over so those will also be using these common files.

PR #415, opened 2023-01-23
chore: Updated dev mode integration tests

Danwakeem:
## Description
Updated the dev mode integration tests to use the shared integration files


node/test/lib/function-actions.js
```diff
@@ -0,0 +1,146 @@
+'use strict';
```
medikoo: @Danwakeem can you list what exactly options you needed (that were not available), so it's refactored as you proposed?

I see the custom reports retrieval (as e.g. `DEV-MODE: Log###`) as one thing. Was there anything else?

medikoo (commented):
@Danwakeem great to see that PR!

Please see my comment. I think there's some duplication introduced with this changes, and probably we can minimize this refactor to something more configurable (instead of overwritable)

node/test/lib/function-actions.js
Danwakeem: What you described is basically everything. Just the logs grouped for traces, reqRes, and logs grouped by those dev mode prefixes strings. 

I guess thinking about this a bit more we could probably just add those dev mode collection strings to the shared process file? 🤔 

My thinking originally was that it might be best if dev mode specific analysis is closer to the dev mode tests but I am not opposed to bringing this into the shared file. 

What do you think?

medikoo: In my eyes logs that describe _internal_ and _external_ extension performance timing should be same in all cases. So best if we stick to one common format that doesn't have to be customized.

Now the question is what extra dev mode extension output need to be read and included in test report, that is not already read in the scope of `process-function` (?) Can you define those?

Danwakeem: Yeah it should just be 👇

Logs - `⚡ DEV-MODE: Log###`
Request/Response - `⚡ DEV-MODE: ReqRes###`
Traces - `⚡ DEV-MODE: Traces###`

I could just add these directly to the shared config without having to change much.

medikoo: What exactly those report? The data that DEV MODE receives from the internal extension?

Danwakeem: Not exactly, it reports the data that will be transported out of dev mode. So instead of sending the data to an API endpoint we just log the data to CW.

medikoo: Ah yes..  so I assume those are regular protobuf payloads.

I think we need to generalize those, so e.g. any log starting with `SERVERLESS_TELEMETRY.` is considered as protobuf payload to retrieve. Then above could start as `SERVERLESS_TELEMETRY.DL`, `SERVERLESS_TELEMETRY.DR` and `SERVERLESS_TELEMETRY.DT`

I just checked and it seems not difficult to adjust `process-function` to do that. Still I've also noticed an issue where we provide custom handling to _trace_ payload, only if it appears in single message which feels as test configuration bug. If you want I can handle fix to that, and at the same time upgrade it so it supports any other protobuf payloads prefixed as proposed above

Danwakeem: Yeah that sounds good to me 👍 I think standardizing these integration test outputs would be helpful. Would you want to add those changes to this PR or would you want to open a separate one?

medikoo: I'll create a new one shortly

Danwakeem: Awesome thanks man 😎 I’ll merge to this branch once that is merged into main

medikoo: Addressed that with https://github.com/serverless/console/pull/417.

After this is in, you'll have automatically provided any protobuf payload (recognized with `SERVERLESS_TELEMETRY.<name>.` starting message).

If you want you may bake in payload decoding into generic util, by adding handling for it, in this switch clause: https://github.com/serverless/console/blob/106bf2e47fa8615a64c9312cf3a4edb71037f1c9/node/test/lib/get-process-function.js#L226

Danwakeem:
Okay @medikoo I finally got a change to update this with the changes from #417 😅  Let me know what you think 😎

node/test/lib/function-actions.js
medikoo: Do we need that refactor?

node/test/lib/get-process-function.js
```diff
@@ -257,6 +136,26 @@ module.exports = async (basename, coreConfig, options) => {
               });
             }
             break;
+          case 'DD':
```
medikoo: What`DD` stands for?  `DM` could be for "Dev Mode" :)

medikoo (commented):
Thanks @Danwakeem, we're near there :)

node/test/lib/get-process-function.js
Danwakeem: I was just thinking `DD` since it is Dev Mode logging a Dev Mode payload. But I do think `DM` looks a bit better than `DD` 😎

node/test/lib/function-actions.js
Danwakeem: No I don't think we need it anymore, I can back this out 👍

node/test/lib/get-process-function.js
```diff
@@ -35,9 +35,10 @@ module.exports = async (basename, coreConfig, options) => {
       },
       Timeout: 15,
     },
-    ensurePlainObject(options.baseLambdaConfiguration)
+    ensurePlainObject(options.baseLambdaConfiguration || {})
```
medikoo: This I think should stay as mandatory option

node/test/lib/get-process-function.js
```diff
@@ -168,7 +169,7 @@ module.exports = async (basename, coreConfig, options) => {
     let invocationsData;
     let processesData;
     do {
-      let events = await retrieveAllEvents();
+      let events = await retrieveAllEvents(testConfig);
```
medikoo: This change shouldn't be needed

medikoo (commented):
Thanks @Danwakeem, last few minor things :)

node/test/lib/get-process-function.js
Danwakeem: Ahh yeah my bad 🤦‍♂️

node/test/lib/get-process-function.js
Danwakeem: Ahh yeah my bad 🤦‍♂️

medikoo (approved):
Looks great 👍

Danwakeem:
Thanks @medikoo for taking the time to pull out these shared files 😎

PR #417, opened 2023-01-24
Test: Collect any printed protobuf payload

medikoo:
- Store any printed protobuf payloads in `printedPayloads` collection on invocations data
- Fix issue where custom handling was not applied to trace payload if it was split into multiple messages

### DIFF
--- a/node/test/lib/get-process-function.js
+++ b/node/test/lib/get-process-function.js
@@ -218,6 +218,48 @@ module.exports = async (basename, coreConfig, options) => {
         }
         return currentInvocationData;
       };
+      const finalizePayload = (name, payloadString) => {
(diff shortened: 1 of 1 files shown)


Danwakeem (approved):
This looks good 👍 So I will probably just want to add more cases to the `finalizePayload` switch in my PR to support the dev mode payloads, or do you think I should just use the `printedPayloads` and parse out the data in the dev mode integration tests?

medikoo:
> This looks good 👍 So I will probably just want to add more cases to the finalizePayload switch in my PR to support the dev mode payloads, or do you think I should just use the printedPayloads and parse out the data in the dev mode integration tests?

Either will be fine I think. If you feel it's more comfortable to have decoding configured in util it can go there (those payloads are not specific to dev mode)

PR #419, opened 2023-01-25
SDK: Store stacktrace with each captured event

medikoo:


medikoo:
@Danwakeem we've got a green light form @skierkowski

node/packages/sdk/lib/resolve-stack-trace-string.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const resolveStackString = (error) => {
+  if (!error) return null;
```
Danwakeem, line 4: Is it correct that we would store the stacktrace as `null` if we don't have an error object?

medikoo: No, in this case we fallback to locally resolved stack (this function is re-run with error instance created on spot

Danwakeem: Ahh okay awesome 😎

Danwakeem (approved):
:shipit:

PR #425, opened 2023-01-27
AWS Lambda SDK: Ensure to not propagate root span without requestId tag do dev mode

medikoo:
As we unconditionally require `aws.lambda.requestId` tag by schema on root span, we cannot report root span if lambda crashes at initialization.

This was taken into account in report writing, but it's not taken into account with dev mode, this patch fixes that

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js
+++ b/node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js
@@ -38,6 +38,11 @@ const sendData = () => {
   };
   pendingSpans.length = 0;
   pendingCapturedEvents.length = 0;
+  if (!invocationContextAccessor.value) {
+    // Root span comes with "aws.lambda.*" tags, which require unconditionally requestId
+    // which we don't have if handler crashed at initialization.
+    payload.spans = payload.spans.filter((span) => span.name !== 'aws.lambda');
+  }
   serverlessSdk._deferredTelemetryRequests.push(
     sendTelemetry('trace', traceProto.TracePayload.encode(payload).finish())
   );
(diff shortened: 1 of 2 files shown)


node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js
```diff
@@ -38,6 +38,11 @@ const sendData = () => {
   };
   pendingSpans.length = 0;
   pendingCapturedEvents.length = 0;
+  if (!invocationContextAccessor.value) {
+    // Root span comes with "aws.lambda.*" tags, which require unconditionally requestId
+    // which we don't have if handler crashed at initialization.
+    payload.spans = payload.spans.filter((span) => span.name !== 'aws.lambda');
```
Danwakeem, line 44: Does this mean the root span will never get propagated when there is an initialization error or does this mean that it will be held until we receive a requestId? 🤔

Danwakeem (approved):
😎

node/packages/aws-lambda-sdk/instrument/lib/auto-send-spans.js
medikoo: It will never be propagated, as lambda is restarted at this point (we will never know the request id)

Danwakeem: Ahh okay got it, this makes sense. Thanks for the clarification 😎

medikoo: We probably need to discuss this request-id requirement, for now it's best we can do

Danwakeem: Yeah for sure. I am going through some code clean up on the dev mode extension right now and I am trying to think through some options where we don't have to rely on it at all and can just rely on the `traceId`.

I think it is possible I just need to do a few experiments to test a few theories. Maybe we can set up a time to discuss next week sometime?

medikoo: Yes, discussing it next week is a good idea

PR #429, opened 2023-01-30
AWS Lambda SDK: Fix AWS SDK error reporting

medikoo:
1. Fix error reporting on AWS SDK v3 side. It appears `aws.sdk.error` tag was not set in case of error
2. Fix handling of special error case, as observed on AWS SDK v2, where `error.message` equals `null` (reported to AWS SDK at https://github.com/aws/aws-sdk-js/issues/4330)


Danwakeem (approved):
Dang that is annoying this one API error is different from the others 🙄 nice job tracking this down 😎

PR #430, opened 2023-01-30
AWS Lambda SDK: Fix invocation closure race condition

medikoo:
When reproducing error addressed with https://github.com/serverless/console/pull/429 I've observed an error situation where invocation closure is processed twice for same invocation.
It appears we didn't have prevention for scenario when uncaught exception happens and right after invocation resolves. This PR fixes that


Danwakeem (approved):
Awesome 😎

PR #435, opened 2023-01-30
AWS Lambda SDK: Prevent internal crashes

medikoo:
Follow up to #431 

- Reconfigure error logs that inform about crashes to rely on internal `_reportSdkError` util
- Wrap certain instrumentation parts of AWS SDK to prevent crashing
- Wrap SDK initialization to prevent crashing
- Ensure that SDK crashes are exposed as crashes in test runs
- Clear left over references to `global.serverlessSdk`
- Ensure to not register uncaught errors listeners in offline unit test run


Danwakeem (approved):
:shipit:

PR #437, opened 2023-01-30
fix: Fixed traceId on generated responses

Danwakeem:
## Description
If the function timed out (or really errored in some other unhandled way) the final generated response event would not have the correct traceId attached 🤦‍♂️


medikoo (approved):
👍

PR #440, opened 2023-02-01
Schema: Distinguish user and sdk caught errors

medikoo:
Dictionary needed for error reporting on the SDK side

- `ERROR_TYPE_CAUGHT_USER` - error reported by user via `serverlessSdk.captureError(..)` or `console.error(..)` call
- `ERROR_TYPE_CAUGHT_SDK_USER` - error reported by the SDK that reports SDK misuage by user (e.g. attempt to set an invalid tag)
- `ERROR_TYPE_CAUGHT_SDK_INTERNAL` - error reported by the SDK that reports SDK internal failure which was silenced to not crash user's lambda

### DIFF
--- a/proto/serverless/instrumentation/tags/v1/error.proto
+++ b/proto/serverless/instrumentation/tags/v1/error.proto
@@ -18,12 +18,23 @@ message ErrorTags {
 
         // An unexpected error that caused the application to fail
         ERROR_TYPE_UNCAUGHT = 1;
-       
-        // An error that was reported via the Serverless SDK.
-        // Error that doesn't explicitly fail the application.
-        // Multiple errors of this type can be reported during a single application run
-        ERROR_TYPE_CAUGHT = 2;
 
+        // An error that was reported by user explictly via the Serverless SDK or console.error call
+        // Error doesn't explicitly fail the application.
+        // Multiple errors of this type can be reported during a single application run
+        ERROR_TYPE_CAUGHT_USER = 2;
+
+        // An error that was reported by the Serverless SDK internally that reports user error
+        // (misuage of the SDK)
+        // Error doesn't explicitly fail the application.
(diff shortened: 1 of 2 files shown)


proto/serverless/instrumentation/tags/v1/error.proto
```diff
@@ -18,12 +18,23 @@ message ErrorTags {
...
         // An unexpected error that caused the application to fail
         ERROR_TYPE_UNCAUGHT = 1;
-       
-        // An error that was reported via the Serverless SDK.
-        // Error that doesn't explicitly fail the application.
+
+        // An error that was reported by user explictly via the Serverless SDK or console.error call
+        // Error doesn't explicitly fail the application.
+        // Multiple errors of this type can be reported during a single application run
+        ERROR_TYPE_CAUGHT_USER = 2;
```
Mmarzex, line 25: Let's leave this as `ERROR_TYPE_CAUGHT`, we don't want to have to change any existing data that is already in prod. 

I think it is good to have a catchall `CAUGHT` and `UNCAUGHT` and we can add other types as needed for specific usecases.

Mmarzex (changes requested):

proto/serverless/instrumentation/tags/v1/error.proto
medikoo: > Let's leave this as ERROR_TYPE_CAUGHT, we don't want to have to change any existing data that is already in prod.

@Mmarzex  Value remains unchanged. it's the enum name that's changed, prod infra relies on those name references?

Mmarzex: Good point, we don't rely on the names in ingest, we do have some data in prod that has the name already parsed into clickhouse, but I don't believe it will be harmful. We can go ahead and change this.

Mmarzex (approved):

PR #452, opened 2023-02-06
feat: Added traceId when SDK is not present

Danwakeem:
## Description
Added traceId to all payloads even if the SDK is not present on the customer's lambda function

### DIFF
--- a/go/packages/dev-mode/agent/forwarder.go
+++ b/go/packages/dev-mode/agent/forwarder.go
@@ -77,8 +77,7 @@ func FindTraceId(logs []LogItem) string {
 	// If we don't have the internal extension included
 	// then we should generate one
 	if hasInternalExtension == false {
-		id := uuid.New()
-		return id.String()
+		return uuid.NewString()
 	}
 
 	for _, log := range logs {
@@ -122,13 +121,13 @@ func FindRuntimeDone(logs []LogItem) *LogItem {
 	return nil
 }
 
-func StringIncludesInitError(str string) bool {
+func stringIncludesInitError(str string) bool {
 	return strings.Contains(str, "\tundefined\tERROR") && strings.Contains(str, "\tundefined\tUncaught Exception")
 }
 
 func FindInitErrorLog(logs []LogItem) *LogItem {
 	for _, log := range logs {
-		if log.LogType == "function" && StringIncludesInitError(log.Record.(string)) {
+		if log.LogType == "function" && stringIncludesInitError(log.Record.(string)) {
 			return &log
 		}
 	}
--- a/go/packages/dev-mode/agent/forwarder.go
+++ b/go/packages/dev-mode/agent/forwarder.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	tags "go.buf.build/protocolbuffers/go/serverless/sdk-schema/serverless/instrumentation/tags/v1"
 	schema "go.buf.build/protocolbuffers/go/serverless/sdk-schema/serverless/instrumentation/v1"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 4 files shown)


medikoo (dismissed):
@czubocha I've requested a review from you, as you're familiar with Go, and reviewing this can (1) help you better understand our internals (2) help us get more experienced eyes on Go codebase :)

go/packages/dev-mode/agent/forwarder.go
```diff
@@ -72,6 +73,14 @@ func FindRequestId(logs []LogItem) string {
 
 func FindTraceId(logs []LogItem) string {
 	var traceId string = ""
+
+	// If we don't have the internal extension included
+	// then we should generate one
+	if hasInternalExtension == false {
+		id := uuid.New()
+		return id.String()
```
czubocha:
```suggestion
		return uuid.NewString()
```

go/packages/dev-mode/main.go
```diff
@@ -58,8 +57,10 @@ func (e *Extension) ExternalExtension() {
 	signal.Notify(sigs, syscall.SIGTERM, syscall.SIGINT)
 	go func() {
 		s := <-sigs
-		cancel()
-		s.String()
+		lib.Info("Received signal: %s %v", s.String(), cancel)
```
czubocha:
```suggestion
		lib.Info("Received signal: %s", s.String())
```
`cancel` is a function so it won't print anything meaningful

go/packages/dev-mode/main.go
```diff
@@ -91,8 +92,9 @@ func (e *Extension) ExternalExtension() {
 		for !(logQueue.Empty() && (force || receivedRuntimeDone)) {
 			logs, err := logQueue.Get(1)
 			if err != nil {
-				logger.Error(printPrefix, zap.Error(err))
-				return
+				lib.Error(err)
+				time.Sleep(time.Millisecond * 50)
+				continue
```
czubocha, line 97: have you encountered errors here or just you just adding it just in case?

go/packages/dev-mode/agent/forwarder.go
```diff
@@ -113,6 +122,19 @@ func FindRuntimeDone(logs []LogItem) *LogItem {
...
+}
+
+func FindInitErrorLog(logs []LogItem) *LogItem {
+	for _, log := range logs {
+		if log.LogType == "function" && StringIncludesInitError(log.Record.(string)) {
+			return &log
+		}
+	}
+	return nil
+}
```
czubocha:
```suggestion
func stringIncludesInitError(str string) bool {
	return strings.Contains(str, "\tundefined\tERROR") && strings.Contains(str, "\tundefined\tUncaught Exception")
}

func FindInitErrorLog(logs []LogItem) *LogItem {
	for _, log := range logs {
		if log.LogType == "function" && stringIncludesInitError(log.Record.(string)) {
			return &log
		}
	}
	return nil
}
```
`stringIncludesInitError` doesn't need to be exported (public)

go/packages/dev-mode/main.go
```diff
@@ -145,28 +147,25 @@ func (e *Extension) ExternalExtension() {
 				requestId = agent.FindRequestId(arr)
 			}
 
+			hasInitErrorLog := agent.FindInitErrorLog(arr) == nil
```
czubocha:
```suggestion
			hasInitErrorLog := agent.FindInitErrorLog(arr) != nil
```
shouldn't be `!=`?  🤔 when `== nil` then init error was not found

go/packages/dev-mode/main.go
Danwakeem: Yeah TBH I was mostly doing this just to get rid of the warning 🤷‍♂️

go/packages/dev-mode/main.go
Danwakeem: Lol my bad I probably flipped this to do some tests and forgot to flip it back. Thanks for noticing 🙏

go/packages/dev-mode/main.go
Danwakeem: Yeah if we return here then it is possible we exit too early in the event of an initialization crash 👍

czubocha (approved):
looks good now!

Danwakeem:
Thanks @czubocha 🙏 I will start tagging you on my PRs.

Also if you have any feedback on changes I should make to the code base I would really appreciate it. I am still somewhat new to Golang so I might not be doing things the best.

czubocha:
sure! we'll work on it 👍

PR #453, opened 2023-02-07
Schema: Notice event tags

medikoo:


Danwakeem (dismissed):
LGTM 😎

Mmarzex (changes requested):
I think this looks good @medikoo , but let's have @skierkowski take a look too since the naming we pick here will affect UX potentially.

Danwakeem (approved):
Looks good to me 👍

proto/serverless/instrumentation/tags/v1/notice.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+package serverless.instrumentation.tags.v1;
+
+option go_package = ".;protoc";
+
+message NoticeTags {
+    // Notice Message
+    string message = 1;
```
Mmarzex, line 9: If we are going to have a single type here, can we get it written out what possible messages will be here given the shortened set of NoticeType's?

Mmarzex (changes requested):

proto/serverless/instrumentation/tags/v1/notice.proto
medikoo: @max Do you mean to list here all the messages we configure in code (like `Binary body was removed`) and ensure this list is updated whenever we update the grammar of these messages or add new ones?

medikoo: Ok, I think you rather mean listing fingerprints (message codes), as we should never rely on messages programmatically (they're either not used, or human readable messages for users)

Mmarzex (approved):

PR #454, opened 2023-02-08
AWS Lamdba SDK: ESM support related test coverage improvements

medikoo:
Added test coverage for following cases:
- Ensure that automatic instrumentation works for ESM import of `express`, `aws-sdk` `@aws-sdk/*` and `http`
- Ensure that manual instrumentation of `express` and AWS SDK v2 and v3 works in context of bundles
- Ensure that import of `@serverless/aws-lambda-sdk` works in context of native ESM module - It's supported starting from `nodejs18.x` runtime, as it's where AWS ensured that `NODE_PATH` is supported for ESM imports

### DIFF
--- a/node/lib/run-esbuild.js
+++ b/node/lib/run-esbuild.js
@@ -1,0 +1,14 @@
+'use strict';
+
+const path = require('path');
+const spawn = require('child-process-ext/spawn');
+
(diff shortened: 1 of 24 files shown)


Danwakeem (approved):
Looking good 👍

PR #456, opened 2023-02-09
SDK: Interface for notice reporting and large body stripping

medikoo:
- Internal `_reportNotice` interface
- Remove `traceCaptureBodySizeKb` in favor of internal hard coded `_maximumBodyByteLength` (this setting should not be open for customization as it may introduce payloads too large to process
- Do not strip non JSON bodies (limit stripping just to binary and too large payloads)
- Report notice on large HTTP request and response bodies


Danwakeem (approved):
🎸

PR #458, opened 2023-02-10
AWS Lamda SDK: Upgrade input/output body stripping

medikoo:
- Ensure not to resolve request/response payload when it's not meant to send
- Strip binary lambda request
- Strip binary lambda large requests and responses
- Strip AWS SDK large requests and responses
- Notify with notice events body stripping

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -54,7 +54,9 @@ const resolveResponseString = (response) => {
       response = { ...response };
       if (response.isBase64Encoded) {
         delete response.body;
-        response.isBodyExcluded = true;
+        serverlessSdk._reportNotice('Binary body excluded', 'OUTPUT_BODY_BINARY', {
+          _traceSpan: awsLambdaSpan,
+        });
       }
     }
   }
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -28,6 +28,22 @@ const { awsLambda: awsLambdaSpan, awsLambdaInitialization: awsLambdaInitializati
 
 const toProtobufEpochTimestamp = awsLambdaSpan.constructor._toProtobufEpochTimestamp;
 
+const resolveBodyString = (data, prefix) => {
+  if (data === undefined) return null;
+  if (awsLambdaSpan.tags.get('aws.lambda.event_source') === 'aws.apigateway') {
+    if (typeof data.body === 'string') {
+      data = { ...data };
+      if (data.isBase64Encoded) {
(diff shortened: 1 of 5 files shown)


node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -28,6 +28,31 @@ const { awsLambda: awsLambdaSpan, awsLambdaInitialization: awsLambdaInitializati
...
+        serverlessSdk._reportNotice('Binary body excluded', `${prefix}_BODY_BINARY`, {
+          _traceSpan: awsLambdaSpan,
+        });
+      }
+    }
+  }
+  if (
+    data.body &&
+    Buffer.byteLength(JSON.stringify(data.body)) > serverlessSdk._maximumBodyByteLength
+  ) {
```
Danwakeem: Wouldn't we also want to check if `data` by itself is too large? I am thinking of the scenario when you might be passing in a custom object as an input to your lambda function.

Danwakeem (commented):
One question 👍

node/packages/aws-lambda-sdk/instrument/index.js
medikoo: @Danwakeem that's a very good point. We should actually validate the whole request size, and if it's too large, return null. I've updated PR

node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -28,6 +28,29 @@ const { awsLambda: awsLambdaSpan, awsLambdaInitialization: awsLambdaInitializati
 
 const toProtobufEpochTimestamp = awsLambdaSpan.constructor._toProtobufEpochTimestamp;
 
+const resolveBodyString = (data, prefix) => {
+  if (data === undefined) return null;
+  data = { ...data };
```
Danwakeem:
```suggestion
```
Do we need to do a spread here?

node/packages/aws-lambda-sdk/instrument/index.js
Danwakeem: Ahh okay awesome just left one more comment 🙏

node/packages/aws-lambda-sdk/instrument/index.js
medikoo: Indeed more optimal would be to make it only when we strip the binary body (and in that case is needed, as the same object is processed by user code). Updated

Danwakeem: Ahh got it, you wanted to do a copy when you `delete data.body` that way you do not mutate the input param and only effect the output of the function. I think where you moved this spread to makes the most sense 👍

Danwakeem (approved):
Looking good :shipit:

PR #462, opened 2023-02-17
feat(Go SDK): Record basic lambda spans and custom error events

czubocha:
* flow: wrap original user handler -> register initialization, invocation start time -> create span and add to context -> allow user to capture custom event -> register invocation end time -> marshal proto -> print telemetry log
* example of use in `example` directory
* simple documentation in `README.md`

public interface:
* `Start(handler any, options ...Option)`
* `WithEnvironment(env string) Option`
* `CaptureError(ctx context.Context, err error)`
* `CaptureWarning(ctx context.Context, msg string)`

image
image
image
image


Mmarzex (changes requested):
Looks really good, a few things, 

1. We should be ensuring we are using conventional commit messages so that we can auto-generate changelogs.
2. Can we add support for sending warning events as well? I can see that being useful initially.
3. Are we able to support uncaught panics from the Go SDK if we use `recover`?  
4. If we could have some simple documentation included for now that has what needs set in environment variables, as well as points to the example you wrote, that'd be great.

czubocha:
I've added warnings support, panic recovery and a short readme based on [NodeJS one ](https://github.com/serverless/console/blob/main/docs/nodejs.md)

Mmarzex (approved):
Looks great 👍 

I think for now this is good since we will only be using it internally at first, but let's think about how to get a good PR workflow with integration tests similar to the Node SDK prior to supporting this for our users.

PR #464, opened 2023-02-20
SDK: SC-260 Package with a handler wrapper (latest version)

selcukcihan:
### Description
Related issue https://linear.app/serverless/issue/SC-260/aws-lambda-sdk-setup-package-with-a-handler-wrapper

This PR builds on https://github.com/serverless/console/pull/382 by
* Fixing imports & handler path
* Publishing serverless-aws-lambda-sdk on PyPI
* Publishing sls-sdk-python lambda layer on AWS Lambda, containing two entities:
  * The SDK
  * Internal extension (`AWS_LAMBDA_EXEC_WRAPPER ` should be set to `/opt/sls-sdk-python/exec_wrapper.py`)

### Testing done
* I've tested the lambda layer part on my GitHub account, which published the layer to AWS Lambda and I was able to attach the layer to my test function and successfully run the test function.

### Out of Scope
To release the SDK and the layer, we need to publish our python packages to PyPI one by one. That's the only way to actually validate the GitHub flow end to end when we finally cut the tag for serverless-aws-lambda-sdk.

PR #466, opened 2023-02-21
Fix publish SDK CI workflows

selcukcihan:
### Description
Related issue https://linear.app/serverless/issue/SC-260/aws-lambda-sdk-setup-package-with-a-handler-wrapper
This PR addresses the suggestion https://github.com/serverless/console/pull/464#discussion_r1113056600

* Make sure CI detects changes in sdk-schema when releasing these packages.
* Make sure only the last version of package is processed (this was causing errors if there were more than two versions since last publish, which ideally should not happen but happened to me while testing, so being defensive just in case)
* Checkout the repo when publishing sdk
* Fix naming: serverless_sdk_schema
* Remove redundant python installation step
* Use python 3.9 for the validate workflow (I'll update other flows in subsequent PRs to use 3.9 instead of the soon to be "end of support" 3.7)

### Testing done
Tested by imitating these flows on my GitHub account using different package names.

### DIFF
--- a/.github/workflows/python-integrate.yaml
+++ b/.github/workflows/python-integrate.yaml
@@ -14,7 +14,6 @@ jobs:
     name: "[Linux] Python 3.7: Unit Tests"
     runs-on: ubuntu-latest
     outputs:
-      pathChangedAwsLambdaSdk: ${{ steps.pathChanges.outputs.awsLambdaSdk }}
       pathChangedSdk: ${{ steps.pathChanges.outputs.sdk}}
       pathChangedSdkSchema: ${{ steps.pathChanges.outputs.sdkSchema}}
       pathChangedProto: ${{ steps.pathChanges.outputs.proto }}
@@ -36,8 +35,6 @@ jobs:
         id: pathChanges
         with:
           filters: |
-            awsLambdaSdk:
-              - 'python/packages/aws-lambda-sdk/**'
             sdk:
               - 'python/packages/sdk/**'
             sdkSchema:
@@ -102,20 +99,6 @@ jobs:
           python3 -m pip install poethepoet pytest
           python3 -m poethepoet test
 
-      - name: Install AWS Lambda SDK project and dependencies
-        if: steps.pathChanges.outputs.awsLambdaSdk == 'true'
-        run: |
-          cd python/packages/aws-lambda-sdk
-
-          python3 -m pip install --editable .
-
-      - name: Unit tests /aws-lambda-sdk
-        if: steps.pathChanges.outputs.awsLambdaSdk == 'true'
-        run: |
-          cd python/packages/aws-lambda-sdk
-          python3 -m pip install pytest
-          python3 -m pytest
-
   integratePythonSdk:
     name: Integrate Python SDK
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/python-integrate.yaml
```diff
@@ -14,7 +14,9 @@ jobs:
     name: "[Linux] Python 3.7: Unit Tests"
     runs-on: ubuntu-latest
     outputs:
+      pathChangedAwsLambdaSdk: ${{ steps.pathChanges.outputs.awsLambdaSdk }}
```
medikoo: I think this doesn't belong here (yet)

.github/workflows/python-integrate.yaml
```diff
@@ -34,6 +36,8 @@ jobs:
         id: pathChanges
         with:
           filters: |
+            awsLambdaSdk:
```
medikoo: Same this and other AWS SDK Lambda references

medikoo (changes requested):
Thanks @selcukcihan ! Please see my comments.

When PR is ready for re-review, all you need to do is to re-request review in reviewers box (while it's not possible to request first review, it's possible to re-request reviews afterwards)

.github/workflows/python-integrate.yaml
selcukcihan: What's the reasoning behind it? I think it would be safe to introduce these. Is it because it concerns aws lambda sdk which is not yet in place? Would you like to capture this change in the original PR? Please let me know 🙏

medikoo: Yes, logically it doesn't make sense here, when this package is not yet part of the project.

This should be included in PR where `aws_lambda_sdk` package is introduced

medikoo (commented):
I see review being re-requested, but no changes pushed. Are you sure you've pushed changes in your repo?

selcukcihan:
> I see review being re-requested, but no changes pushed. Are you sure you've pushed changes in your repo?

I did that to make sure you get notified for my comment https://github.com/serverless/console/pull/466#discussion_r1113156180

I'll make sure I only request review when there's a code change, as it seems you already get notified when there's a comment.

I've removed the lambda-sdk related changes from this PR, marked as ready for review again 🙏

.github/workflows/python-integrate.yaml
```diff
@@ -70,10 +71,13 @@ jobs:
...
         uses: bufbuild/buf-setup-action@v1
+      - name: Buf Lint
+        if: steps.pathChanges.outputs.proto == 'true'
+        uses: bufbuild/buf-lint-action@v1
         with:
-          github_token: ${{ github.token }}
-
+          input: "proto"
       - name: Build Protobufs
+        if: steps.pathChanges.outputs.sdkSchema == 'true' || steps.pathChanges.outputs.proto == 'true'
```
medikoo, line 80: Good call with that. 

As I look, I believe the same `if` condition should be put for `Install SDK Schema` and `Run SDK Schema unit tests`

.github/workflows/python-sdk-schema-publish.yaml
```diff
@@ -1,28 +1,19 @@
-name: "Python: Publish python/serverless_aws_schema_sdk"
+name: "Python: Publish python/serverless_sdk_schema"
 
 on:
   push:
     tags:
-      - "python/serverless_aws_schema_sdk@[0-9]+.[0-9]+.[0-9]+"
+      - "python/serverless_sdk_schema@[0-9]+.[0-9]+.[0-9]+"
```
medikoo, line 6: Great catch!

medikoo:
> I did that to make sure you get notified for my comment #466 (comment)

I receive email notifications on every comment. No need to re-request review for that 👍 

There can be the case when you may re-request review without pushed changes. it's when you feel that reviewer suggestions are not applicable for some reason. Then please comment with explanations and re-request the review

medikoo (approved):
Looks great 👍 

I think we can follow up now with PR that will provide basic documentation for SDK ?

PR #468, opened 2023-02-22
Fix pyproject.toml formatter

selcukcihan:
### Description
Currently linting step in the CI fails as in https://github.com/serverless/console/actions/runs/4240504929/jobs/7369594955#step:8:53

The problem is that new versions of `pyproject-fmt` is normalizing the package names and rewrites "serverless_sdk" as "serverless-sdk" and returns an error code:

```
 [project]
-name = "serverless_sdk"
+name = "serverless-sdk"
```

I've did a quick search to see if there's a better linter for pyproject.toml files but couldn't find one. Removing this linter step is another possible resolution. I've also removed the dependency for `pyproject-fmt` from the package as it's only required for the CI step, which installs the specific version of the package.

This fix is in https://github.com/serverless/console/pull/464 as well, I'll update that PR once we agree on a resolution.

### Testing done
I've tried `pyproject-fmt==0.4.1` and this version accepts underscores as in `serverless_sdk`.

### DIFF
--- a/.github/workflows/python-integrate.yaml
+++ b/.github/workflows/python-integrate.yaml
@@ -121,7 +121,7 @@ jobs:
           NEW_VERSION=`git diff -U0 ${{ github.event.before }} python/packages/sdk/pyproject.toml | grep 'version = ' | tail -n 1 | grep -oE "[0-9]+\.[0-9]+\.[0-9]+"` || :
           if [ -n "$NEW_VERSION" ] && [ $NEW_VERSION != "0.0.0" ];
           then
-            git tag python/serverless_sdk@$NEW_VERSION
+            git tag python/serverless-sdk@$NEW_VERSION
             git push --tags
           fi
 
@@ -147,6 +147,6 @@ jobs:
           NEW_VERSION=`git diff -U0 ${{ github.event.before }} python/packages/sdk-schema/pyproject.toml | grep 'version = ' | tail -n 1 | grep -oE "[0-9]+\.[0-9]+\.[0-9]+"` || :
           if [ -n "$NEW_VERSION" ] && [ $NEW_VERSION != "0.0.0" ];
           then
-            git tag python/serverless_sdk_schema@$NEW_VERSION
+            git tag python/serverless-sdk-schema@$NEW_VERSION
             git push --tags
           fi
--- a/.github/workflows/python-sdk-publish.yaml
+++ b/.github/workflows/python-sdk-publish.yaml
@@ -1,9 +1,9 @@
-name: "Python: Publish python/serverless_sdk"
+name: "Python: Publish python/serverless-sdk"
 
 on:
   push:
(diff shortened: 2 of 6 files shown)


medikoo (dismissed):
Great call, thanks for thorough explanations @selcukcihan 👍

medikoo (commented):
@selcukcihan actually, maybe if linter changes the package names to `-` separated, maybe it's the convention we should use (?)

I've just browsed PyPI, and it seems that `-` is used as convention

WDYT?

selcukcihan:
I've found this https://peps.python.org/pep-0621/#table-name regarding the use of `[project]` where "name" should follow the pattern described in https://packaging.python.org/en/latest/specifications/core-metadata/#name

https://github.com/tox-dev/pyproject-fmt/issues/55 is related to the issue we have with pyproject-fmt, I don't understand why the tool is normalizing the project name and rewriting it as described in https://packaging.python.org/en/latest/specifications/name-normalization/ Because as I understand, the name metadata can use underscores as per the pattern linked above.

An example of a project using underscores: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/pyproject.toml (this one is using poetry, instead of setuptools which is what we have right now).

Sentry is using name with hyphens https://github.com/getsentry/sentry-python/blob/master/setup.py#L23

That being said, pep8 https://peps.python.org/pep-0008/#package-and-module-names mentions

>Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

I'm learning as we speak :) I think it makes sense to change the names in the metadata to use hyphens instead of underscores.

medikoo:
@selcukcihan great feedback! I also like shifting towards hyphens, as we use that in Node.js, so package names will be more alligned

medikoo (approved):
Looks great 👍

medikoo (changes requested):
@selcukcihan sorry I've just realized this change is not complete.

We need to rename all references in repository, there are also some in CI and .gitignore

selcukcihan:
> @selcukcihan sorry I've just realized this change is not complete.
> 
> We need to rename all references in repository, there are also some in CI and .gitignore

I've renamed references in the CI, but those were related to the release tag name and name of the CI flow.

I think we should not rename our packages, I haven't found a single example where they use dashes instead of underscores.

For example: https://github.com/aws/aws-xray-sdk-python recommends installing package with `pip install aws-xray-sdk` but then you import it as in `from aws_xray_sdk.core import xray_recorder`. All the major packages I've checked like sentry, powertools etc. all use underscores in the package names instead of dashes. Now this is not to be confused with how we name the project and how it's published on PyPI.

medikoo (approved):
Thanks @selcukcihan !

I was not aware that we speak about projects and packages, and that change should just apply to projects

PR #467, opened 2023-02-22
(SC-372) SDK: Document and publish v0.1.0

selcukcihan:
### Description
Related issue https://linear.app/serverless/issue/SC-372/sdk-document-and-publish-v010
* Add manually created changelog for the initial 0.1.0 version
* Bump serverless-sdk version to 0.1.0
* Add docs for python serverless-sdk package.

Note that the base sdk is missing some features, for instance `.instrumentation` is just a placeholder. As such I haven't completed those parts of the API in documentation.

### Testing done
N/A

PR #469, opened 2023-02-22
Fix python integrate flow

selcukcihan:
### Description
Related issue https://github.com/serverless/console/pull/467#issuecomment-1439810314

The integrate flow runs in the context of the main branch, it does not need a step to checkout another branch which was causing no diffs and the flow ending prematurely.

### Testing done
I've pushed the changes on my forked repo and verified the path changes are picked up

Screenshot 2023-02-22 at 14 14 58

### DIFF
--- a/.github/workflows/python-integrate.yaml
+++ b/.github/workflows/python-integrate.yaml
@@ -20,15 +20,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
-        with:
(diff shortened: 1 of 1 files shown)


medikoo (approved):
@selcukcihan I don't think it's relevant to our issue, yet indeed this setup is obsolete in Python. In Node.js it's needed for `git-list-updated` method, which needs access to current state of `main`.

Therefore taking it in, as a rightful cleanup

PR #473, opened 2023-02-23
Isolate package installations

selcukcihan:
### Description
Related issue https://github.com/serverless/console/pull/464/files#r1115357043

The "python-validate" CI flow currently installs all SDK related packages in the same python environment. This prevents unit tests to depend on pypi packages and instead the tests will be referring to locally installed packages.

This will change the CI flow such that each package gets their own python environment isolated from the other packages.

### Testing done
To be tested on the CI with this PR.


selcukcihan:
Apparently I had created this PR for the exact case in https://github.com/serverless/console/pull/516 anyway, it's covered in the other PR so closing this one.

PR #474, opened 2023-02-24
AWS Lambda SDK: Ensure not to capture logs as issued by the AWS runtime

medikoo:
Reported internally by @eahefnawy 

On function crash AWS runtime logs the error to stdout via `console.error` call, which we instrument. In result we've observed that error is also captured as _user caught_ error, which is not true.

This patch ensures we do not capture `console.*` calls as issued by the runtime.

Additionally improved tests to confirm on whole set of captured events


Danwakeem (commented):
This looks good but would it make sense to still capture this error but as an uncaught error? I would image this would still contain some useful information 🤔

medikoo:
> This looks good but would it make sense to still capture this error but as an uncaught error? I would image this would still contain some useful information

@Danwakeem it is captured as uncaught. Issue is that it was reported as both _uncaught_ and _caught_

Danwakeem (approved):
Ahh got it 👍 Thanks @medikoo

PR #477, opened 2023-02-24
AWS Lambda SDK: Test against AWS SDK that doesn't issue deprecation error

medikoo:
Addresses CI fail: https://github.com/serverless/console/actions/runs/4263772005

Recent version of AWS SDK v2 started to issue deprecation warning which is logged via `console.error` (more information at https://github.com/serverless/serverless/issues/11753), and that started to issue capture warning events which broke our tests.

This patch ensures we rely on latest AWS SDK v2 version that doesn't print that warning


Danwakeem (approved):
✅

PR #479, opened 2023-02-28
AWS Lambda SDK: Document and publish v0.1.0

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-373/aws-lambda-sdk-document-and-publish-v010
* Add manually created changelog for the initial 0.1.0 version
* Bump serverless-aws-lambda-sdk version to 0.1.0
* Add docs for python serverless-aws-lambda-sdk package

### Testing done
N/A


medikoo (changes requested):
@selcukcihan this should be released after we have first traces configured (so `SC-356` done), as without that it won't be any way functional for us yet

PR #480, opened 2023-02-28
SDK: Ensure full stack traces in debug mode

medikoo:
When inspecting the issue related to the delayed AWS SDK requests, I've approached truncated stack traces in gathered logs, which weren't informative enough. This patch will ensure full stack traces when next case is recorded


Danwakeem (dismissed):
👍

PR #481, opened 2023-03-01
Trace span

selcukcihan:
### Description
Related issue https://linear.app/serverless/issue/SC-356#comment-3690173f
* Enhance/fix trace span parent/child relation
* Add unit tests that validate these relations as well as other functionality as tested in the nodejs sdk.
* Protobuf object and json serialization

### Testing done
Unit tested with
```bash
cd python/packages/sdk
python3 -m venv .venv
source .venv/bin/activate
python3 -m pip install -e .
python3 -m pytest
```

### DIFF
--- a/python/packages/sdk/serverless_sdk/sdk/base.py
+++ b/python/packages/sdk/serverless_sdk/sdk/base.py
@@ -37,7 +37,7 @@ class ServerlessSdk:
     def _initialize(self, org_id: Optional[str] = None):
         self.org_id = environ.get(SLS_ORG_ID, default=org_id)
 
-    def create_trace_span(
+    def _create_trace_span(
         self,
         name: str,
         input: str,
--- a/python/packages/sdk/serverless_sdk/sdk/base.py
+++ b/python/packages/sdk/serverless_sdk/sdk/base.py
@@ -2,10 +2,11 @@ from __future__ import annotations
 
 from os import environ
 from typing import List, Optional
-
 from typing_extensions import Final
+from types import SimpleNamespace
 
 from ..base import Nanoseconds, SLS_ORG_ID, __version__, __name__
+from ..span import trace
 from ..span.trace import TraceSpan
 from ..span.tags import Tags
 
@@ -15,14 +16,23 @@ __all__: Final[List[str]] = [
 ]
 
 
+class TraceSpans(SimpleNamespace):
+    @property
+    def root(self):
+        return trace.root_span
+
+
 class ServerlessSdk:
     name: Final[str] = __name__
     version: Final[str] = __version__
 
-    trace_spans: Final = ...
+    trace_spans: TraceSpans
     instrumentation: Final = ...
 
     org_id: Optional[str] = None
+
+    def __init__(self):
+        self.trace_spans = TraceSpans()
 
     def _initialize(self, org_id: Optional[str] = None):
         self.org_id = environ.get(SLS_ORG_ID, default=org_id)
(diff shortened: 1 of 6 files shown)


python/packages/sdk/serverless_sdk/sdk/base.py
```diff
@@ -15,19 +16,28 @@
...
+        return trace.root_span
+
+
 class ServerlessSdk:
     name: Final[str] = __name__
     version: Final[str] = __version__
 
-    trace_spans: Final = ...
+    trace_spans: TraceSpans
     instrumentation: Final = ...
```
medikoo, line 30: Just wondering, what this `Final = ...` notation means? (looks as syntax error)

python/packages/sdk/serverless_sdk/tests/test_trace_span.py
```diff
@@ -1,152 +1,196 @@
...
+    ), "should expose `trace_id` of a root span"
+    assert type(child_span.id) == str, "should automatically generate `id`"
+    assert child_span.id != root_span.id, "should have a unique `id`"
+    assert (
+        type(child_span.start_time) == int
+    ), "should automatically generate `start_time`"
+    assert (
+        child_span.start_time != root_span.start_time
+    ), "should have a different `start_time`"
+    assert child_span.name == "child", "should expose `name`"
```
medikoo, line 42: Do we have a test that confirms that `parent` span is attributed correctly (and that created spans ends in its child spans) ?

medikoo (commented):
Thank you @selcukcihan 👍

python/packages/sdk/serverless_sdk/sdk/base.py
selcukcihan: That's like a placeholder, we'll change it when we actually implement `serverlessSdk.instrumentation`. It's not a syntax error in my environment, maybe you have an old Python version?

python/packages/sdk/serverless_sdk/tests/test_trace_span.py
selcukcihan: Can you elaborate a bit more please 🙏  What do you mean by `parent` span being attributed correctly? And also this bit: 
> created spans ends in its child spans

python/packages/sdk/serverless_sdk/sdk/base.py
medikoo: >  It's not a syntax error in my environment, maybe you have an old Python version?

I just mean that it looks as _syntax error_ and not that it is one.

If I currently will try to access `serverless_sdk.instrumentation` what will be returned? Some empty value? I wonder to what value `...` resolves

python/packages/sdk/serverless_sdk/tests/test_trace_span.py
medikoo: When `child_span` is created, we assign a _parent_ to it: https://github.com/serverless/console/blob/052af60e4979974e30386f5808964a210add38d2/python/packages/sdk/serverless_sdk/span/trace.py#L128
So do we confirm in tests that expected _span_ is assigned as `child_span.parent_span` ?

Also do we confirm that `child_span` ends in `parent_span.sub_spans` collection?

medikoo: Those relations are crucial for tracing to work properly, therefore need to be covered with tests

python/packages/sdk/serverless_sdk/sdk/base.py
selcukcihan: Got it, so it returns an empty object like a stub:
```
>>> from serverless_sdk import serverlessSdk
>>> serverlessSdk.instrumentation
Ellipsis
```

python/packages/sdk/serverless_sdk/tests/test_trace_span.py
selcukcihan: Ah I see, yeah good point, I've added these assertions 👍

medikoo (approved):
Thanks @selcukcihan 👍

PR #483, opened 2023-03-02
Initial integration tests for Python AWS Lambda SDK

medikoo:
/cc @selcukcihan 

Test basic _success_ and _error_ scenario in all AWS supported Python runtimes.

To have this test run successfully following needs to be ensured:
- Python layer build script (located at `python/packages/aws-lambda-sdk/scripts/build-layer-archive.sh`) needs to be updated to:
  - Generate python lambda layer into `python/packages/aws-lambda-sdk/dist/<filename>` when run with `<filename>` arguments
  - Working directory in which script is run should not affect the path of destination file, may just influence resolution of `<filename>` of which path input may  be relative
- Fixtures for `success` and `error`  cases need to be sunred in `python/packages/aws-lambda-sdk/test/fixtures/lambdas`

Having above covered, it should be possible to have a successful test run (from context of `node` folder) by ensuring `SLS_ORG_ID` (can be anything, e.g. `test`), `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`and `AWS_REGION` environment variables and running:

```shell
npx mocha test/python/aws-lambda-sdk/integration.test.js
```

--- 

Additionally:
- Generalize few test utils
- Improve process start resolution in CW logs processing, which is easier since introduction of `INIT_START Runtime Version:..` log
- Rename test basenames, so all resources as created when  testing _node_, _dev mode_ and _python_ have different basename (rename _node_ basename from `test-sdk` to `test-node-sdk` and _dev mode_ basename from `test-sdk` to `test-dev-mode` - cc @Danwakeem )

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -70,6 +70,7 @@
     "standard-version": "^9.5.0",
     "stream-promise": "^3.2.0",
     "timers-ext": "^0.1.7",
+    "toml": "^3.0.0",
     "ts-proto": "^1.140.0",
     "tslib": "^2.5.0",
     "type": "^2.7.2",
--- a/node/packages/aws-lambda-sdk/test/integration/index.test.js
+++ b/node/packages/aws-lambda-sdk/test/integration/index.test.js
@@ -1855,16 +1855,28 @@ describe('integration', function () {
             const [, initializationSpan, invocationSpan] = spans;
             expect(String(initializationSpan.parentSpanId)).to.equal(String(lambdaSpan.id));
             expect(String(invocationSpan.parentSpanId)).to.equal(String(lambdaSpan.id));
+            expect(lambdaSpan.startTimeUnixNano).to.equal(initializationSpan.startTimeUnixNano);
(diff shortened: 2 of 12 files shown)


selcukcihan:
@medikoo can you clarify the following requirement for the build script please, seems like a typo but I couldn't infer the intent? When run with which arguments?

> Generate python lambda layer into ``python/packages/aws-lambda-sdk/dist/when run with`arguments

medikoo:
> @medikoo can you clarify the following requirement for the build script please, seems like a typo but I couldn't infer the intent? When run with which arguments?

@selcukcihan sorry, I've fixed the formatting in description. Ideally if all that script takes is filename path to where the zip should be generated (not a _directory_ and _filename_, just a _filename_ which can be full absolute path or relative path to be resolved against current working directory)

medikoo:
@Danwakeem @Mmarzex I'm going to take this in, as we need this for Python SDK work. Still if you have any comments. please write, we can always update with following PR

PR #484, opened 2023-03-02
Fix trace span logic for Python base SDK

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-356/aws-lambda-sdk-create-lambda-process-trace-spans-with-related-tags
* Fix trace timer by using the high resolution nanosecond timer
* Fix trace span close logic such that the context is updated to the first open ancestor, in case the closing span is the current span.
* Add base sdk to vscode python path so that intellisense works.

### Testing done
I've tested it on the cloud and was able to create traces together with the aws-serverless-sdk (which is wip, PR dedicated to it will be provided separately).

### DIFF
--- a/python/packages/sdk/serverless_sdk/sdk/base.py
+++ b/python/packages/sdk/serverless_sdk/sdk/base.py
@@ -40,8 +40,8 @@ class ServerlessSdk:
     def _create_trace_span(
         self,
         name: str,
-        input: str,
-        output: str,
+        input: Optional[str] = None,
+        output: Optional[str] = None,
         start_time: Optional[Nanoseconds] = None,
         tags: Optional[Tags] = None,
     ) -> TraceSpan:
--- a/python/packages/sdk/serverless_sdk/span/trace.py
+++ b/python/packages/sdk/serverless_sdk/span/trace.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 from collections.abc import Iterable
 import logging
-from time import time_ns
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -1,5 +1,4 @@
 {
-    "cSpell.words": [
```
medikoo: Hmm.. actually, this file should not be a part of the repository, it landed here by mistake (when migrating documentation from other repo)

VS Code settings is mostly about personal settings (and it also implies VS Code as an editor choice, which also should be up to developer)

I've removed it from the repo. You can ensure it doesn't end in your git diff by updating gitignore global rules (see https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

medikoo (changes requested):
Great find @selcukcihan all is good, let's just remove this .vscode update

.vscode/settings.json
selcukcihan: Yeah makes sense 👍

medikoo (dismissed):
Looks great 👍

medikoo (commented):
@selcukcihan Again I have "This branch cannot be rebased due to conflicts", can you rebase the branch?

selcukcihan:
> @selcukcihan Again I have "This branch cannot be rebased due to conflicts", can you rebase the branch?

Can you try again please? I've squashed the commits.

medikoo (approved):

PR #485, opened 2023-03-02
Create layer archive in destination

selcukcihan:
### Description
* Related PR https://github.com/serverless/console/pull/483
* Remove the shopt command which we don't need
* Install `serverless-aws-lambda-sdk` from pypi registry
* Accept both absolute and relative path for the output artifact file.

### Testing done
* Tested locally using pip install from local repository (we don't have the package on pypi yet as it's not published)
* Tested both with relative and absolute path

@medikoo Notes:
* I've noticed you mentioned the script to create output within `python/packages/aws-lambda-sdk/dist` folder, it depends on where the script is located because that's a relative path. So I haven't implemented it like that, the consumer can decide whether to use a relative path depending on where the script is being executed from or an absolute path. I think you've used the absolute path version in the integration test script so this change is compatible with that PR.
* I've also noticed that the integration test script does not build/install the SDK, whereas the previous version of this script assumed that the SDK was installed and it required an argument to point to the installation directory. I've now removed that logic and it will install from pypi. Does that make sense? It assumes the package is already published. I feel like we should not publish a version if integration tests are not passing, we need to run integration tests before the package is published.


.github/workflows/python-publish-aws-lambda-sdk.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         run: |
           cd python/packages
 
-          ./aws-lambda-sdk/scripts/build-layer-archive.sh build extension.internal.zip
+          ./aws-lambda-sdk/scripts/build-layer-archive.sh extension.internal.zip
```
medikoo: To maintain same convention it'll be probably best if same as on Node.js side it's built into `aws-lambda-sdk/dist` folder, and that we also ensure that this folder is in `.gitignore` (as Node.js one is: https://github.com/serverless/console/blob/687ed085e40b003c41aa1fde5012b69e0b887cb4/node/packages/aws-lambda-sdk/.gitignore#LL1C6-L1C6)

medikoo (commented):
Looks good, I have just one extra suggestion

medikoo (approved):
👍

python/packages/aws-lambda-sdk/scripts/build-layer-archive.sh
```diff
@@ -1,23 +1,31 @@
...
+case $1 in
+  /*) OUTPUT=$1 ;;
+  *) OUTPUT=$CURRENT_DIR/$1 ;;
+esac
+
+SITE_PACKAGES_DIR=python/lib/python3.9/site-packages
+mkdir -p $DIST/{$SITE_PACKAGES_DIR,sls-sdk-python}
+
+INSTALL_DIR=$DIST/$SITE_PACKAGES_DIR
+python3 -m pip install serverless-aws-lambda-sdk --target=$INSTALL_DIR
```
medikoo, line 16: @selcukcihan I just started thinking about that, and I think that's incorrect.

This script should build the layer which represents state in repostotyr. It's important as it's used in integration tests, and in integration tests confirm that current state in `main` is working, (and not latest release in PyPI)

selcukcihan: Yeah I think that makes sense, that integration tests use `main`. Which means we somehow need to build and install the main version of the package before creating the layer. Should we do this within the integration test script or can we push it out of the script like for instance previous step in the CI flow? What do you think?

medikoo: > Which means we somehow need to build and install the main version of the package before creating the layer. 

Integration tests are run in the assumption that all dependencies are installed (integration tests on its own should not pursue deps installation step). It's just build layer step that's integrated in integration tests flow (and that's covered in https://github.com/serverless/console/pull/483).

And in CI we already have ensured that dependencies from PyPI are installed, so if this build script will ensure that layer is build of code in repository, all should work out of a box

selcukcihan: https://github.com/serverless/console/pull/488 to build the lambda sdk package from code in the repo instead of pypi.

PR #488, opened 2023-03-03
Build Lambda SDK from local when packaging layer

selcukcihan:
### Description
* Install `aws-lambda-sdk` from local instead of pypi when build script executes
* Determine directory of local source for the package relatively to the script. This implies that we should not copy the script to another directory. It can be executed from any directory though.

### Testing done
I've tested the script as used in https://github.com/serverless/console/pull/483

`/Users/selcukcihan/serverless/console/python/packages/aws-lambda-sdk/scripts/build-layer-archive.sh /Users/selcukcihan/serverless/console/python/packages/aws-lambda-sdk/dist/extension.internal.zip`

Which resulted in `/Users/selcukcihan/serverless/console/python/packages/aws-lambda-sdk/dist/extension.internal.zip`

The build used local version of aws-lambda-sdk for sure, because it's not yet published on pypi.


medikoo (approved):
👍

PR #486, opened 2023-03-02
Add aws lambda trace spans

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-356/aws-lambda-sdk-create-lambda-process-trace-spans-with-related-tags
* Add AWS Lambda Trace spans (root span, initialization and invocation)
* Print the trace at the end

### Testing done
Tested with unit tests to ensure
  * lambda spans are created
  * tags are attached

PR #489, opened 2023-03-06
Ensure spans are ordered by creation time

selcukcihan:
Related issue  https://github.com/serverless/console/pull/486#pullrequestreview-1324005295

### Description
Python "set" is not guaranteed to be ordered, so spans are now stored in a list that ensures they are sorted by the time of creation.

### Testing done
Enhanced the test case to assert on list equality, which validates order of the list.


medikoo (approved):
Looks great 👍

PR #490, opened 2023-03-06
Fix root span handling

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-220#comment-908752da

### Description
Integration testing revealed some issues with traces, I've isolated them with a unit test and provided a fix. The fix is regarding how we handle the root span.

### Testing done
Unit tested


medikoo (approved):
:+1

PR #496, opened 2023-03-07
Fix publish lambda sdk flow

selcukcihan:
Fixes the CI workflow issue encountered in https://github.com/serverless/console/actions/runs/4354871821

### Description
Fixes the artifact directories to make sure correct paths are being used.
The flow can be summarized as:
1. Publish package to pypi, this uses python's build module to bundle the SDK code.
2. Install the SDK in a specific directory, which is used by the build-layer-archive script to build the lambda layer zip file
3. Publish the layer on AWS Lambda

### Testing done
I've reproduced the error by following CI steps locally and also verified that these steps run successfully after the fix, again verified locally.

### DIFF
--- a/.github/workflows/python-publish-aws-lambda-sdk.yml
+++ b/.github/workflows/python-publish-aws-lambda-sdk.yml
@@ -25,31 +25,31 @@ jobs:
           cache-dependency-path: |
             **/pyproject.toml
 
-      - name: Install main project dependencies
-        run: |
-          cd python/packages
-          python3 -m venv .venv
-          source .venv/bin/activate
-          python3 -m pip install ./aws-lambda-sdk --target=dist
-
       - name: Publish New Version on PyPI
         env:
           TWINE_USERNAME: "__token__"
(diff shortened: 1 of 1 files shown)


medikoo (approved):
:+1 

After merging I will reset the release tag to this new commit, and we'll see how publication we'll go

medikoo:
> After merging I will reset the release tag to this new commit, and we'll see how publication we'll go

@selcukcihan  Actually I can't do that, as PyPI publication was successful, so republish of v0.1.0 will now crash (as expected) on PyPI publish step.

In light of that let's prepare v0.1.1

selcukcihan:
> > After merging I will reset the release tag to this new commit, and we'll see how publication we'll go
> 
> @selcukcihan Actually I can't do that, as PyPI publication was successful, so republish of v0.1.0 will now crash (as expected) on PyPI publish step.
> 
> In light of that let's prepare v0.1.1

https://github.com/serverless/console/pull/497 🤞

PR #498, opened 2023-03-07
Create layers registry if it doesn't exist yet

medikoo:
When publishing to Python registry, CI failed as expected registry file to exist already.

This patch improves script so it doesn't assume already existing registry, and creates one if it isn't there yet


Danwakeem (approved):
👍 Good call. I think when I created the dev mode registry I had manually created the file first.

PR #500, opened 2023-03-07
feat: Support user defined spans in Go SDK

Mmarzex:
To support tracing our clickhouse benchmarks, we need a way to create spans for each segment of the performance test. This PR introduces the ability to create spans through the Go SDK to meet that goal.

### DIFF
--- a/go/packages/slslambda/slslambda.go
+++ b/go/packages/slslambda/slslambda.go
@@ -2,9 +2,10 @@ package slslambda
 
 import (
 	"context"
-	"github.com/aws/aws-lambda-go/lambda"
 	"runtime/debug"
 	"time"
+
+	"github.com/aws/aws-lambda-go/lambda"
 )
 
 var initializationStart = time.Now()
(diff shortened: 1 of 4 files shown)


czubocha (commented):
IMO adding a `UserSpan` type to the public interface doesn't seem very convenient for customers because they will need to pass it everywhere they want to use custom spans. I think `Context` is the standard way in Go to pass metadata like tracing so we can just use it to store the current span there and functions like `CaptureError`, `AddTags` will just operate on the current span. I prepared [implementation of it here](https://github.com/serverless/console/compare/go-sdk-custom-spans) and extended the example so you can see how it works. Also introducing the `span` interface and special types like `initializationSpan` satisfying this interface make it easier to extend the SDK in the future with AWS services spans or any other types of spans. Let me know what you think.

image

Mmarzex:
@czubocha I'm fine with the implementation you suggested, that is cleaner than what I wrote. Can you go ahead and create a PR for that? I just had a few comments on it that I think would be easier to follow in a PR than in this comment.

PR #505, opened 2023-03-08
Code compatibility with Python 3.8+

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-560/ensure-support-for-python-37-and-38

### Description
1. Make sure the code is compatible with Python v3.7 by running the unit tests in v3.7 environment. Turns out the only problem is with a unit test mock setup and it's fixed now.
2. Paths should not use specific python version, instead the extension is installed in `/opt/python`
3. CI flows are updated to use v3.7
4. v3.8 is added to the Integration tests.

### Testing done
1. Unit tested
2. Integration tested:
```
(.venv) ➜  node git:(python37-compatibility) ✗ npx mocha test/python/aws-lambda-sdk/integration.test.js             
2023-03-08T12:37:06.701Z ℹ test test uid: cdab


  Python: integration
2023-03-08T12:37:06.812Z ℹ test Creating core resources test-python-sdk-cdab
2023-03-08T12:37:36.492Z ℹ test Process function success-v3-8
2023-03-08T12:37:36.492Z ℹ test Process function success-v3-9
2023-03-08T12:37:36.493Z ℹ test Process function error-v3-8
2023-03-08T12:37:36.493Z ℹ test Process function error-v3-9
2023-03-08T12:37:36.751Z ℹ test Function success-v3-8 already exists, deleting and re-creating
    ✔ success-v3-8 (22946ms)
    ✔ success-v3-9 (2422ms)
    ✔ error-v3-8
    ✔ error-v3-9
2023-03-08T12:38:01.871Z ℹ test Cleanup test-python-sdk-cdab
2023-03-08T12:38:02.200Z ℹ test Detached IAM policy arn:aws:iam::692593327170:policy/test-python-sdk-cdab from test-python-sdk-cdab
2023-03-08T12:38:02.228Z ℹ test Deleted 18 version of the layer test-python-sdk-cdab-internal
2023-03-08T12:38:02.383Z ℹ test Deleted IAM role test-python-sdk-cdab
2023-03-08T12:38:02.442Z ℹ test Deleted IAM policy arn:aws:iam::692593327170:policy/test-python-sdk-cdab


  4 passing (56s)
```
3. Manual testing on my AWS account just to make sure.

### Out of Scope
Unfortunately, this setup does not work with AWS Lambda Python3.7 runtime. It seems this particular runtime does not care about `AWS_LAMBDA_EXEC_WRAPPER` and thus it won't ever run the exec-wrapper script. Docs around AWS_LAMBDA_EXEC_WRAPPER always mention Python3.8 and this is in agreement with my findings.

For the case of codeguru, they instruct that the customer sets their handler to a function bundled in the layer:
> Change the Lambda handler function to codeguru_profiler_agent.aws_lambda.lambda_handler.call_handler.

So, we still need to figure out how to auto instrument Python3.7 runtime.

**Docs:**
1. https://d1.awsstatic.com/events/reinvent/2020/Observability_logging_and_more_with_AWS_Lambda_extensions_SVS317.pdf
2. https://docs.aws.amazon.com/codeguru/latest/profiler-ug/python-lambda-layers.html

### DIFF
--- a/.github/workflows/python-integrate.yaml
+++ b/.github/workflows/python-integrate.yaml
@@ -10,8 +10,8 @@ on:
       - proto/**
 
 jobs:
-  linuxPython39:
-    name: "[Linux] Python 3.9: Unit Tests"
+  linuxPython37:
+    name: "[Linux] Python 3.7: Unit Tests"
     runs-on: ubuntu-latest
     outputs:
       pathChangedAwsLambdaSdk: ${{ steps.pathChanges.outputs.awsLambdaSdk }}
@@ -36,10 +36,10 @@ jobs:
             proto:
               - 'proto/**'
 
-      - name: Install Python 3.9
+      - name: Install Python 3.7
         uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
+          python-version: '3.7'
 
           # ensure project dependencies are cached
           # When using only `pyproject.toml` for dependencies, see:
@@ -110,8 +110,8 @@ jobs:
   integratePythonSdk:
     name: Integrate Python SDK
     runs-on: ubuntu-latest
-    needs: [ linuxPython39 ]
-    if: needs.linuxPython39.outputs.pathChangedSdk == 'true'
+    needs: [ linuxPython37 ]
+    if: needs.linuxPython37.outputs.pathChangedSdk == 'true'
     timeout-minutes: 5 # Default is 360
     steps:
       - name: Checkout repository
@@ -136,8 +136,8 @@ jobs:
   integrateAwsLambdaSdk:
     name: Integrate Python AWS Lambda SDK
     runs-on: ubuntu-latest
-    needs: [ linuxPython39 ]
-    if: needs.linuxPython39.outputs.pathChangedAwsLambdaSdk == 'true'
+    needs: [ linuxPython37 ]
+    if: needs.linuxPython37.outputs.pathChangedAwsLambdaSdk == 'true'
     timeout-minutes: 5 # Default is 360
     env:
       AWS_REGION: us-east-1
@@ -155,10 +155,10 @@ jobs:
           # Hence we're passing 'serverless-ci' user authentication token
           token: ${{ secrets.USER_GITHUB_TOKEN }}
 
-      - name: Install Python 3.9
+      - name: Install Python 3.7
         uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
+          python-version: '3.7'
           cache: 'pip'
           cache-dependency-path: |
             **/pyproject.toml
@@ -196,8 +196,8 @@ jobs:
   integratePythonSdkSchema:
     name: Integrate Python SDK Schema
     runs-on: ubuntu-latest
-    needs: [ linuxPython39 ]
-    if: needs.linuxPython39.outputs.pathChangedSdkSchema == 'true'
+    needs: [ linuxPython37 ]
+    if: needs.linuxPython37.outputs.pathChangedSdkSchema == 'true'
     timeout-minutes: 5 # Default is 360
     steps:
       - name: Checkout repository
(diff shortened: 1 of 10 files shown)


.github/workflows/python-integrate.yaml
```diff
@@ -36,10 +36,10 @@ jobs:
             proto:
               - 'proto/**'
 
-      - name: Install Python 3.9
+      - name: Install Python 3.7
         uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
+          python-version: '3.7'
```
medikoo: If I see correctly unit tests now will be run only in context of 3.7. Can we make them run in all supported versions (we can rely on matrix CI setting if that's helpful)

.github/workflows/python-validate.yml
```diff
@@ -40,10 +40,10 @@ jobs:
             packages:
               - 'python/packages/**'
 
-      - name: Install Python 3.9
+      - name: Install Python 3.7
         uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
+          python-version: '3.7'
```
medikoo: Same here

node/test/python/aws-lambda-sdk/integration.test.js
```diff
@@ -36,7 +36,8 @@ describe('Python: integration', function () {
       'success',
       {
         variants: new Map([
-          // TODO: Cover v3.6 and v3.7 once support for those is fixed
+          // TODO: Cover v3.7 once support for it is fixed
+          ['v3-8', { configuration: { Runtime: 'python3.8' } }],
```
medikoo, line 40: So support for 3.7 is still pending? (title of the PR says otherwise)

medikoo (changes requested):

node/test/python/aws-lambda-sdk/integration.test.js
selcukcihan: Totally understand the confusion. Yes, 3.7 is still pending in terms of the internal extension. Code is compatible though, so if customer wanted to import and manually set up instrumentation they can do it.

.github/workflows/python-integrate.yaml
selcukcihan: That's correct, I'll work on updating it to run on [3.7, 3.8, 3.9]. I'll need to research what matrix CI is.

node/test/python/aws-lambda-sdk/integration.test.js
medikoo: @selcukcihan so can you clarify specifically, why it's not assumed as ready for v3.7. You say that it'll work for a customer, so why we do not assume it as ready?

selcukcihan: Because Python v3.7 runtime does not support setting `AWS_LAMBDA_EXEC_WRAPPER`. So we need to find another mechanism to wrap the customer handler. More details are in the "out of scope" section in the PR description.

.github/workflows/python-integrate.yaml
selcukcihan: Turns out it's pretty easy to set up and very much relevant to our needs 👍

node/test/python/aws-lambda-sdk/integration.test.js
medikoo: Ok, thanks for pointing that out. I think that means that Python 3.7 can only be instrumented manually. We will discuss it internally, for now let's stick to v3.8 (unless we find a way to instrument it automatically)

node/test/python/aws-lambda-sdk/integration.test.js
```diff
@@ -37,7 +37,8 @@ describe('Python: integration', function () {
       'success',
       {
         variants: new Map([
-          // TODO: Cover v3.6 and v3.7 once support for those is fixed
+          // TODO: Cover v3.7 once support for it is fixed
```
medikoo: Let's remove for now those TODO comments, as currently we don't see a reliable way to support Python v3.7

medikoo (approved):
👍

PR #507, opened 2023-03-08
feat(Go SDK): Support user defined spans in Go SDK

czubocha:
PR created based on https://github.com/serverless/console/pull/500
> To support tracing our clickhouse benchmarks, we need a way to create spans for each segment of the performance test. This PR introduces the ability to create spans through the Go SDK to meet that goal.

image

### DIFF
--- a/go/packages/slslambda/basicspan.go
+++ b/go/packages/slslambda/basicspan.go
@@ -1,0 +1,120 @@
+package slslambda
+
+import (
+	"context"
+	"encoding/json"
+	"github.com/aws/aws-sdk-go/aws"
+	"go.buf.build/protocolbuffers/go/serverless/sdk-schema/serverless/instrumentation/tags/v1"
+	"go.buf.build/protocolbuffers/go/serverless/sdk-schema/serverless/instrumentation/v1"
+	"sync"
+	"time"
+)
+
+type basicSpan struct {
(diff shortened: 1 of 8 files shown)


go/packages/slslambda/basicspan.go
```diff
@@ -0,0 +1,120 @@
...
+		Id:                spanID,
+		TraceId:           traceID,
+		ParentSpanId:      parentSpanID,
+		Name:              s.name,
+		StartTimeUnixNano: uint64(s.start.UnixNano()),
+		EndTimeUnixNano:   uint64(s.end.UnixNano()),
+		CustomTags:        convertCustomTags(s.customTags),
+		Tags: &tagsv1.Tags{
+			OrgId: (*string)(&tags.OrganizationID),
+			Aws: &tagsv1.AwsTags{
```
Mmarzex: These tags should only be attached to the root span. We shouldn't attach them to every span. I think at the moment for custom spans the only tag that will be set is `orgId`

Mmarzex (changes requested):

go/packages/slslambda/basicspan.go
czubocha: done

Mmarzex (approved):

PR #509, opened 2023-03-09
Catch all customer handler errors

selcukcihan:
Somewhat related issue https://linear.app/serverless/issue/SC-562/recognize-handled-and-unhandled-error-resolution-if-applicable

### Description
While reviewing the linked issue, I've realized that we were not catching all exceptions from customer's handler.

### Testing done
Unit tested.


medikoo (commented):
@selcukcihan Great find! Can we cover newly added fixtures with integration tests?

selcukcihan:
> @selcukcihan Great find! Can we cover newly added fixtures with integration tests?

I was trying to do that, but turns out I've found an interesting case with the `exit(1)` handler. This code piece, with or without the SDK, somehow emits `INIT_START` log twice:

```python
import json

print("IN THE HANDLER MODULE")
def lambda_handler(event, context):
    print("in the handler function")
    exit(1)
    return {
        'statusCode': 200,
        'body': json.dumps('Hello from Lambda!')
    }
```

### WITHOUT EXTENSION
Screenshot 2023-03-09 at 22 18 22

### WITH EXTENSION
Screenshot 2023-03-09 at 22 19 24

Which is causing the integration tests to fail with
```
  2 failing

  1) Python: integration
       exit_with_error-v3-8:
     Error: Unexpected count of processes (2) for: exit_with_error-v3-8
      at retrieveReports (test/lib/get-process-function.js:386:13)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async /Users/selcukcihan/serverless/console/node/test/lib/get-process-function.js:434:21
      at async Context.<anonymous> (test/python/aws-lambda-sdk/integration.test.js:110:26)

  2) Python: integration
       exit_with_error-v3-9:
     Error: Unexpected count of processes (2) for: exit_with_error-v3-9
      at retrieveReports (test/lib/get-process-function.js:386:13)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async /Users/selcukcihan/serverless/console/node/test/lib/get-process-function.js:434:21
```

I think we need to handle this special case in a different way. @medikoo you may have a better idea on how to modify the `INIT_START` check logic that was introduced with https://github.com/serverless/console/commit/d175fcfd3706f3f378e0f4a39bb37658221537c4

Or we can ignore this edge case, it falls under the class of exceptions that are not deriving from `Exception`

> [BaseException](https://docs.python.org/3/library/exceptions.html#BaseException) is the common base class of all exceptions. One of its subclasses, [Exception](https://docs.python.org/3/library/exceptions.html#Exception), is the base class of all the non-fatal exceptions. Exceptions which are not subclasses of [Exception](https://docs.python.org/3/library/exceptions.html#Exception) are not typically handled, because they are used to indicate that the program should terminate. They include [SystemExit](https://docs.python.org/3/library/exceptions.html#SystemExit) which is raised by [sys.exit()](https://docs.python.org/3/library/sys.html#sys.exit) and [KeyboardInterrupt](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt) which is raised when a user wishes to interrupt the program.

I think what happens is, it breaks the Python Runtime and causes the process to exit and then it inits the process again where we see that INIT_START second time, but the handler is not invoked again.

medikoo:
@selcukcihan What happens here, is that AWS treats `exit(1)` as unclean crash (something like uncaught exception in Node.js), and when this happens, AWS will always restart the process for a new invocation, so again we deal with initialization phase.

That's totally expected, and that actually showcases the `error:unhandled` outcome case in Python, which we initially assumed  (in SC-562) cannot take place.

Extension if possible should detect this scenario happening and set the outcome to `error:unhandled` for the report. Is this possible?

Tests are already prepared to handle such case, assuming the proper outcome is set. See: https://github.com/serverless/console/blob/89d3160ea9b013217ab236d4d00edafbf1af2e9f/node/test/python/aws-lambda-sdk/integration.test.js#L124 this line ensures that in the case of `error:unhandled` we treat every invocation as first one which is preceded with initialization

selcukcihan:
> @selcukcihan What happens here, is that AWS treats `exit(1)` as unclean crash (something like uncaught exception in Node.js), and when this happens, AWS will always restart the process for a new invocation, so again we deal with initialization phase.
> 
> That's totally expected, and that actually showcases the `error:unhandled` outcome case in Python, which we initially assumed (in SC-562) that cannot take place.
> 
> Extension if possible should detect this scenario happening and set the outcome to `error:unhandled` for the report. Is this possible?
> 
> Tests are already prepared to handle such case, assuming the proper outcome is set. See:
> 
> https://github.com/serverless/console/blob/89d3160ea9b013217ab236d4d00edafbf1af2e9f/node/test/python/aws-lambda-sdk/integration.test.js#L124
> 
> this line ensures that in the case of `error:unhandled` we treat every invocation as first one which is preceded with initialization

Thanks for the clear explanation 🙏 

* I've made the change to distinguish between handled/unhandled exceptions.
* Added unit tests to cover both cases and refactored unit tests to use the fixtures
* Added integration test to cover unhandled exception

Integration tests are passing locally:
```
➜  node git:(catch-all-errors) ✗ npx mocha test/python/aws-lambda-sdk/integration.test.js
2023-03-10T10:11:42.435Z ℹ test test uid: cdab


  Python: integration
2023-03-10T10:11:42.537Z ℹ test Creating core resources test-python-sdk-cdab
2023-03-10T10:12:14.874Z ℹ test Process function success-v3-8
2023-03-10T10:12:14.875Z ℹ test Process function success-v3-9
2023-03-10T10:12:14.875Z ℹ test Process function error-v3-8
2023-03-10T10:12:14.875Z ℹ test Process function error-v3-9
2023-03-10T10:12:14.876Z ℹ test Process function error_unhandled-v3-8
2023-03-10T10:12:14.876Z ℹ test Process function error_unhandled-v3-9
2023-03-10T10:12:14.876Z ℹ test Process function sdk-v3-8
2023-03-10T10:12:14.876Z ℹ test Process function sdk-v3-9
    ✔ success-v3-8 (30942ms)
    ✔ success-v3-9
    ✔ error-v3-8
    ✔ error-v3-9
    ✔ error_unhandled-v3-8
    ✔ error_unhandled-v3-9
    ✔ sdk-v3-8
    ✔ sdk-v3-9
2023-03-10T10:12:45.838Z ℹ test Cleanup test-python-sdk-cdab
2023-03-10T10:12:46.177Z ℹ test Deleted 30 version of the layer test-python-sdk-cdab-internal
2023-03-10T10:12:47.051Z ℹ test Detached IAM policy arn:aws:iam::692593327170:policy/test-python-sdk-cdab from test-python-sdk-cdab
2023-03-10T10:12:47.252Z ℹ test Deleted IAM role test-python-sdk-cdab
2023-03-10T10:12:47.709Z ℹ test Deleted IAM policy arn:aws:iam::692593327170:policy/test-python-sdk-cdab


  8 passing (1m)
```

medikoo (approved):
That's a great improvement @selcukcihan Thank you!

PR #511, opened 2023-03-09
Refactor SDK modules

selcukcihan:
@medikoo while working on https://linear.app/serverless/issue/SC-561/report-detailed-error-information-in-case-of-error-resolution I realized that modules in the base SDK are not organized well. For instance, we have a `serverless_sdk.span` module which contains things like `trace`, `tags` etc. I was going to add a `captured_event` module, at which point I realized this problem with module organization.

This will require changes in the AWS Lambda SDK package as well, please let me know how you'd prefer those changes, in a separate PR or amended to this PR?

### Description
1. Mimic the module organization of nodejs package
2. Move tests out of the SDK package, following the convention we have in the AWS Lambda package.

### Testing done
Unit tested.


medikoo (approved):
@selcukcihan Good call 👍 it's preferrable to issue changes to packages separately.

In this case it's even more important, as if I understand correctly we're introducing breaking changes, so first we need to release this update with v0.2 of SDK, to be able to adapt AWS Lambda SDK and ensure integration tests remain passing

selcukcihan:
> @selcukcihan Good call 👍 it's preferrable to issue changes to packages separately.
> 
> In this case it's even more important, as if I understand correctly we're introducing breaking changes, so first we need to release this update with v0.2 of SDK, to be able to adapt AWS Lambda SDK and ensure integration tests remain passing

Makes sense 👍 https://github.com/serverless/console/pull/513 for the release.

PR #512, opened 2023-03-09
Release AWS Lambda SDK v0.1.6

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-560/ensure-support-for-python-38#comment-76ad2b27

PR #515, opened 2023-03-09
Adapt lambda sdk changes

selcukcihan:
Related issue https://github.com/serverless/console/pull/512#issuecomment-1461951387

### Description
After breaking changes in the base SDK, update the lambda SDK.

### Testing done
Unit tested


selcukcihan:

medikoo (changes requested):
I think we also need to update dependency version (?)

selcukcihan:
> I think we also need to update dependency version (?)

Yeah you're right 👍  Ok, let's park this for now and first fix the CI so that we can release lambda SDK in the current state. I'll follow up with a PR for that

medikoo (approved):

PR #516, opened 2023-03-09
Fix unit test ci flow

selcukcihan:
Related issue https://github.com/serverless/console/pull/512

### Description
* Unit tests should install dependencies from PyPI repository. Both integrate and validate flow unit tests are now adhering to this principle.
* I've noticed that we depend on `serverless-sdk~=0.1` which caused `serverless-sdk v0.2.0' to be pulled and that broke the CI. This is also fixed to make sure it does not upgrade to newer minor version.

### Testing done
To be tested on the CI.

### DIFF
--- a/python/packages/aws-lambda-sdk/pyproject.toml
+++ b/python/packages/aws-lambda-sdk/pyproject.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 authors = [{ name = "serverlessinc" }]
 requires-python = ">=3.7"
 dependencies = [
-    "serverless-sdk~=0.1",
+    "serverless-sdk~=0.1.0",
     "serverless-sdk-schema~=0.1",
     "strenum~=0.4",
     "typing-extensions~=4.5", # included in Python 3.8 - 3.11
--- a/python/packages/aws-lambda-sdk/pyproject.toml
+++ b/python/packages/aws-lambda-sdk/pyproject.toml
@@ -14,7 +14,7 @@ authors = [{ name = "serverlessinc" }]
 requires-python = ">=3.7"
 dependencies = [
     "serverless-sdk~=0.1.0",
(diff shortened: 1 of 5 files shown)


selcukcihan:
@medikoo this is ready for review. I'd like to clarify something: https://semver.org

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> MAJOR version when you make incompatible API changes
> MINOR version when you add functionality in a backwards compatible manner
> PATCH version when you make backwards compatible bug fixes

But we seem to be bumping just the minor version for incompatible changes, where it's advised that a major version is used. Is this because we are still at version 0 ?

medikoo:
> But we seem to be bumping just the minor version for incompatible changes, where it's advised that a major version is used. Is this because we are still at version 0 ?

Yes, when we're at v0.x, rules are that MINOR is about breaking changes, and PATCH about everything else.

Still, at some point we'll consider this is stable, and then we'll switch to v1

python/packages/aws-lambda-sdk/pyproject.toml
```diff
@@ -13,7 +13,7 @@ readme = "README.md"
 authors = [{ name = "serverlessinc" }]
 requires-python = ">=3.7"
 dependencies = [
-    "serverless-sdk~=0.1",
+    "serverless-sdk~=0.1.0",
```
medikoo, line 16: Is this update needed?

selcukcihan: Yes, otherwise it installs v0.2 which breaks the tests for lambda sdk.

medikoo: Ok, so same way we should restrict sdk-schema

medikoo (approved):
👍

PR #518, opened 2023-03-10
Report error events

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-561/report-detailed-error-information-in-case-of-error-resolution

### Description
* Add `capture_error` sdk method.
* Support update of tags with a prefix

### Testing done
Unit tested.


selcukcihan:
@medikoo this is ready for review. I'll follow up with a release once this is in, and then finally will create another PR for actually using the `capture_error` while instrumenting customer code.

python/packages/sdk/serverless_sdk/__init__.py
```diff
@@ -46,5 +49,14 @@ def _create_trace_span(
     ) -> trace.TraceSpan:
         return trace.TraceSpan(name, input, output, start_time, tags)
 
+    def capture_error(self, error, **kwargs) -> CapturedEvent:
+        try:
+            _error = create_error_captured_event(error, **kwargs)
+            self._captured_events.append(_error)
+            return _error
+        except Exception:
+            # TODO report error
```
medikoo: What's the plan of addressing that `TODO`?

python/packages/sdk/serverless_sdk/lib/captured_event.py
```diff
@@ -0,0 +1,67 @@
+from __future__ import annotations
```
medikoo, line 1: Can we have some tests for this file and `error_captured_event` (?)

medikoo (changes requested):
@selcukcihan thank you! Looks very promising, I have just few remarks

python/packages/sdk/serverless_sdk/lib/captured_event.py
selcukcihan: Good point, I've added tests 👍

python/packages/sdk/serverless_sdk/__init__.py
selcukcihan: I've implemented that bit now 👍

python/packages/sdk/serverless_sdk/lib/error_captured_event.py
```diff
@@ -0,0 +1,55 @@
...
+    if isinstance(error, Exception):
+        _tags["name"] = builtins_type(error).__name__
+        _tags["message"] = str(error)
+    else:
+        _tags["name"] = name or builtins_type(error).__name__
+        _tags["message"] = str(error)
+    _tags["stacktrace"] = stack or resolve_stack_trace_string(error)
+    captured_event.tags.update(_tags, prefix="error")
+
+    logger.error(
```
medikoo, line 58: It looks as unconditionally logged, while this function might be called from internal error reporter which also logs this log unconditionally.

In Node.js version we fallback to logging only if it's an error explicitly reported by user (and not reported internally by the SDK) https://github.com/serverless/console/blob/430c6d08f603223c05ad71b9b44e58ea4415855d/node/packages/sdk/lib/create-error-captured-event.js#L40-L45

python/packages/sdk/serverless_sdk/__init__.py
```diff
@@ -46,5 +50,13 @@ def _create_trace_span(
     ) -> trace.TraceSpan:
         return trace.TraceSpan(name, input, output, start_time, tags)
 
+    def capture_error(self, error, **kwargs) -> CapturedEvent:
```
medikoo: I think it's better if we proposed this public method with next PR, at least it needs to also come with integration testing updated to cover it, and documentation

medikoo (changes requested):
Thanks @selcukcihan

python/packages/sdk/serverless_sdk/lib/error_captured_event.py
selcukcihan: Good catch, I've amended it 🙏

python/packages/sdk/serverless_sdk/__init__.py
selcukcihan: Cool, I've removed it now. But I don't think we can write integration test yet because we'd need the lambda sdk to actually make the call to capture_error.

medikoo: > But I don't think we can write integration test yet because we'd need the lambda sdk to actually make the call to capture_error.

True, integration test should be in follow up PR that concerns AWS Lambda SDK, and not part of SDK update 👍

python/packages/sdk/serverless_sdk/lib/error.py
```diff
@@ -0,0 +1,46 @@
+import os
```
medikoo, line 1: Why module is not named `report_error.py` ?

medikoo (commented):
@selcukcihan looks good, I just have one question towards file naming

python/packages/sdk/serverless_sdk/lib/error.py
selcukcihan: I think in NodeJS that makes sense as you can specify a default export. However, in Python when importing from a module you'd need to repeat the name like:

```
import report_error
report_error.report("stuff")
```

or

```
from report_error import report
report("stuff")
```

So I tend to choose module names as nouns that describe what the module is related to and put action in the object like the function in this case. I think this is better in terms of readability:

```
from error import report as report_error
report_error("stuff")
```

There is no repetition in the import statement

medikoo: Ok, thanks for explanations 👍

medikoo (approved):
Looks great 👍

PR #519, opened 2023-03-13
SC-584: Fix possible out of range error

Danwakeem:
## Description
Since our dev mode payload is a mix of spans and events it is not always guaranteed that a span will exist in the dev mode payload which could cause this to fail. I went ahead and added some extra guard rails here.


medikoo (approved):
👍

PR #520, opened 2023-03-13
Add event emitter to Python base SDK

selcukcihan:
This is not related to any particular issue.

### Description
Add pub/sub style event emitter functionality to be used for `captured-event` & `trace-span-close` events.

### Testing done
Unit tested


selcukcihan:

medikoo (commented):
Thanks @selcukcihan. 

In Node.js we use event emitter to internally track reported events and spans (so they can be reached as they happen whatever functionality needs that). It's not an interface that's expected to be consumed by user.

So my question is, Is this introduced here, as you like this idea of internal tracking and want to rely on similar interface internally, or is it because you've assumed it's an requirement for the SDK to expose?

selcukcihan:
> Thanks @selcukcihan.
> 
> In Node.js we use event emitter to internally track reported events and spans (so they can be reached as they happen whatever functionality needs that). It's not an interface that's expected to be consumed by user.
> 
> So my question is, Is this introduced here, as you like this idea of internal tracking and want to rely on similar interface internally, or is it because you've assumed it's an requirement for the SDK to expose?

Yes, I was aware that this is not part of the consumer facing API. I think it will ease the development of further functionality in the SDK, that's why I wanted to get it in.

medikoo (approved):
Thanks for explanations @selcukcihan, all clear 👍

PR #526, opened 2023-03-15
Use event emitter for storing captured events

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-311/python-sdk-captured-errors-and-warnings-api

### Description
Use the new functionality to store captured events.

### Testing done
Unit & integration tested

PR #527, opened 2023-03-15
Release python base sdk v0.3.0

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-564/python-sdk-custom-tags-api

@medikoo This should be merged after https://github.com/serverless/console/pull/526 to make sure lambda SDK does not break with the base v0.2.2, because this release contains a change that pushes storage of captured events to the lambda SDK. It's not a breaking change if we get https://github.com/serverless/console/pull/526 in first.

### DIFF
--- a/python/packages/sdk/CHANGELOG.md
+++ b/python/packages/sdk/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## 0.2.2 (2023-03-15)
+
+### Features
+
+- Add `serverlessSdk.set_tag` method, providing a way to set custom tags on the trace.
+
 ## 0.2.1 (2023-03-14)
 
 ### Features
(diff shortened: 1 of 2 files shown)


medikoo (changes requested):
@selcukcihan if it requires changes to be made in `aws_lambda_sdk` first  so it's not broken. It means it's a breaking change. In such case let's switch to v0.3

medikoo:
To further clarify:

Adding a new feature that doesn't affect how the package behaves for dependents, but for which to be used by dependent it needs to upgrade to a new version - is not breaking, but adding changes, which require changes on dependents side as otherwise they will  crash it's breaking

What you describe feels as the second case.

if we release an SDK v0.2.2 version that will crash the AWS Lambda SDK v0.1.7 (as published to PyPI), is a bug.

selcukcihan:
> To further clarify:
> 
> Adding a new feature that doesn't affect how the package behaves for dependents, but for which to be used by dependent it needs to upgrade to a new version - is not breaking, but adding changes, which require changes on dependents side as otherwise they will crash it's breaking
> 
> What you describe feels as the second case.
> 
> if we release an SDK v0.2.2 version that will crash the AWS Lambda SDK v0.1.7 (as published to PyPI), is a bug.

Got it, makes sense 👍

medikoo (approved):
👍

PR #529, opened 2023-03-15
Enhance build script

selcukcihan:
Related issues
* https://linear.app/serverless/issue/SC-605/python-sdk-build-layer-script-crashes-when-not-invoked-from-its
* https://linear.app/serverless/issue/SC-606/python-sdk-very-large-extension-layer-size

### Description
* The script will return an error if `dist` folder does not contain an installation or if the folder itself is missing.
* The script will remove the zip archive if the `dist` folder already contains it.

### Testing done
* Tested with missing dist folder, errors out
* Tested with missing installation but dist folder present, errors out
* Tested with existing zip file in dist, the zip file is recreated but its size does not increase as it does not include the previous zip in it
* Tested with happy path, works as expected.

### DIFF
--- a/python/packages/aws-lambda-sdk/scripts/build-layer-archive.sh
+++ b/python/packages/aws-lambda-sdk/scripts/build-layer-archive.sh
@@ -1,7 +1,13 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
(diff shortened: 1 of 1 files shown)


medikoo (approved):
Thank you @selcukcihan 👍

PR #535, opened 2023-03-16
Add capture_warnings, origin and fingerprint

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-311/python-sdk-captured-errors-and-warnings-api

### Description
* Add `capture_warning` API.
* Add `fingerprint` parameter.
* Add `origin` parameter
* Fix `nodeConsole` as `pythonConsole`

### Testing done
Unit tested

### DIFF
--- a/python/packages/sdk/serverless_sdk/lib/error_captured_event.py
+++ b/python/packages/sdk/serverless_sdk/lib/error_captured_event.py
@@ -26,6 +26,7 @@ def create(
     name=None,
     stack=None,
     origin: Optional[str] = None,
+    fingerprint: Optional[str] = None,
 ):
     timestamp = timestamp or time.perf_counter_ns()
     tags = tags or Tags()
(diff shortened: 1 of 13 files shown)


selcukcihan:

python/packages/sdk/serverless_sdk/lib/error_captured_event.py
```diff
@@ -51,19 +53,20 @@ def create(
     from .. import serverlessSdk
 
     if (
-        origin == "nodeConsole"
+        origin == "pythonConsole"
```
medikoo, line 56: In Node.js this marks logs coming from captured `console.*` (native Node.js interface) calls. 

At this moment it's not applicable in Python, although one of the next steps will be to instrument default login interface in Python (I will open an issue for that)

medikoo (approved):
Great work @selcukcihan  👍

python/packages/sdk/serverless_sdk/lib/error_captured_event.py
medikoo: Additional note: Just realized we're using to also to mark events which should not be additionally written to `console.*`. So that means this setting as is a bit overused in Node.js

medikoo (approved):

PR #537, opened 2023-03-16
test(Go SDK): Set up integration tests

czubocha:
* add a success case of integration test for two runtimes: `go1.x` and `provided.al2`
* fix the Go SDK to `aws.lambda` end time match `aws.lambda.invocation` end time (`Close(t ...time.Time)` instead of `Close()`)
* do not execute `Array.from` on `options.layersConfig` if `options.layersConfig` is undefined (`node/test/lib/create-core-resources.js`)


node/test/go/aws-lambda-sdk/build-go-binary.sh
medikoo, line 1: Can we have this build script in context of go packages? 

e.g. currently Node.js SDK build script is at `node/packages/aws-sdk-lambda`, Python SDK build script is at `python/packages/aws-sdk-lambda`, and I think same way should be with Go

node/test/go/aws-lambda-sdk/integration.test.js
```diff
@@ -0,0 +1,225 @@
...
+      const outputDir = path.resolve(moduleDir, 'build', architecture);
+      const zipFile = await resolveDirZipBuffer(outputDir);
+      const isArm64 = architecture === 'arm64';
+      return {
+        Runtime: isArm64 ? 'provided.al2' : 'go1.x',
+        Architectures: isArm64 ? ['arm64'] : ['x86_64'],
+        Code: { ZipFile: zipFile },
+      };
+    };
+  }
```
medikoo: Ideally if zip file generation is done via some script in context of Go package. So we can build the layer manually, being in context of the Go package (and don't have to run Node.js integration test for that)

node/test/go/aws-lambda-sdk/integration.test.js
```diff
@@ -0,0 +1,225 @@
...
+            },
+          ],
+        ]),
+      },
+    ],
+  ]);
+
+  const testVariantsConfig = resolveTestVariantsConfig(useCasesConfig, {
+    commonTestConfig: {
+      configuration: { Handler: 'bootstrap' },
```
medikoo, line 73: So with Go SDK, user is required to build the handler together with our SDK? We cannot inject in similar non-destructive way as it's with Node.js and Python, right?

medikoo (commented):
Great work @czubocha 👍 I have just a few  code organization concerns

medikoo:
Additionally, wouldn't it be good to configure CI for that? (but that can handled in next PR)

node/test/go/aws-lambda-sdk/integration.test.js
czubocha: Go SDK won't have a layer. This part is just preparing the Lambda deployment package (building the code and zipping the binary). I'll move the shell script to `go` directory but there will be no need to run this script in another context than the integration test.

node/test/go/aws-lambda-sdk/integration.test.js
czubocha: Yes, exactly.
And `bootstrap` handler name is a requirement of `provided.al2` Lambda runtime. The rule is:
* Amazon Linux 1: `go1.x` runtime
* Amazon Linux 2: `provided.al2` runtime
> As one of the later supported runtimes for Lambda, Go is implemented differently than other native runtimes. Under the hood, Go is treated as a custom runtime and runs accordingly. A Go developer can take advantage of this by choosing the provided.al2 runtime and providing the required bootstrap file. 

https://aws.amazon.com/blogs/compute/migrating-aws-lambda-functions-to-al2/

node/test/go/aws-lambda-sdk/integration.test.js
medikoo: Ok, so this script bundles fixtures (as located in go package) with the SDK (?) Then it's indeed specific to testing, but as we keep fixtures in Go package, I believe it's best if it's put somewhere aside them (?)

node/test/go/aws-lambda-sdk/integration.test.js
medikoo: Great thanks for clarifications 👍

node/test/go/aws-lambda-sdk/integration.test.js
czubocha: Moved to `go/packages/slslambda/scripts/build-go-binary.sh`

medikoo (approved):
Great work @czubocha 👍

PR #542, opened 2023-03-17
Fix custom tags serialization

selcukcihan:
### Description
While trying to extend the integration tests, I've realized a difference in how `customTags` are serialized. This fixes the issue, by not converting custom tags and using them as is.

### Testing done
Unit tested


selcukcihan:

medikoo (approved):
👍

PR #544, opened 2023-03-17
Fix serialization of custom tags on events

selcukcihan:
### Description

Further fix similar to https://github.com/serverless/console/pull/542

Custom tag names should not be processed, should be used as is. Example: "user.tag" should not be converted to:
```
{
  "user": {
    "tag": ...
```

It should appear as is: `{ "user.tag": ... }`


selcukcihan:

medikoo (approved):
👍

PR #552, opened 2023-03-20
Fix python logging source name

selcukcihan:
### Description
While trying out the latest officially published layer `arn:aws:lambda:eu-west-1:177335420605:layer:sls-sdk-python-v0-1-9:1` I've noticed my `logging.error` call resulted in two log lines instead of a single one:

```
[ERROR]	2023-03-20T13:59:55.936Z	e4511c1e-5c22-48d3-a93f-925d8763b334	Hello world!
[ERROR]	2023-03-20T13:59:55.937Z	e4511c1e-5c22-48d3-a93f-925d8763b334	{'source': 'serverlessSdk', 'type': 'ERROR_TYPE_CAUGHT_USER', 'name': 'str', 'message': 'Hello world!', 'stack': '  File "/opt/sls-sdk-python/exec_wrapper.py", line 25, in <module>\n    main()\n  File "/opt/sls-sdk-python/exec_wrapper.py", line 21, in main\n    bootstrap.main()\n  File "/var/runtime/bootstrap.py", line 57, in main\n    awslambdaricmain.main([os.environ["LAMBDA_TASK_ROOT"], os.environ["_HANDLER"]])\n  File "/var/runtime/awslambdaric/__main__.py", line 21, in main\n    bootstrap.run(app_root, handler, lambda_runtime_api_addr)\n  File "/var/runtime/awslambdaric/bootstrap.py", line 405, in run\n    handle_event_request(\n  File "/var/runtime/awslambdaric/bootstrap.py", line 149, in handle_event_request\n    response = request_handler(event, lambda_context)\n  File "/opt/python/serverless_aws_lambda_sdk/instrument/__init__.py", line 213, in stub\n    result = user_handler(event, context)\n  File "/var/task/mymodule/lambda_function.py", line 7, in lambda_handler\n    logging.error("Hello world!")\n  File "/var/lang/lib/python3.9/logging/__init__.py", line 2064, in error\n    root.error(msg, *args, **kwargs)\n'}
```

As you can see, the second line is from `create_error_captured_event`, but it should have skipped the logging part as this is already logged. The problem is that I was using `pythonConsole` as the source name in the base SDK, but the lambda SDK refers to the same source as `pythonLogging`.

I think it makes more sense to stick to `pythonLogging` as we are instrumenting the logging module in Python, so this fixes the base SDK to use the correct source for these logs.

### Testing done
Unit & integration tested.

```
[...]

PR #554, opened 2023-03-20
Upgrade & test behind the scenes logging

selcukcihan:
Related issue https://github.com/serverless/console/pull/552

### Description
Upgrade to use the latest base sdk version to have the logging fix

### Testing done
Reproduced the issue with a unit test and now asserting on the number of actual logging calls made to make sure we are not logging instrumented calls which are already logged by the platform logger.


selcukcihan:

medikoo (approved):
👍

PR #556, opened 2023-03-21
Python sdk dev mode (RequestResponse)

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-563/python-sdk-support-dev-mode

### Description
* Report request and response payloads in dev mode
* Report captured events and spans within trace payloads in dev mode
* use asyncio to handle network requests asynchronously
* Rearrange unit tests to follow similar structure as in the package
* Buffer payloads for 50ms before sending the request to dev mode extension
* Dev mode related modules are only loaded if the dev mode flag is on, to optimize for production workflows

### Testing done
Unit & integration tested
```
➜  node git:(blocking-dev-mode) ✗ npx mocha test/python/aws-lambda-sdk/integration.test.js
2023-03-23T12:43:35.846Z ℹ test test uid: cihan


  Python: integration
2023-03-23T12:43:35.952Z ℹ test Creating core resources test-python-sdk-cihan
2023-03-23T12:44:19.405Z ℹ test Process function success-v3-8
2023-03-23T12:44:19.406Z ℹ test Process function success-v3-9
2023-03-23T12:44:19.407Z ℹ test Process function success-sampled
2023-03-23T12:44:19.407Z ℹ test Process function error-v3-8
2023-03-23T12:44:19.407Z ℹ test Process function error-v3-9
2023-03-23T12:44:19.407Z ℹ test Process function error_unhandled-v3-8
2023-03-23T12:44:19.408Z ℹ test Process function error_unhandled-v3-9
2023-03-23T12:44:19.409Z ℹ test Process function sdk-v3-8
2023-03-23T12:44:19.409Z ℹ test Process function sdk-v3-9
2023-03-23T12:44:19.409Z ℹ test Process function sdk-dev-mode
    ✔ success-v3-8 (20361ms)
    ✔ success-v3-9 (1094ms)
    ✔ success-sampled (5008ms)
    ✔ error-v3-8 (730ms)
    ✔ error-v3-9 (4850ms)
    ✔ error_unhandled-v3-8
    ✔ error_unhandled-v3-9
    ✔ sdk-v3-8
    ✔ sdk-v3-9
    ✔ sdk-dev-mode (5607ms)
2023-03-23T12:44:57.087Z ℹ test Cleanup test-python-sdk-cihan
2023-03-23T12:44:57.432Z ℹ test Deleted 8 version of the layer test-python-sdk-cihan-external
2023-03-23T12:44:57.433Z ℹ test Deleted 26 version of the layer test-python-sdk-cihan-internal
[...]

PR #562, opened 2023-03-23
Add report_notice method

selcukcihan:
Related issue https://github.com/serverless/console/pull/556#discussion_r1144674640

### Description
* Add `_report_notice` method on base sdk

### Testing done
Unit tested


selcukcihan:

medikoo (commented):
@selcukcihan can we split this into two PR's? 

Ideally if different unrelated enhancements are proposed with individual PR's. That way one doesn't block another in case of issues and also in follow up commit history they end as distinct commits

medikoo (approved):
Looks good 👍

PR #566, opened 2023-03-27
Remove pydantic dependency

selcukcihan:
I've noticed two dependencies we list but we don't use. Removal of these (pydantic is huge and pulls other transitive dependencies as well) reduces the layer size from 5.6M down to 1.2M.

### Testing done
Integration tested.


selcukcihan:

medikoo (approved):
Great find 👍

PR #570, opened 2023-03-27
Support only python 3.8+ and remove unnecessary dependencies

selcukcihan:
### Description

I've found out that a significant amount of time is spent when we import `importlib_metadata` just to extract the package version of our SDKs:

Screenshot 2023-03-27 at 21 43 24

We can remove that dependency and improve both layer size and the init latency. I've also removed other dependencies that are only required for Python3.7, since we support 3.8+.

### Testing done
CI flows should cover testing.

### DIFF
--- a/.github/workflows/python-integrate.yaml
+++ b/.github/workflows/python-integrate.yaml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python_version: ["3.7", "3.8", "3.9"]
+        python_version: ["3.8", "3.9"]
     outputs:
       pathChangedAwsLambdaSdk: ${{ steps.pathChanges.outputs.awsLambdaSdk }}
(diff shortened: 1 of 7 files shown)


selcukcihan:

medikoo (commented):
@selcukcihan it's likely that we will want to support Python v3.7 but simply via manual instrumentation. So let's not drop the support for that.

Otherwise, what speed improvement do you observe after this change? Have the latency for initialization dropped from 800ms?

selcukcihan:
> @selcukcihan it's likely that we will want to support Python v3.7 but simply via manual instrumentation. So let's not drop the support for that.
> 
> Otherwise, what speed improvement do you observe after this change? Have the latency for initialization dropped from 800ms?

It makes a difference of around 100ms. We can do it as in https://github.com/serverless/console/pull/572 & https://github.com/serverless/console/pull/573

selcukcihan:
I've isolated the latency to be from dynamically fetching the package name, I'll create a separate PR with just that change. Closing this.

PR #574, opened 2023-03-28
Use static name for sdk package

selcukcihan:

### Description
`packages_distributions` call is causing some latency which we don't need, we can use the package name as a constant.

### Testing done
Unit & Integration tested


selcukcihan:

medikoo (approved):
👍

PR #576, opened 2023-03-28
Switch to protoc for compiling protobuf

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-593/python-sdk-big-latency-overhead

### Description
Switch from https://github.com/danielgtaylor/python-betterproto to https://protobuf.dev/reference/python/python-generated/

### Testing done
I've unit & integration tested generated sdk-schema through the lambda sdk package.

PR #579, opened 2023-03-28
chore: Release python sdk-schema v0.2.0

selcukcihan:
Release protobuf changes introduced with https://github.com/serverless/console/pull/576

### DIFF
--- a/python/packages/sdk-schema/CHANGELOG.md
+++ b/python/packages/sdk-schema/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### 0.2.0 (2023-03-28)
+
+### ⚠ BREAKING CHANGES
+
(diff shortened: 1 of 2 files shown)


selcukcihan:
@medikoo ready for review 🙏

medikoo (approved):
👍

python/packages/sdk-schema/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### 0.2.0 (2023-03-28)
+
+### ⚠ BREAKING CHANGES
+
+- Switch to protoc compiler instead of betterproto
```
medikoo, line 9: Why exactly it's a breaking change? Does it change the way in how this dependency is used once installed?

selcukcihan: It's a breaking change, because the generated classes are different, they have different attributes and methods etc. It also does change how the dependency is used in terms of import path.

PR #583, opened 2023-03-29
[Python AWS Lambda SDK] Enable performance tests

selcukcihan:


selcukcihan:
@medikoo ready for review please 🙏  Would be good if we can merge it after https://github.com/serverless/console/pull/584 so that it runs against the increased threshold.

medikoo (approved):
👍

PR #585, opened 2023-03-29
AWS Lambda SDK: Write `sdk.runtime` tag

medikoo:


node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -84,7 +84,7 @@ const reportResponse = async (response, context, endTime) => {
     slsTags: {
       orgId: serverlessSdk.orgId,
       service: process.env.AWS_LAMBDA_FUNCTION_NAME,
-      sdk: { name: pkgJson.name, version: pkgJson.version },
+      sdk: { name: pkgJson.name, version: pkgJson.version, runtime: 'nodejs' },
```
Danwakeem, line 87: Is there a way for us to get the specific nodejs runtime? It might be nice to have that way we know which runtime the trace came from 🤷

Danwakeem (commented):
Not sure my comment is part of the requirements or whatever but it might be nice to capture 🤷

node/packages/aws-lambda-sdk/instrument/index.js
medikoo: @Danwakeem we've discussed that in Linear. All of this will be read from `INIT START` log by the backend (on SDK level we do not have access to this log)

Danwakeem (approved):
🎸

PR #588, opened 2023-03-30
Fix error handling in tags

selcukcihan:
Related issue https://linear.app/serverless/issue/SC-693/tag-validation-error-on-capture-error-is-wrong-type

### Description
We had a bug where we treat a user error as internal SDK error. This was in the `Tags` implementation. Public methods for setting tags will now report a user error. Private methods will raise an exception and we handle them accordingly.

### Testing done
Unit & integration tested


selcukcihan:

medikoo (approved):
👍

PR #596, opened 2023-04-03
Instrument http requests

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-629/python-sdk-instrument-https-requests
* Instrument http[s] requests by
  * intercepting calls to `http.client.HTTPConnection` to support `http.client`, `urllib`, `urllib3` & `requests` http client modules.
  * using the tracer functionality of `aiohttp` to support `aiohttp` async client.

### Testing done
* Unit tested & integration tested. Integration tests will be in a separate PR together with the required changes in the lambda SDK.8
* I've tried it with the console, but even though the tags are set on the span (I verified by deserializing the printed trace and checking the actual values), console does not display the http span tags. I'm guessing this is because console does not recognize `python.http.request` spans.

Screenshot 2023-04-10 at 13 22 12


selcukcihan:
@Mmarzex ready for review please 🙏

Mmarzex (approved):

PR #597, opened 2023-04-03
fix: Ensure aws-lambda-sdk types match aws-sdk

Mmarzex:
When we fixed the [types for serverless-sdk](https://github.com/serverless/console/pull/371). We never applied this fix to the `serverless-lambda-sdk`.


Danwakeem (approved):
👍

PR #609, opened 2023-04-05
Improve wrapper and clean up unnecessary code

selcukcihan:
### Description
* Make sure we do the checks on customer function only for the first invocation (if function exists and if it's a callable).
* Clean up unused code.

### Testing done
* Unit/Integration tested
* Manual testing of the case when function does not exist, no change in output & behaviour as expected


selcukcihan:

medikoo (approved):
👍

PR #610, opened 2023-04-05
AWS Lambda SDK: Improved sampling algorithm

medikoo:
- Recognition of ALB HTTP event
- Improvement to sampling algorithm according to internal agreements

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -11,6 +11,7 @@ const sendTelemetry = require('./lib/send-telemetry');
 const flushSpans = require('./lib/auto-send-spans').flush;
 const filterCapturedEvent = require('./lib/filter-captured-event');
 const invocationContextAccessor = require('./lib/invocation-context-accessor');
+const isApiEvent = require('./lib/is-api-event');
 const pkgJson = require('../package');
 
 const serverlessSdk = require('./lib/sdk');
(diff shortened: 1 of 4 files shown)


node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -98,15 +99,27 @@ const reportResponse = async (response, context, endTime) => {
   await sendTelemetry('request-response', payloadBuffer);
 };
 
+let requestCounter = 0;
+let lastCounterResetTime = 0;
+
 const reportTrace = ({ isErrorOutcome }) => {
   // Sample out 80% of traces in production mode if no warning or error event was reported
```
Danwakeem: I think I am following looking at the code but not 100% sure. I think an updated comment would help 🙏 

Does this sound right?

```
/**
 * This function determines if a trace should be sampled out or not
 * 
 * 
 * Traces will be sampled out when
 *    Trace contains Errors or Warnings
 *    Dev Mode = enabled
 *    Debug Mode = enabled
 *    Cold start
 *    Lambda is an API and has received >2 requests within one second
 * 
 * Given none of the conditions above are met we will output 90% of traces
*/
```

~~I think I am missing the part about `isApiEvent`~~ 🤔 Updated to include the API event exclusion

node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -98,15 +99,27 @@ const reportResponse = async (response, context, endTime) => {
...
+      if (serverlessSdk._isDebugMode) return false;
+      if (serverlessSdk._isDevMode) return false;
+      if (capturedEvents.some(({ name }) => alertEventNames.has(name))) return false;
+      const currentTime = Date.now();
+      if (currentTime - lastCounterResetTime > 1000) {
+        lastCounterResetTime = currentTime;
+        requestCounter = 0;
+      }
+      ++requestCounter;
+      if (isApiEvent()) return requestCounter > 2;
```
Danwakeem, line 126: Does this mean that after the first 2 API requests all other requests within 1 second will be reported?

Danwakeem (commented):
This is clever, I like it 😎 But I think the logic is complicated enough it might be worth updating the comment.

node/packages/aws-lambda-sdk/instrument/index.js
medikoo: After 2 API requests they won't be reported (`true` resolves to sampled out)

Danwakeem: Ahh okay got I had it backwards. Thanks for the clarification 👍

node/packages/aws-lambda-sdk/instrument/index.js
medikoo: Good point, I forgot to update this comment. I've updated it now

Danwakeem (approved):
Looks great :shipit:

PR #611, opened 2023-04-06
AWS Lambda SDK: Ensure to not error continuously on internal error.

medikoo:
When internal error happened, due to flaw in logic the next invocation was also destined to crash. This patch ensures that state for the next invocation should be properly cleared


Danwakeem (approved):
Nice find 👍 is this related to the issue Eslam posted about or is this something different?

medikoo:
>Nice find 👍 is this related to the issue Eslam posted about or is this something different?

@Danwakeem it's about one reported by @rudeluv

Danwakeem:
Ahh okay got it, thanks for the clarification @medikoo.

PR #613, opened 2023-04-06
Match nodejs sdk behaviour for empty custom tags

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-768/python-sdk-not-capturing-events-in-devmode-in-some-cases
* While doing investigation, I've noticed a slight difference in how we report traces. For spans without any custom tags, we were reporting a json string that represents the empty object `{}`. This is now removed and we do not include `custom_tags` field if there are no custom tags, like in NodeJS SDK.

### Testing done
Reproduced in a unit test.


selcukcihan:

medikoo (approved):
👍

PR #617, opened 2023-04-06
Stick with sls prefix only

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-708#comment-246509d6
* Remove `serverless_sdk` module and stick with `sls_sdk` only.

### Testing done
Unit tested


selcukcihan:
@medikoo ready for review please 🙏 I'll follow up with another release of base sdk (breaking change) and then finally release the python sdk that gets rid of any reference to `serverless_sdk`.

medikoo (commented):
@selcukcihan let's hold on that until we have full agreement on Linear, on the scope of the change.

selcukcihan:
> @selcukcihan let's hold on that until we have full agreement on Linear, on the scope of the change.

Sure, would you like me to release the ~~Python~~ Lambda SDK nevertheless? Or should that wait as well, please let me know.

medikoo:
>Sure, would you like me to release the Python Lambda SDK nevertheless? Or should that wait as well, please let me know.

Let's follow up with the release. It's really important to have  fixes we've merged released.

Rename is nothing crucial, as we can just document one endpoint for users. Whether it's one, or two ways that work behind the scenes is irrelevant at this point.

medikoo:
@selcukcihan this change is not necessary, I'm closing this PR

PR #620, opened 2023-04-06
SDK: Write JSON strings as structured logs

medikoo:
While JS object input also works in CW for structured logging, to ensure consistency among runtimes and follow [AWS guidelines](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-logging.html) write structured logs as JSON string


Danwakeem (approved):
Reviewing from my phone but looks good from here 😎

PR #628, opened 2023-04-07
feat: Adjust default log parsing

Danwakeem:
## Description
We need to adjust the way we detect the default logs because the format in Cloudwatch is different between python and nodejs. I am not sure if I should mark this as a feat or a fix because it is a bug but only because of a new feature 🤔 So I guess I will just mark it as feat 🤷


Danwakeem:
Maciej is requested to review this so I am going to go ahead and merge this in 😎 Feel free to add any comments after the fact and I will address with a follow up PR.

PR #631, opened 2023-04-10
fix: SC-799 Added capture warning when spans are not closed

Danwakeem:
## Description
Updated the span not closed warning to produce a warning event

> I will create an integration test for this after I create a new version of the sdk and create the `aws-lambda-sdk` PR to use the new sdk version 👍

### DIFF
--- a/node/packages/sdk/lib/trace-span.js
+++ b/node/packages/sdk/lib/trace-span.js
@@ -43,6 +43,7 @@ class TraceSpan {
         );
       }
     }
+    this._reportWarning = options._reportWarning;
     this.startTime = startTime || defaultStartTime;
     this.name = ensureSpanName(name);
 
@@ -85,6 +86,7 @@ class TraceSpan {
       new TraceSpan(immediateDescendants.shift(), {
         startTime: this.startTime,
         immediateDescendants,
+        _reportWarning: this._reportWarning,
       });
     }
   }
@@ -132,10 +134,10 @@ class TraceSpan {
         leftoverSpans.push(subSpan.close({ endTime: this.endTime }));
       }
       if (leftoverSpans.length) {
-        process.stderr.write(
+        const message =
           "Serverless SDK Warning: Following trace spans didn't end before end of " +
-            `lambda invocation: ${leftoverSpans.map(({ name }) => name).join(', ')}\n`
-        );
+          `lambda invocation: ${leftoverSpans.map(({ name }) => name).join(', ')}\n`;
+        this._reportWarning(message, 'SDK_SPAN_NOT_CLOSED');
       }
       asyncLocalStorage.enterWith(this);
     } else {
--- a/node/packages/sdk/index.js
+++ b/node/packages/sdk/index.js
@@ -110,7 +110,13 @@ serverlessSdk._initialize = (options = {}) => {
   return serverlessSdk;
 };
 
-serverlessSdk._createTraceSpan = (name, options = {}) => new TraceSpan(name, options);
+serverlessSdk._createTraceSpan = (name, options = {}) =>
+  new TraceSpan(
+    name,
+    Object.assign(options, {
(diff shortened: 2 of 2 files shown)


Danwakeem:
Just trying to run the tests in git because what I am seeing locally doesn't make sense 😅

Danwakeem:
Okay this is what I am seeing locally 🤔 I am guessing this is because of some circular dependency or something.

node/packages/sdk/lib/trace-span.js
```diff
@@ -132,10 +132,10 @@ class TraceSpan {
         leftoverSpans.push(subSpan.close({ endTime: this.endTime }));
       }
       if (leftoverSpans.length) {
-        process.stderr.write(
+        const message =
           "Serverless SDK Warning: Following trace spans didn't end before end of " +
-            `lambda invocation: ${leftoverSpans.map(({ name }) => name).join(', ')}\n`
-        );
+          `lambda invocation: ${leftoverSpans.map(({ name }) => name).join(', ')}\n`;
+        require('./report-warning')(message, 'SDK_SPAN_NOT_CLOSED');
```
Danwakeem: I am assuming because of the circular dependency it gets mad when I put this require at the top of the file. 

Not sure if there is a better way to do this with out duplicating any code though 🤔 Kind of have a feeling this might be a bad idea 😬

Danwakeem:
I can't seem to get the integration test to pass in this branch or in main 😢 I just did a quick test and I am able to get them to pass in a separate branch on github actions so I must be missing something locally? 🤷

Mmarzex:
@Danwakeem Do you have a link to one of the failed runs?

Danwakeem:
@Mmarzex No because it was working when I tested in github actions just not when I ran them locally 😢

Mmarzex:
>  @Mmarzex No because it was working when I tested in github actions just not when I ran them locally 😢

Ah, sorry I misunderstood. Do you see any errors or is it just failing?

Danwakeem:
Yeah when I run the tests locally a few of the tests pass but a good chunk of them give me errors ranging from security tokens being out of date, timeouts, etc. TBH it is kind of hard to read because of all the output.

I am getting the same result if I run the tests against `main` or this branch so I am sure I am doing something wrong I just have no idea what it is.

My test command looks like this in the `node/packages/aws-lambda-sdk` folder 👇 

```
AWS_PROFILE= AWS_ACCESS_KEY_ID=*** AWS_SECRET_ACCESS_KEY=*** AWS_REGION=us-east-1 SLS_ORG_ID=123 TEST_UID=vzC4DR npm run test:integration
```

> AWS_PROFILE intentionally left blank so it will override the global AWS_PROFILE I have set on my computer 😎 

I limited the `useCasesConfig` to only the `express` use case and I see this error 👇 
Screenshot 2023-04-11 at 2 37 07 PM

Danwakeem:
The downside without being able to run these locally I can't really test this SDK change via integrations tests until we create an sdk release 😢 

Usually I can test the SDK changes early by modifying the `node_modules/@serverless/sdk` files locally and then run the integration tests with the updated sdk but just having trouble running the tests locally.

Mmarzex:
Weird, I've never had luck running the tests locally... Not sure why it would be causing problems on what you are doing though.

Danwakeem:
@Mmarzex I found [a handy website](https://gitpkg.vercel.app/) that would give me the npm install command to install the branch version of the `@serverless/sdk` in the `aws-lambda-sdk` folder.

So I ran the [integration tests early on github actions](https://github.com/serverless/console/actions/runs/4672999884) and it looks like with the expected changes in the `aws-lambda-sdk` folder the integration tests pass and we see the event we are expecting so I am pretty sure this should be good 🎉

Mmarzex (approved):

PR #637, opened 2023-04-13
Instrument flask

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-691/python-sdk-instrument-flask
* Instrument flask on-demand: Whenever flask is imported, if instrumentation is enabled
* This will also fix https://linear.app/serverless/issue/SC-698/python-sdk-instrument-flask-unhandled-exceptions as we know intercept these exceptions and capture them.

### Testing done
Unit/Integration tested

### Further work
* A separate PR will cover the lambda SDK once this feature is released in the base SDK.
* `aws.lambda.http_router.path` will need to be overriden, once we implement these tags.


selcukcihan:
@Mmarzex ready for review please

python/packages/sdk/sls_sdk/lib/instrumentation/flask.py
```diff
@@ -0,0 +1,155 @@
...
+            (self._flask.appcontext_popped, self._handle_appcontext_popped),
+        ]
+        self._root_span = None
+        self._route_span = None
+        self._error_span = None
+        self._reported_exception = None
+
+    def sanitize_span_name(self, name):
+        if name is None:
+            return "unknown"
```
Mmarzex, line 32: If a span name is `unknown` do we still send it with the final trace payload?

Mmarzex (changes requested):
Looks good @selcukcihan 👍 . Had a question in the code, as well as wanted to clarify, are we not attaching any of the tags in this PR for the API paths?

python/packages/sdk/sls_sdk/lib/instrumentation/flask.py
selcukcihan: Yes, we still send it. I think it's the same behaviour as in nodejs SDK https://github.com/serverless/console/blob/910cbf53b638b1fa51bcdf2da54cb8666a68c27f/node/packages/sdk/lib/instrumentation/express/instrument-layer-prototype.js#L48 I couldn't find a filter that filters out "unknown" spans.

selcukcihan:
> Looks good @selcukcihan 👍 . Had a question in the code, as well as wanted to clarify, are we not attaching any of the tags in this PR for the API paths?

Could you expand on it a bit please? If you mean `aws.lambda.http_router.path` tags, then yes it's not part of this PR (yet, if you disagree): We haven't implemented the lambda event tags (I've created https://linear.app/serverless/issue/SC-844/aws-lambda-python-sdk-add-lambda-eventresponse-tags today to address that) yet, I thought this might be tackled as part of that task.

Mmarzex:
Gotcha. I think that is fine to do in a follow up, but we will need that router path attached to the root span in order to support API filtering.

Mmarzex (approved):

PR #638, opened 2023-04-13
feat: Capture Structured Logs as Errors & Warnings

Mmarzex:
Introduces instrumentation of `process.stdout.write` and `process.stderr.write` so that we can attempt to capture errors & warning events from popular structured logging libraries. 

It has been tested against,
1. pino
2. winston
3. bunyan
4. AWS Powertools Logger

We are guaranteeing support of those 4 but the logic is in place to attempt parsing of any.

Screen Shot 2023-04-13 at 10 42 32 AM

### DIFF
--- a/node/packages/sdk/lib/structured-log-to-event.js
+++ b/node/packages/sdk/lib/structured-log-to-event.js
@@ -1,0 +1,135 @@
+'use strict';
+
+const isObject = require('type/object/is');
+const { debuglog } = require('util');
+const createErrorCapturedEvent = require('./create-error-captured-event');
+const createWarningCapturedEvent = require('./create-warning-captured-event');
+
+const parseLogLevel = (level) => {
+  if (typeof level === 'string') {
+    const levelUpperCase = level.toUpperCase();
+    if (levelUpperCase !== 'ERROR' && levelUpperCase !== 'WARN') {
+      throw new Error('Unsupported level');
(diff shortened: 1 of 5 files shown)


node/packages/sdk/lib/structured-log-to-event.js
```diff
@@ -0,0 +1,135 @@
...
+
+  if (msg) {
+    createWarningCapturedEvent(msg, { tags, _origin: 'nodeConsole' });
+  }
+};
+
+module.exports.attemptParseStructuredLogAndCapture = (logLine) => {
+  try {
+    const logLineParsed = JSON.parse(logLine.toString());
+    // console.log(logLineParsed);
```
Danwakeem:
```suggestion
```

Danwakeem (commented):
This looks good

Side note, I will need to update how dev mode filters out log messages before we release this so we don't see duplicates 🤔

Mmarzex:
@Danwakeem So we'd just rely on these error events coming through to dev mode now you are thinking?

Danwakeem:
Not exactly, we will still need to capture logs that are not warnings or errors. So I think I will need to copy this logic a bit in the dev mode extension so that I know when to ignore warnings/errors from these libraries.

Danwakeem (approved):

PR #670, opened 2023-04-20
SDK: Improve `serverlessSdk._reportWarning` resolution

medikoo:
`_reportWarning` ideally should always be accessed directly from the SDK instance.

### DIFF
--- a/node/packages/sdk/lib/trace-span.js
+++ b/node/packages/sdk/lib/trace-span.js
@@ -43,7 +43,6 @@ class TraceSpan {
         );
       }
     }
-    this._reportWarning = options._reportWarning;
     this.startTime = startTime || defaultStartTime;
     this.name = ensureSpanName(name);
 
@@ -86,7 +85,6 @@ class TraceSpan {
       new TraceSpan(immediateDescendants.shift(), {
         startTime: this.startTime,
         immediateDescendants,
-        _reportWarning: this._reportWarning,
       });
     }
   }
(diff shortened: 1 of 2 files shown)


node/packages/sdk/lib/trace-span.js
```diff
@@ -137,7 +135,8 @@ class TraceSpan {
         const message =
           "Serverless SDK Warning: Following trace spans didn't end before end of " +
           `lambda invocation: ${leftoverSpans.map(({ name }) => name).join(', ')}\n`;
-        this._reportWarning(message, 'SDK_SPAN_NOT_CLOSED', { type: 'USER' });
+        // Require on spot to avoid otherwise difficult to mitigate circular dependency
+        require('..')._reportWarning(message, 'SDK_SPAN_NOT_CLOSED', { type: 'USER' });
```
Danwakeem, line 139: Oh man 😆 I did this originally but I thought it was not good practice to require on the spot so I went the other way.

Danwakeem (approved):
Looking good 👍 I wasn't sure about the rules for requiring on the spot so I didn't go that route.

node/packages/sdk/lib/trace-span.js
medikoo: If it can be avoided it's best to avoid it. Still, I think it's better than passing value through multiple layers (also in this scenario it actually opened room for bug in context of AWS Lambda package - will patch with next PR)

PR #674, opened 2023-04-20
Support python v3.10

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-900/support-aws-lambda-python-v310-runtime
* I've tried to run the unit tests with python3.10 and found a single piece of code that needed a change because they've changed the signature of `traceback.format_exception` (first argument is ignored since python 3.5+)

### Testing done
Unit and integration tested against python 3.10


selcukcihan:

medikoo (approved):
👍

PR #677, opened 2023-04-21
Fix request response decoding

selcukcihan:
### Description
* Further manual testing revealed a bug in how we report request/response.
* The body object can be a file-like object, in which case it needs to be handled differently.

### Testing done
* Added a unit test that reproduces the issue and fixed it.
* Integration tested.


selcukcihan:

medikoo (approved):
Thank you 👍

PR #678, opened 2023-04-21
SDK: Optimize and improve error handling in structured logging case

medikoo:
- Ensure eventual unexpected parsing errors are surfaced as SDK internal errors (so we're aware of eventual bugs in the parsing algorithm)
- Improve speed of target structured log detection

Confirmed additionally against integration tests that no unwanted changes were introduced

### DIFF
--- a/node/packages/sdk/lib/structured-log-to-event.js
+++ b/node/packages/sdk/lib/structured-log-to-event.js
@@ -117,7 +117,9 @@ const handleWarningLog = (logLineParsed) => {
 
 module.exports.attemptParseStructuredLogAndCapture = (logLine) => {
   try {
-    const logLineParsed = JSON.parse(logLine.toString());
+    if (typeof logLine !== 'string') return;
+    if (logLine[0] !== '{') return;
+    const logLineParsed = JSON.parse(logLine);
     if ('level' in logLineParsed) {
       const logLevel = parseLogLevel(logLineParsed.level);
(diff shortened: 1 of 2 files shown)


node/packages/sdk/lib/structured-log-to-event.js
```diff
@@ -118,17 +119,23 @@ const handleWarningLog = (logLineParsed) => {
 
 module.exports.attemptParseStructuredLogAndCapture = (logLine) => {
   try {
-    const logLineParsed = JSON.parse(logLine.toString());
-    if ('level' in logLineParsed) {
-      debuglog('LOG LEVEL', logLineParsed.level);
-      const logLevel = parseLogLevel(logLineParsed.level);
-      if (logLevel === 'ERROR') {
+    if (typeof logLine !== 'string') return;
```
Mmarzex, line 122: Not sure that we should limit this to just string types. The stdout/stderr write method can be a string or UInt8Array, seems like we should support both.

Mmarzex (changes requested):

node/packages/sdk/lib/structured-log-to-event.js
medikoo: Yes, but we know that tools we want to detect are not using those types to write structured logs

It's about speeding up the detection process and aborting fast if we know it's not a chunk of our interest

Mmarzex (approved):

PR #679, opened 2023-04-21
AWS Lambda SDK: Remove no longer supported option

medikoo:
Follow up to #670

### DIFF
--- a/node/packages/aws-lambda-sdk/trace-spans/aws-lambda.js
+++ b/node/packages/aws-lambda-sdk/trace-spans/aws-lambda.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const TraceSpan = require('@serverless/sdk/lib/trace-span');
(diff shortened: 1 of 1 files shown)


Danwakeem (approved):
:shipit:

Danwakeem:
@medikoo You mentioned there was a potential bug with this, what was the bug?

medikoo:
> @medikoo You mentioned there was a potential bug with this, what was the bug?

@Danwakeem I thought that importing at this point from SDK can lead to situation where report might be written to other (not effective) `serverlessSdk` instance. Yet as I looked more deeply it wasn't the case here (hence I marked this as _refactor_ not _bug_)

It would be the case if we would import module like that in context of instrument folder (see README: https://github.com/serverless/console/tree/main/node/packages/aws-lambda-sdk/instrument )as it's bundled separately.

PR #683, opened 2023-04-25
Fix transient dependency version and add more tests for aiohttp

selcukcihan:
### Description
While working on https://linear.app/serverless/issue/SC-905/python-sdk-incorrect-span-hierarchy I've noticed that the integration test `aiohttp_requester` was failing. I've noticed it fails against main as well. I could not trace it to a particular change we had. So I started looking at the dependencies.

I've traced the issue to a release in one of dependencies of `aiohttp`, named `yarl`. That repo has an issue that look similar to the problem we are facing: https://github.com/aio-libs/yarl/issues/854

The problem is, `url` reported in the tracing callbacks https://docs.aiohttp.org/en/stable/tracing_reference.html lack the query string parameter.

This PR adds a unit test that reproduces this problem ~, fix is yet to be provided.~ fix is provided as well.

For reference, if we install `yarl==1.8.2` the tests pass (previous version). If we use the latest version `yarl==1.9.1` the tests fail https://pypi.org/project/yarl/#history

### Testing done
Unit/integration tested

### DIFF
--- a/python/packages/aws-lambda-sdk/pyproject.toml
+++ b/python/packages/aws-lambda-sdk/pyproject.toml
@@ -18,6 +18,7 @@ dependencies = [
     "serverless-sdk-schema~=0.2.1",
     "typing-extensions~=4.5", # included in Python 3.8 - 3.11
     "wrapt~=1.15.0",
+    "yarl~=1.8.0",
 ]
 [project.optional-dependencies]
 tests = [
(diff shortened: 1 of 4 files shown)


selcukcihan:
I've described the exact problem on yarl repo https://github.com/aio-libs/yarl/issues/854#issuecomment-1521670598

selcukcihan:

medikoo (approved):
Great catch 👍

PR #685, opened 2023-04-26
SDK: Full resolution of router paths when configured with nested express routers

medikoo:
Addresses use case reported by one of our customers:

```
is it possible for the path column in the list view to show the full path that shows in the detail view?
we started seeing this once we switched to using individual routers per resource, e.g.

const postsRouter = require('./posts/router')
app.use('/v1/posts', postsRouter)
```

Additionally configured more extensive test of express instrumentation in context of `sdk` package (so far this was mainly tested in context of `aws-lambda-sdk`)


Danwakeem (approved):
👍

PR #688, opened 2023-04-26
Fix docs to ensure previous runs do not interfere

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-952/python-cannot-run-integration-tests
* Before running integration tests, we need to build the SDK and its dependencies in a target folder. If this folder is not empty and contains any of the dependencies, those dependencies might be of previous versions and they won't be overridden by pip, which causes stale packages to end up in the layer that is tested.
* Added section to make sure the target directory is removed before building the packages.

### Testing done
Tested by following the directions to run integration tests.


selcukcihan:
@medikoo ready for review please. Some of the tests will still fail (aiohttp) due to the aiohttp issue I raised yesterday, if you install the SDK from pypi instead of building from local. I'll create a release for that fix separately.

medikoo (commented):
Thanks @selcukcihan 

Can we add this instruction also to: https://github.com/serverless/console/tree/main/node/test/benchmark#12-python

Other question:
Do we really need to activate venv? It feels pretty invasive (changes shell prompt etc.)

What does venv usage guarantee ? (Node.js and Go seem to run ok without similar step)

selcukcihan:
> Thanks @selcukcihan
> 
> Can we add this instruction also to: https://github.com/serverless/console/tree/main/node/test/benchmark#12-python
> 
> Other question: Do we really need to activate venv? It feels pretty invasive (changes shell prompt etc.)
> 
> What does venv usage guarantee ? (Node.js and Go seem to run ok without similar step)

`venv` guarantees a fresh sandbox environment, without it your Python environment's current state may interfere with the tests which might make it hard to debug.

medikoo (approved):
👍

PR #692, opened 2023-04-26
feat: Support Custom Tags & Custom Fingerprints to events

Mmarzex:


go/packages/slslambda/slslambda.go
```diff
@@ -47,7 +48,16 @@ func CaptureError(ctx context.Context, err error) {
 		debugLog("capture error:", ctxErr)
 		return
 	}
-	span.captureError(err)
+	span.captureError(err, nil)
+}
+
+func CaptureErrorWithOptions(ctx context.Context, err error, options EventOptions) {
```
Danwakeem, line 49: Is it typical to have two functions where one says `WithOptions` in go?

Danwakeem (dismissed):
Looks good to me 👍

go/packages/slslambda/slslambda.go
Mmarzex: I see it pretty commonly, go doesn't support method overloading or default parameters, so if we wanted to add options to the existing `CaptureError` method it would be a breaking change. This way we don't break existing calls to that method.

go/packages/slslambda/event.go
```diff
@@ -16,19 +18,37 @@ const (
...
 		ToProto(traceID, parentSpanID []byte) (*instrumentationv1.Event, error)
 	}
 )
 
+func marshalCustomTags(options *EventOptions) *string {
+	if options.CustomTags != nil && len(options.CustomTags) > 0 {
+		b, err := json.Marshal(options.CustomTags)
+		if err != nil {
+			return nil
+		}
```
czubocha: I think we should report possible error here somehow

go/packages/slslambda/caughterror.go
```diff
@@ -1,19 +1,21 @@
...
 }
 
-func newCaughtError(err error, timestamp time.Time) *caughtError {
+func newCaughtError(err error, timestamp time.Time, options *EventOptions) *caughtError {
 	return &caughtError{errorEvent: errorEvent{
-		timestamp: timestamp,
-		error:     err,
+		timestamp:    timestamp,
+		error:        err,
+		EventOptions: options,
```
czubocha, line 18: I think we can keep this field private (not exported)

go/packages/slslambda/event.go
```diff
@@ -16,19 +18,37 @@ const (
...
+		*EventOptions
 	}
 	protoEvent interface {
 		ToProto(traceID, parentSpanID []byte) (*instrumentationv1.Event, error)
 	}
 )
 
+func marshalCustomTags(options *EventOptions) *string {
+	if options.CustomTags != nil && len(options.CustomTags) > 0 {
+		b, err := json.Marshal(options.CustomTags)
```
czubocha:
```suggestion
	if options.CustomTags == nil || len(options.CustomTags) = 0 {
	  return nil
	}
		b, err := json.Marshal(options.CustomTags)
```

It's better to return early to reduce nesting https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_return_early_rather_than_nesting_deeply

go/packages/slslambda/slslambda.go
czubocha, line 49: Can now we change `CaptureError(ctx context.Context, err error)` function implementation to
```
CaptureErrorWithOptions(ctx, nil)
```
to not avoid duplicated code? And the same with `CaptureWarning`?

go/packages/slslambda/event.go
```diff
@@ -16,19 +18,37 @@ const (
 )
 
 type (
+	EventOptions struct {
+		CustomFingerprint *string
```
czubocha, line 22: I think it would be more convenient for users to provide just `string` instead of a pointer. We can make a pointer under the hood if we need it.

czubocha (changes requested):

go/packages/slslambda/slslambda.go
Mmarzex: Good call.

go/packages/slslambda/event.go
Mmarzex: I think we could do that if we had an interface for internal sdk errors in the go sdk, but we don't yet. I can see if it is feasible  without too much of a rewrite.

go/packages/slslambda/event.go
Mmarzex: I'm not sure I agree with this. This is an optional field that users do not need to set. At the same time, a custom fingerprint of `""` would technically still be supported, so we would either need to change the event spec to say that isn't a valid fingerprint or have the go SDK behave differently from node and python.

czubocha: Yeah, agree. If `""` is a valid fingerprint then the pointer is a must.

go/packages/slslambda/event.go
czubocha: I mean, for now, even a simple log is ok. I'd like to avoid ignoring errors completely.

Mmarzex: Let's just do a debug log then. I don't want us creating noise in users logs.

czubocha (approved):
👌 thanks!

PR #693, opened 2023-04-26
AWS Lambda SDK Improve sampling

medikoo:
- Do not prevent sampling on warnings
- Resign from reiterating over time frame window (in some cases, requests could be handled in near 200ms+ time frame, and in such cases, sampling efficiency would be very low)

At the same time keep rules where we:
- Do not sample first invocation
- Ensure to not sample two API requests in a row

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -11,7 +11,7 @@ const sendTelemetry = require('./lib/send-telemetry');
 const flushSpans = require('./lib/auto-send-spans').flush;
 const filterCapturedEvent = require('./lib/filter-captured-event');
 const invocationContextAccessor = require('./lib/invocation-context-accessor');
-const isApiEvent = require('./lib/is-api-event');
+const resolveIsApiEvent = require('./lib/is-api-event');
 const pkgJson = require('../package');
 
 const serverlessSdk = require('./lib/sdk');
@@ -24,7 +24,7 @@ const coreTraceSpanNames = new Set([
   'aws.lambda.initialization',
   'aws.lambda.invocation',
 ]);
-const alertEventNames = new Set(['telemetry.error.generated.v1', 'telemetry.warning.generated.v1']);
+const alertEventNames = new Set(['telemetry.error.generated.v1']);
 
 const capturedEvents = [];
 serverlessSdk._eventEmitter.on('captured-event', (capturedEvent) =>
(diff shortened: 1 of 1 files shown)


Mmarzex (approved):

node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -113,21 +115,23 @@ const reportTrace = ({ isErrorOutcome }) => {
...
-      ++requestCounter;
-      // In case of API backed function sample out 3rd and following invocations,
-      // that happen in a frame of a second
-      if (isApiEvent()) return requestCounter > 2;
-      // In other cases do not sample 1st invocation, and sample out 90% of following invocations
-      if (requestCounter === 1) return false;
-      return Math.random() > 0.1;
+      // Do not sample first invocation, otherwise set sampling rate at 10%
+      // (for API we apply correction as requests are passed through in pairs)
+      if (!isFirstInvocation && Math.random() > (isApiEvent ? 0.05 : 0.1)) return true;
```
Danwakeem, line 127: The OPTIONS + METHOD request sequence would only happen in a `/{proxy+}` API 🤔 I am curious how this might perform in a non proxy API 🤔 Do you think it would be possible to only apply this rule if we see the path come through with proxy set on the path?

Danwakeem (commented):
One question about the OPTIONS + Method request sequence

node/packages/aws-lambda-sdk/instrument/index.js
medikoo: > The OPTIONS + METHOD request sequence would only happen in a /{proxy+} API 🤔

Why exactly? API detection is bound purely to event format, it doesn't have to do anything with routing and how it's set up in the application

Danwakeem: For example, if I have a function configured like this 👇 

```
api:
  handler: index.handler
  events:
    - httpApi:
        path: /
        method: get
```

and I send (Or the browser sends) an OPTIONS request to the `/` endpoint my lambda function will not be invoked until the GET request comes though. So even though it would be considered an API via the API detection method my function would never see the OPTIONS + Method request sequence.

Alternatively if I have a function configured like this 👇 

```
api2:
  handler: index.handler
  events:
    - httpApi:
        path: /{proxy+} # or "*"
        method: "*"
```

Then my function would see the OPTIONS + Method requests sequence.

So I guess maybe my original question is worded kind of weird but that is because typically people use proxy routes with the method wildcarded. 

I probably should have said assuming the OPTIONS + Method request sequence would only be valid if their function is configured to include the OPTIONS method. I am curious how this will work for API endpoints that are not configured to receive the OPTIONS request.

medikoo: Yes it can be the case. Best if you raise this scenario in a meeting. We're still discussing what approach is best.

This PR doesn't introduce the idea of sending two consecutive API requests (which is already in main) it just removes the idea of 1 second time frame, which we agreed may not work well.

We will definitely introduce further improvements

Danwakeem: Ahh okay I didn't realize this was the intention of the `requestCounter` previously 🤔 Okay I can at least find a relevant linear ticket to post this in 👍

Danwakeem (approved):
This should be alright 👍 I can post the info about the OPTIONS request in linear

PR #694, opened 2023-04-26
fix: Do not send SDK errors for unsupported log levels

Mmarzex:
When I pushed out the newest SDK version to dev I noticed quite a few more SDK Caught Errors coming in and it appears to be sending an error for every info log.

### DIFF
--- a/node/packages/sdk/lib/structured-log-to-event.js
+++ b/node/packages/sdk/lib/structured-log-to-event.js
@@ -132,6 +132,11 @@ module.exports.attemptParseStructuredLogAndCapture = (logLine) => {
       case 'WARN':
         handleWarningLog(logLineParsed);
         return;
+      case 'INFO':
+      case null:
+        // We don't capture info logs but we don't want to throw an error either.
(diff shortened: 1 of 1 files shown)


node/packages/sdk/lib/structured-log-to-event.js
```diff
@@ -132,6 +132,11 @@ module.exports.attemptParseStructuredLogAndCapture = (logLine) => {
       case 'WARN':
         handleWarningLog(logLineParsed);
         return;
+      case 'INFO':
```
medikoo: I think there's no way we can have `INFO` returned here (?)

medikoo (commented):
Thanks @Mmarzex actually it's a fix to bug I've introduced. It was supposed to have `case null` I  forgot to add it.

node/packages/sdk/lib/structured-log-to-event.js
Mmarzex: You're right. I thought we were returning `info` or null.

medikoo (approved):
Looks great 👍

PR #697, opened 2023-04-27
[Python SDK] Fix DynamoDB expression tags

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-951/python-sdk-error-generated-by-user
* There's a particular usage pattern where the call to DynamoDB succeeds, but the instrumentation part fails generating an internal SDK error, which does not break the lambda but the trace ends up with an extra event representing the error.
* The fix is to ensure all the expressions within dynamodb request parameters are stored as tags with string values
* Added tests using `pynamodb` to access DynamoDB
* Removed moot internal/external variant for `aws_sdk` test

### Testing done
* Reproduced in unit/integration tests
* Unit/integration/performance tested


selcukcihan:

medikoo (approved):
Looks great 👍

@selcukcihan can you follow up with release? It'll be great to roll out this today

PR #710, opened 2023-05-02
SDK: Ensure to resolve stack trace only for exposed logs

medikoo:
/cc @eahefnawy 

Addresses internally reported issue, where it was observed that for instrumented function duration is significantly worse.

It was caused by two things:
- Source maps being enabled  (set via `NODE_OPTIONS=--enable-source-maps`)
- Unconditionally resolved stack trace for debug log (which was to be written at start of each monitored HTTP request)

When source maps are enabled, resolution of stack traces is significantly more expensive, and as there was an attempt to resolve a stack trace with each monitored HTTP request, it put performance of the function down.

This patch ensures that we do not resolve the stack trace if debug mode is not explicitly turned on for the SDK, having that issue is not longer observed (unless `SLS_SDK_DEBUG` is set for the lambda)

### DIFF
--- a/node/packages/sdk/lib/instrumentation/http.js
+++ b/node/packages/sdk/lib/instrumentation/http.js
@@ -144,11 +144,15 @@ const install = (protocol, httpModule) => {
 
   const request = function request(...args) {
     const startTime = process.hrtime.bigint();
-    serverlessSdk._debugLog(
-      'HTTP request',
-      shouldIgnoreFollowingRequest,
-      !shouldIgnoreFollowingRequest && new Error().stack
-    );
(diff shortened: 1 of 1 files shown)


Danwakeem (approved):
Great find! Is it possible the same could happen for other instrumentations beyond just http?

medikoo:
> Great find! Is it possible the same could happen for other instrumentations beyond just HTTP?

@Danwakeem it was specific to this debug log, which now I've wrapped in this additional _if_ clause. There are no other debug logs in which we resolve the stack trace, so it affected only HTTP instrumentation

PR #718, opened 2023-05-04
[Python SDK] Remove importlib_metadata runtime dependency

selcukcihan:
### Description
* Hard-code the version number instead of relying on importlib_metadata to figure it out

### Testing done
Unit tested

### DIFF
--- a/python/packages/sdk/sls_sdk/base.py
+++ b/python/packages/sdk/sls_sdk/base.py
@@ -3,14 +3,13 @@ from __future__ import annotations
 from datetime import datetime
 from typing import List, Union
 
-from importlib_metadata import version
 from typing_extensions import Final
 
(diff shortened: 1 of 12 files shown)


selcukcihan:

medikoo (approved):
👍

python/packages/sdk/sls_sdk/base.py
```diff
@@ -3,15 +3,14 @@
...
 
-from importlib_metadata import version
 from typing_extensions import Final
 
 
 SLS_ORG_ID: Final[str] = "SLS_ORG_ID"
 
 # module metadata
 __name__ = "serverless-sdk"
-__version__: Final[str] = version(__name__)
```
medikoo, line 14: Isn't that change problematic maintenance wise? Will we now have to bump this version manually every time? We definitely should not do that

PR #719, opened 2023-05-04
[Python Lambda SDK] Remove importlib dependency

selcukcihan:
### Description
* Similar PR https://github.com/serverless/console/pull/718
* Use VERSION file instead of relying on importlib_metadata to figure out the version.

### Testing done
Unit tested


selcukcihan:

medikoo (commented):
@selcukcihan sorry, can you rebase against main (I needed to repush there last commit to trigger github action, and that put this PR out of sync)

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/__init__.py
```diff
@@ -28,7 +27,7 @@
 
 # module metadata
 __name__: Final[str] = "serverless-aws-lambda-sdk"
-__version__: Final[str] = version(__name__)
+__version__: Final[str] = "0.1.15"
```
medikoo: Isn't that change problematic maintenance wise? Will we now have to bump this version manually every time? We definitely should not do that

selcukcihan: Fair point, we'll need to do that correct. I thought we were already doing it with pyproject.toml file anyway so why not do it here as well.

I think a middle ground would be to still do it dynamically but do a lazy initialization, what do you think?

selcukcihan: Removing the import of `importlib_metadata` saves us 50ms that's for sure. We can choose to move it from INIT time to request time by lazy loading it, but the latency will be there. If we do it manually like in this PR, we get rid of the latency with the drawback of bumping the version manually for each release.

selcukcihan: I'm now evaluating the options listed at https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

medikoo: > Fair point, we'll need to do that correct. I thought we were already doing it with pyproject.toml file anyway so why not do it here as well.

it's where we control package version, but it's easy to forget to update the second place where we're doing that.

We had same issue in Node.js SDK, and when release was made by other developers obviously version in other file was not updated. It's very easy to get this out of sync

medikoo: Why exactly to simply retrieve the current version of the package, we need to rely on dependency that eats 50ms of the initializatio time?

Is there really no other way?

selcukcihan: There are some other ways as linked above, I'm thinking of going with the "VERSION" file path. We'll define the version in a separate file and the project.toml has a way of reading it from that file, in the code we also read it from the file:

```
[tool.setuptools.dynamic]
version = {file = "./serverless_aws_lambda_sdk/VERSION"}
```

and in the code:

```
with open(Path(__file__).parent / "VERSION") as version_file:
    __version__ = version_file.read().strip()
```

What do you think? I think this is good such that we only have a single source of truth as well as no latency of importing a huge library just for the version reading.

selcukcihan: I've amended the PR with the `VERSION` file implementation, please take a look 🙏

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/VERSION
```diff
@@ -0,0 +1 @@
+0.1.15
```
medikoo: Can we ensure EOL at EOF?

python/packages/aws-lambda-sdk/pyproject.toml
```diff
@@ -33,6 +35,7 @@ tests = [
     "black>=22.12",
     "boto3>=1.16.112",
     "flask>=2.2.3",
+    "importlib_metadata>=5.2", # included in Python >=3.8
```
medikoo, line 38: Why do we need to list it as optional dependency?

medikoo (commented):
Thank you @selcukcihan, that looks as clever idea.

python/packages/aws-lambda-sdk/pyproject.toml
selcukcihan: The test still uses it to infer the version, it will only be installed for the tests and not included in the package that get's distributed.

python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/VERSION
selcukcihan: 👍

medikoo (commented):
Looks great 👍  Still, I've just realized I think we need to update the CI version change detection, is that correct?

selcukcihan:
> Looks great 👍 Still, I've just realized I think we need to update the CI version change detection, is that correct?

Disabled for this commit

medikoo (approved):
👍

medikoo:
@selcukcihan it appears integration tests failed after we've merged that in: https://github.com/serverless/console/actions/runs/4893984397/jobs/8737620240#step:7:22445

selcukcihan:
> @selcukcihan it appears integration tests failed after we've merged that in: https://github.com/serverless/console/actions/runs/4893984397/jobs/8737620240#step:7:22445

I'm taking a look at it

PR #720, opened 2023-05-04
Remove blinker dependency

selcukcihan:
### Description
* Remove blinker usage and use a simple pub/sub logic
* Blinker dependency is still needed in case of flask instrumentation, so it's still packaged.

### Testing done
Unit tested


selcukcihan:

medikoo (commented):
If we improved the initialization time, it'll be good to narrow the initialization duration threshold in performance test

selcukcihan:
> If we improved the initialization time, it'll be good to narrow the initialization duration threshold in performance test

Yes, we did improve init time, I've reduced it to 400ms now.

medikoo (approved):
👍

PR #721, opened 2023-05-05
[Python base SDK] Switch to using version file

selcukcihan:
### Description
* Similar PR https://github.com/serverless/console/pull/719
* Use VERSION file instead of relying on importlib_metadata to figure out the version.

### Testing done
Unit tested


selcukcihan:

medikoo (commented):
Looks great, but what about CI detection?

Should we do it, in the next PR so it doesn't accidentally try to release new version?

selcukcihan:
> Looks great, but what about CI detection?
> 
> Should we do it, in the next PR so it doesn't accidentally try to release new version?

Good point, checking it

selcukcihan:
@medikoo I've disabled auto publish of new tag, we can enable after this is merged.

medikoo (approved):
👍

PR #726, opened 2023-05-08
Handle import errors

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-977/user-thetie-seeing-error-in-production-mode-with-python-sdk
* Handle import errors from the SDK (including 3rd party dependencies) by falling back to just executing the customer's handler function.

### Testing done
* Reproduced in a unit test
* Unit/integration tested

### DIFF
--- a/python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/internal_extension/wrapper.py
+++ b/python/packages/aws-lambda-sdk/serverless_aws_lambda_sdk/internal_extension/wrapper.py
@@ -15,18 +15,31 @@ if environ.get("_ORIGIN_HANDLER") is None:
 environ["_HANDLER"] = environ.get("_ORIGIN_HANDLER")
 del environ["_ORIGIN_HANDLER"]
 
+_IMPORT_FAILED = False
 try:
-    from serverless_aws_lambda_sdk import serverlessSdk
-except ModuleNotFoundError:
-    from .. import serverlessSdk
+    try:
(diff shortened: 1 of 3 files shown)


selcukcihan:
@medikoo ready for review 🙏

medikoo (changes requested):
@selcukcihan There are conflicts, it looks it needs rebase with main

medikoo:
@selcukcihan still, let's postpone this work until we agree on final solution in Linear. I'll convert it to draft for a meanitime

selcukcihan:
> @selcukcihan still, let's postpone this work until we agree on final solution in Linear. I'll convert it to draft for a meanitime

Thanks @medikoo I think this is still relevant and a good practice to implement, after we solve the problem of managing conflicting versions we should bake this in as well.

medikoo:
> Thanks @medikoo I think this is still relevant and a good practice to implement, after we solve the problem of managing conflicting versions we should bake this in as well.

But if we ensure that we import the module we packaged, then there's no chance that import may crash, right?

selcukcihan:
> > Thanks @medikoo I think this is still relevant and a good practice to implement, after we solve the problem of managing conflicting versions we should bake this in as well.
> 
> But if we ensure that we import the module we packaged, then there's no chance that import may crash, right?

Yes 💯

medikoo:
Closing as I believe with taken approach is not longer needed

PR #728, opened 2023-05-08
SDK: Fix detection of not observed response body

medikoo:
Internally we've observed no closed span detection in case, firing whenever Datadog instrumentation is included as well.

Datadog instrumentation initializes the request, waits for the response, but doesn't observe response body.

HTTP instrumentation had such a scenario taken into account from the start, the intention was to auto-close span when we parent span is closed, and we confirmed that the response was already obtained, but no response body reading is initialized.

Still, when testing, I found that detection on whether the response body reading is initialized was not configured properly (it was assumed that `responseReadableState.flowing` will equal `false` in such case, while as tested, it equals `null` instead) this patch fixes that.


Danwakeem (approved):
✅

PR #730, opened 2023-05-08
Use native http module instead of aiohttp

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-606/python-sdk-very-large-extension-layer-size#comment-6c3fcd36
* Use `http.client` package instead of `aiohttp` for dev-mode telemetry requests
* On my dev environment:
  * Unzipped layer size is down to 4.1MB from 7.5MB
  * Zipped layer size is down to 1.1MB from 2.1MB

### Testing done
Unit/integration/performance tested

### DIFF
--- a/python/packages/aws-lambda-sdk/pyproject.toml
+++ b/python/packages/aws-lambda-sdk/pyproject.toml
@@ -23,7 +23,6 @@ dynamic = [
     "version",
 ]
 dependencies = [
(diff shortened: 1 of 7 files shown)


selcukcihan:

medikoo (approved):
Looks great 👍

PR #731, opened 2023-05-09
[Python base SDK] Remove wrapt dependency

selcukcihan:
### Description
* Related issues https://linear.app/serverless/issue/SC-606/python-sdk-very-large-extension-layer-size & https://linear.app/serverless/issue/SC-593/python-sdk-big-latency-overhead
* We were using wrapt when instrumenting 3rd party libraries, we only use a small functionality which is easy to implement and save some space and performance by not importing wrapt.

### Testing done
Unit/integration tested.


selcukcihan:
@medikoo ready for review 🙏

medikoo (dismissed):
Great call 👍

selcukcihan:
@medikoo I've rebased this as well just in case.

medikoo (approved):
Great call 👍

PR #738, opened 2023-05-10
AWS Lambda SDK: Improve size of written telemetry log by gzipping it

medikoo:
An internal investigation showed that gzipped trace payloads take 40-60% less of space

@czubocha @Mmarzex I'll hold off merging it until I have confirmation that we support that on the backend side

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -2,6 +2,7 @@
 
 'use strict';
 
+const zlib = require('zlib');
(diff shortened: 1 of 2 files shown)


Danwakeem (approved):
Zip ⚡️

czubocha (approved):

selcukcihan:
@medikoo Wanted to let you know just in case, I'm receiving email notifications on these issues but I'm not taking any action. Please let me know if it needs to be addressed on the Python SDK as well.

medikoo:
@selcukcihan thanks for heads up. I remember about Python, and I'm about to create an issue in Linear for that 👍

PR #740, opened 2023-05-10
Remove yarl dependency

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-606/python-sdk-very-large-extension-layer-size#comment-18e1b26b
* Remove dependency on yarl, we do not need yarl directly, it's a transitive dependency of aiohttp which we don't depend on anymore

### Testing done
Integration/unit tested.


selcukcihan:
@medikoo ready for review 🙏

medikoo (approved):
👍

PR #742, opened 2023-05-10
[Python Lambda SDK] Trim down protobuf when bundling the layer

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-606/python-sdk-very-large-extension-layer-size#comment-18e1b26b
* Remove the unused C backend implementation from protobuf when bundling the layer. More information on the backends available in protobuf => https://github.com/protocolbuffers/protobuf/tree/main/python#implementation-backends
* This cuts down the size of the zipped layer archive around 200KB, resulting in a layer of size 650KB when bundled from my local machine

### Testing done
Integration tested to make sure removal of the binary does not break the SDK.

### DIFF
--- a/.github/workflows/python-integrate.yaml
+++ b/.github/workflows/python-integrate.yaml
@@ -207,6 +207,7 @@ jobs:
         run: |
           cd python/packages/aws-lambda-sdk
           python3 -m pip install . --target=dist
+          rm -rf dist/google/_upb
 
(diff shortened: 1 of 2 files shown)


selcukcihan:
@medikoo ready for review 🙏

medikoo (approved):
Great find 👍

PR #745, opened 2023-05-11
AWS Lambda SDK: Release v0.15.0

medikoo:
I've also bumped a bit threshold in initialization performance test, as we've got CI fail: https://github.com/serverless/console/actions/runs/4950056090 (even though nothing really changed) - this should give bigger room to avoid false warnings


Danwakeem (approved):
:shipit:

PR #747, opened 2023-05-12
Handle datetime json serialization

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-1011/python-sdk-customer-receiving-warning-detected-not-serializable-value
* When json serializing, certain types like datetime are not json serializable. This ensures that the string representation of such types are used. For the case of datetime, string representation is ISO 8601 format, `YYYY-MM-DDTHH:MM:SS.ffffff`

### Testing done
* Reproduced in a unit test
* Integration tested


selcukcihan:

medikoo (approved):
👍

PR #749, opened 2023-05-12
feat: Added a new transport payload for dev mode

Danwakeem:
## Description
I am adding a new dev mode transport payload so that we can send a batch of all types of dev mode activity in a single payload.

> Note: I didn't modify the original `DevModePayload` because this new transport payload will be use for a different purpose than the original dev mode payload was designed for. I will deprecate the `DevModePayload` once this work is complete and deployed to customer lambda functions.


medikoo (approved):
👍

PR #750, opened 2023-05-12
AWS Lambda SDK: Do not write dev mode specific events in trace payload

medikoo:
We've agreed to not expose trace payload events which are specific to dev mode functionality


Danwakeem (dismissed):
👍

PR #751, opened 2023-05-12
AWS Lambda SDK: Improve stream body reporting

medikoo:
When browsing internally, reported warnings and errors. I observed the `Detected not serializable value in AWS SDK request` warning, which pointed out that the Node.js stream value was passed with the `Body` property to the payload.

This patch ensures that we gently handle such case


Danwakeem (approved):
Oh nice 😎 Sorry for the delay on this one

PR #752, opened 2023-05-12
feat!: Group dev mode activity

Danwakeem:
## Description
I added three enhancements here.

1. traces, req/res, and logs will no longer make individual API calls and will share a single request depending on what data we have available
2. We will wait either 500ms or till the runtimeDone event is seen (whichever is first) before forwarding data to dev mode backend
3. Payloads will be gzipped

This will cut down on the number of requests a single lambda function can make when it is connected to dev mode and should prevent issues like we had seen yesterday.

### DIFF
--- a/go/packages/dev-mode/agent/forwarder.go
+++ b/go/packages/dev-mode/agent/forwarder.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"bytes"
+	"compress/gzip"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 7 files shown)


medikoo (commented):
@Danwakeem it looks as CI failed

Danwakeem:
@medikoo I think there might be a test flake issue 🤔 it looks like the mock lambda server was having trouble standing up for some reason. I will try and see if I can figure that one out once I am done with this work 👍

czubocha (commented):
looks good!

czubocha (approved):

medikoo:
@Danwakeem integration tests failed after this merge

Danwakeem:
Ahh okay @medikoo I know they needs updates and I forgot if they ran after pushing to main or when you attempt a deploy. I am planning on updating them in the next PR 👍

PR #755, opened 2023-05-15
feat: Include dev mode transport in sdk schema

Danwakeem:
## Description
Added new dev mode transport to the sdk schema so that I can update the integration tests 😎


medikoo (approved):
👍

PR #756, opened 2023-05-15
feat: Move to function URL rather than API Gateway URL

Danwakeem:
## Description
Moved to function URLs so we reduce the number of API Gateway invocations. I also updated integration tests 👍


node/test/go/dev-mode/test/integration/index.test.js
```diff
@@ -49,8 +48,20 @@ describe('Integration', function () {
       '4s-logger',
       {
         variants: new Map([
-          ['v16', { configuration: { Runtime: 'nodejs16.x', Timeout: 5 } }],
-          ['v18', { configuration: { Runtime: 'nodejs18.x', Timeout: 5 } }],
+          [
+            'v16',
+            {
+              ignoreMultipleInvocations: true,
```
medikoo: Why exactly this was added? It doesn't feel right

czubocha: @Danwakeem I had an issue with multiple invocations during tests in the past. The reason for that was the AWS account where I run the tests was integrated with Console and Console was changing the lambda config during the test, resulting in the creation of a new lambda container. If you run into a "multiple invocations" issue make sure you run the tests on an account not integrated with Console 👍

Danwakeem: 🤔 Ahh okay maybe I was running it in the wrong account then. Let me try a different account and see if I no longer need this.

Danwakeem: Thanks @czubocha that was the issue 👍

czubocha (approved):

PR #757, opened 2023-05-16
chore: Release Dev Mode v0.7.0

Danwakeem:
## Description
Adding better batching and function url support to the external dev mode extension 😎


medikoo (dismissed):
👍

Danwakeem:
Sorry @medikoo I was doing some research about the test flake here and I think it is because multiple invocations were overlapping each other. I added a parallel flag of 1 so the tests would run in serial instead and I think that will help with the test flake 🙏

medikoo (approved):
👍

PR #758, opened 2023-05-17
Schema: Introduce `aws.lambda.response_mode` tag

medikoo:
With response streaming, AWS introduced different lambda resolution mode.

This tag will let us track lambdas which work in this mode, it'll also be helpful to internal to recognize more we're in, as it requires adjustments to internal logic in related parts


Danwakeem (approved):
This is a good idea 👍

PR #759, opened 2023-05-17
SDK: Enable filters configuration for HTTP requests

medikoo:
To properly support response streaming, we need to ignore HTTP requests made internally by the AWS runtime (as in case of response streaming AWS runtime sends the data to AWS backend chunk by chunk with regular HTTP requests).

This generic filtering method, will let us configure specific HTTP request filter in context of the AWS Lambda SDK package


Danwakeem (approved):
👍

PR #762, opened 2023-05-17
AWS Lambda SDK: Support response streaming

medikoo:
Fix handling of lambdas which are configured for response streaming, and also recognize response streaming


Danwakeem (approved):
🙌

PR #767, opened 2023-05-18
SDK: Fix AWS Lambda Powertools error log parsing

medikoo:
When browsing internal error reports. I found following:

```
{
  "message": "undefined is not iterable (cannot read property Symbol(Symbol.iterator))",
  "name": "TypeError",
  "stacktrace": "at handleErrorLog (/opt/nodejs/node_modules/@serverless/sdk/index.js:3622:34)\nat module2.exports.attemptParseStructuredLogAndCapture (/opt/nodejs/node_modules/@serverless/sdk/index.js:3662:13)\nat nodeStdout.write (/opt/nodejs/node_modules/@serverless/sdk/index.js:3698:9)\nat Console.log (/var/task/src/triggers/handle-db-change.js:201027:23)\nat Console._write (/var/task/src/triggers/handle-db-change.js:198701:19)\nat doWrite (/var/task/src/triggers/handle-db-change.js:118934:139)\nat writeOrBuffer (/var/task/src/triggers/handle-db-change.js:118925:5)\nat Writable.write (/var/task/src/triggers/handle-db-change.js:118846:11)\nat DerivedLogger.ondata (/var/task/src/triggers/handle-db-change.js:117942:20)\nat DerivedLogger.emit (node:events:513:28)",
  "type": "ERROR_TYPE_CAUGHT_SDK_INTERNAL"
}
```

It appears there's a bug when processing an AWS Lambda Powertools error log which was written without passing the error instance.

This patch ensures it's handled gently

### DIFF
--- a/node/packages/sdk/lib/instrumentation/node-process-stdout-stderr.js
+++ b/node/packages/sdk/lib/instrumentation/node-process-stdout-stderr.js
@@ -19,12 +19,28 @@ module.exports.install = () => {
 
   nodeStdout.write = function (...args) {
     original.stdout.write.apply(this, args);
-    attemptParseStructuredLogAndCapture(args[0]);
+    try {
+      attemptParseStructuredLogAndCapture(args[0]);
+    } catch (error) {
+      process.nextTick(() => {
+        // Prevent crashes swalling by Node.js console
+        // see: https://github.com/nodejs/node/blob/3538e1bcde45eae60c6bdbedeb3f765dbb9714c2/lib/internal/console/constructor.js#L306-L310
+        throw error;
+      });
+    }
   };
 
   nodeStderr.write = function (...args) {
     original.stderr.write.apply(this, args);
-    attemptParseStructuredLogAndCapture(args[0]);
+    try {
+      attemptParseStructuredLogAndCapture(args[0]);
+    } catch (error) {
+      process.nextTick(() => {
+        // Prevent crashes swalling by Node.js console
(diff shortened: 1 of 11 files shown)


node/packages/sdk/lib/instrumentation/node-process-stdout-stderr.js
```diff
@@ -19,12 +19,28 @@ module.exports.install = () => {
...
     original.stdout.write.apply(this, args);
-    attemptParseStructuredLogAndCapture(args[0]);
+    try {
+      attemptParseStructuredLogAndCapture(args[0]);
+    } catch (error) {
+      process.nextTick(() => {
+        // Prevent crashes swalling by Node.js console
+        // see: https://github.com/nodejs/node/blob/3538e1bcde45eae60c6bdbedeb3f765dbb9714c2/lib/internal/console/constructor.js#L306-L310
+        throw error;
+      });
```
Danwakeem, line 29: Just out of curiosity why `process.nextTick` here?

Danwakeem (approved):
👍

node/packages/sdk/lib/instrumentation/node-process-stdout-stderr.js
medikoo: To escape swallowing (Console method that invokes that method will swallow and not expose error, if there's a crash here).
It's important for our tests, as otherwise we won't see if our logic crashes in unexpected places

PR #771, opened 2023-05-19
Fix dev mode connection

selcukcihan:
### Description
1. Related issue https://linear.app/serverless/issue/SC-1062/python-sdk-function-within-vpc-and-dev-mode-enabled-timesout-function#comment-6bac1208
2. Flush dev-mode requests before finalizing the trace to make sure all events are captured in the trace (the integration tests were not failing because of this, the dev-mode extension request errors were not included in the trace)
3. Fix the actual problem with how we use `http.client`, [as stated in the docs](https://docs.python.org/3/library/http.client.html#http.client.HTTPConnection.getresponse)
> Note that you must have read the whole response before you can send a new request to the server.

### Testing done
* Item 2 above made the integration tests fail and surface the issue. Item 3 provided the fix and now the integration tests are passing.
* Unfortunately, I couldn't reproduce the issue in unit tests, the test server is very basic and does not support streaming the response and thus the test still passes even if you don't read the response.


selcukcihan:
@medikoo ready for review 🙏

medikoo (approved):
👍

PR #775, opened 2023-05-22
AWS Lambda SDK: Fix handling of not responding handler

medikoo:
In the Node.js function handler does not have to be resolved (its callback called or returned promise resolved) for the invocation to successfully end.

In case of the drained event loop, AWS runtime will wrap up the invocation without waiting for the explicitly provided result.

In our logs, I've observed failures that describe a situation where we have a trace with two `aws.lambda.invocation` spans and not properly cleared tags after the previous invocation. As I tested its unresolved handler and drained the event loop, which may introduce such situations.

This patch fixes the handling of such scenario and issues a warning for the user, describing that such a situation occurred (cc @skierkowski)

### DIFF
--- a/node/packages/aws-lambda-sdk/instrument/index.js
+++ b/node/packages/aws-lambda-sdk/instrument/index.js
@@ -182,7 +182,7 @@ const clearRootSpan = () => {
   delete awsLambdaSpan.id;
   delete awsLambdaSpan.endTime;
   awsLambdaSpan.tags.reset();
-  awsLambdaSpan.subSpans.clear();
+  awsLambdaSpan._subSpans.clear();
   capturedEvents.length = 0;
   if (objHasOwnProperty.call(serverlessSdk, '_customTags')) serverlessSdk._customTags.clear();
 };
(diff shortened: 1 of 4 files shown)


node/packages/aws-lambda-sdk/instrument/index.js
```diff
@@ -252,6 +252,34 @@ if (!process.env.SLS_UNIT_TEST_RUN) {
   wrapUnhandledErrorListener('unhandledRejection');
 }
 
+const observeNotRespondingHandler = (() => {
+  // When event loop is drained, AWS closes the invocation without waiting for an eventual
+  // callback to be called or promise resolved.
+  // Below logic we detect and handle this scenario
+  const exit = () => {
+    if (typeof global[Symbol.for('aws.lambda.beforeExit')] === 'function') {
```
Danwakeem, line 260: How did you find the `aws.lambda.beforeExit` reference? I didn't see this in the docs anywhere 🤔

Danwakeem (approved):
Interesting detective work here 🕵️

node/packages/aws-lambda-sdk/instrument/index.js
medikoo: I inspected how AWS runtime internally handles that (and for that I've downloaded runtime code to the S3 from AWS Lambda function)

Danwakeem: Oh cool, very clever 👍

PR #776, opened 2023-05-22
[Python SDK] Do not import typing extensions for new versions

selcukcihan:
### Description
* Related issues
  * https://linear.app/serverless/issue/SC-593/python-sdk-big-latency-overhead
  * https://linear.app/serverless/issue/SC-977/user-thetie-seeing-error-in-production-mode-with-python-sdk
* Typing extensions module is only needed for Python v3.7 and below, this change makes sure we only import this module if it's necessary for the given runtime.

### Testing done
* Unit/Integration/Perf tested
* The init time is further reduced around 20ms.


selcukcihan:

medikoo (approved):
👍

PR #784, opened 2023-05-25
Use friendly names in Event Types and remove INTERNAL

skierkowski:
Small updates to the docs to use the friendly event type names (same ones used in the UI) and remove references to the `INTERNAL` event types.

### DIFF
--- a/docs/application-guide/trace-explorer.md
+++ b/docs/application-guide/trace-explorer.md
@@ -45,36 +45,23 @@ on any of these properties.
 
 ## Event Types
 
-- `ERROR_TYPE_UNCAUGHT` - The Lambda function handler had a fatal error and
-caused the invocation to fail.
-- `ERROR_TYPE_CAUGHT_USER` - The Lambda function handler had an error that was
-captured using the SDK. This can include Error objects or strings that were
-captured implicitly (e.g. `console.error`) or explicitly using the provided
-SDK methods.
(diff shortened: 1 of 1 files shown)


Danwakeem (approved):
✅

docs/application-guide/trace-explorer.md
```diff
@@ -45,36 +45,23 @@ on any of these properties.
...
+Lambda PowerTools, Winston), or standard out (e.g. `console.warn`).
+- **SDK Error** (`ERROR_TYPE_CAUGHT_SDK_USER`) - An SDK usage error that was
+reported due to misuse of the SDK. These errors do not cause handler failures,
+but misusage of the SDK may result in partial collection. For example, using the
+`setTag` method with invalid inputs will result in this type of error, and the
+tag will not be set.
+- **SDK Warning** (`WARNING_TYPE_SDK_USER`) - A warning reported by the SDK due
+to user misuse in the handler, but not due to misuse of the SDK. For example, if
+both a callback and Promise resolution is attached this will cause unwanted
+side-effects on the SDK.
```
medikoo, line 64: @skierkowski To fully clarify in both cases we may deal with issues that concern SDK API usage, and those which do not concern SDK API (but e.g. AWS SDK usage or some other mishandling)

SDK API usage issue can be detected in case of errors by error class name `ServerlessSdkError` and in case of warning with a fingerprint prefix (but as we do not have those yet, we didn't yet established prefix)

PR #787, opened 2023-05-30
AWS Lambda SDK: Fix handling of case where we wait for empty event loop

medikoo:
When lambda is resolved via callback, by default AWS holds the resolution until the event loop drains. 

So far we didn't handle this properly (as it was not clear how it could be handled), now that we learned how AWS runtime internally detects the drain of event loop, we can fix the handling of this special case


Danwakeem (approved):
🎸

PR #789, opened 2023-06-02
AWS Lambda SDK: Fix handling of side effect of unexpected express configuration

medikoo:
There are reports of `aws.lambda.http_router.path` being set prior invocation starts.

It signals that express middlewares were run at initialisation phase (not specifically to handle event input). It's unexpected to us, yet we need to handle it gently (ignore it)


Danwakeem (approved):
Huh yeah I was also unaware this was possible. Nice detective work 🕵️‍♂️

PR #793, opened 2023-06-07
SDK: Restrict tag size

medikoo:
Restrict tag name to max `256` characters and tag value to `32766` bytes.

Ensure `error.message` and `error.stacktrace` are truncated to fit the limit


Mmarzex (approved):

Danwakeem (approved):
✅

PR #810, opened 2023-06-16
Fix pyproject-fmt version

selcukcihan:
### Description
Noticed that a recent PR was failing due to pyproject.toml file formatting, since pyproject-fmt released a version that breaks for us. This PR fixes the version to `0.10.0` which is known to be working for our files.

### Testing done
Tested by running `python -m pyproject_fmt --indent 4 pyproject.toml` and verifying it does not suggest changes.

### DIFF
--- a/.github/workflows/python-validate.yml
+++ b/.github/workflows/python-validate.yml
@@ -68,7 +68,7 @@ jobs:
         run: |
           cd python/packages
 
-          python3 -m pip install tox pyproject-fmt
+          python3 -m pip install tox pyproject-fmt==0.10.0
 
           for directory in */; do
             if [ -f "$directory/pyproject.toml" ]; then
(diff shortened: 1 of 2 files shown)


selcukcihan:
@medikoo ready for review 🙏

medikoo (commented):
@selcukcihan wouldn't it be better to upgrade to latest version of fmt, and apply to its formatting?

selcukcihan:
> @selcukcihan wouldn't it be better to upgrade to latest version of fmt, and apply to its formatting?

We can do that, but it seems like they keep changing the order they expect from version to version.

selcukcihan:
I've amended the PR to do that 👍

medikoo (approved):
Thanks @selcukcihan if we see one more change like that, we'd lock against specific version

PR #812, opened 2023-06-17
Draft: [Python base SDK] Absolute imports should be made internal

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-1200/issue-with-some-imports-in-python-sdk-initialization#comment-f0d95f19
* The absolute imports by our layer are arranged to be made internal and not visible to customer code. This will ensure our imports are not overwritten by customer code, for instance if customer has a "secrets" module we can still import the standard library "secrets" module and not customer's module.
* We should not import the sdk itself internally though, otherwise the customer's version would be different and we'd have two instances of the sdk hanging around
* Relative imports are not affected, as they point precisely to our layer code.
* This is a breaking change

### Testing done
* Unit tested
* Integration/performance tested, against a modified serverless-aws-lambda-sdk module (because of breaking changes), to make sure the approach is sound

### Next
* We should release a breaking version of base SDK
* Same fixes to be applied on the serverless-aws-lambda-sdk package
* Unit tests and integration tests on serverless-aws-lambda-sdk should be enhanced to cover the module name clashes such as "secrets" and some more ("typing", "contextlib" etc.) 
* New version of the serverless-aws-lambda-sdk package to be released

### DIFF
--- a/python/packages/sdk/sls_sdk/__init__.py
+++ b/python/packages/sdk/sls_sdk/__init__.py
@@ -1,13 +1,17 @@
 from __future__ import annotations
-import sys
-from os import environ
-from typing import List, Optional
 
-if sys.version_info >= (3, 8):
-    from typing import Final
-else:
-    from typing_extensions import Final
-from types import SimpleNamespace
+from .lib.imports import internally_imported
+
+with internally_imported():
+    import sys
+    from os import environ
+    from typing import List, Optional
+
+    if sys.version_info >= (3, 8):
+        from typing import Final
+    else:
+        from typing_extensions import Final
+    from types import SimpleNamespace
 
 from .base import Nanoseconds, SLS_ORG_ID, __version__, __name__
 from .lib import trace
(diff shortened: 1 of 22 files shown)


selcukcihan:
@medikoo this is also ready for review, but we should first release https://github.com/serverless/console/pull/813 if we want to get the other fix out first. I've marked the PR as draft to leave the decision to you.

If you decide to release the two fixes for the two bugs at once, then this should be merged first and then we should update https://github.com/serverless/console/pull/813 to make it a breaking change and move on.

medikoo (approved):
Thank you @selcukcihan, that looks very promising 👍

medikoo:
@selcukcihan I understand that this is ready to merge (?). Can you also explain what's exactly breaking here?

selcukcihan:
Yes, ready to merge 👍 Breaking change is the way we isolate imports, when the layer sdk is still importing old way, this will cause some modules to be imported twice. Also there is a dropped parameter so the code won't compile either in the layer sdk if it uses this new version of base sdk.

PR #814, opened 2023-06-17
[Python Lambda SDK] Absolute imports should be made internal 

selcukcihan:
This PR builds on https://github.com/serverless/console/pull/812 and should be reviewed after we release the other PR in a breaking change:

1. We should update the pyproject.toml to use the latest version
2. Review the PR and merge
3. Release a new version of the serverless-aws-lambda-sdk package

### Description
* Related issue https://linear.app/serverless/issue/SC-1200/issue-with-some-imports-in-python-sdk-initialization#comment-f0d95f19
* The absolute imports by our layer are arranged to be made internal and not visible to customer code. This will ensure our imports are not overwritten by customer code, for instance if customer has a "secrets" module we can still import the standard library "secrets" module and not customer's module.
* We should not import the sdk itself internally though, otherwise the customer's version would be different and we'd have two instances of the sdk hanging around
* Relative imports are not affected, as they point precisely to our layer code.
* This is a breaking change

### Testing done
* Unit tested
* Integration/performance tested

### DIFF
--- a/node/test/python/aws-lambda-sdk/integration.test.js
+++ b/node/test/python/aws-lambda-sdk/integration.test.js
@@ -1072,6 +1072,11 @@ describe('Python: integration', function () {
   let sdkVersion;
   let beforeTimestamp;
 
+  const shadowedModules = [
+    ['google', 'protobuf'],
+    ['secrets', '__init__'],
+  ];
+
   before(async () => {
     exec(
       `pip install pynamodb==5.5.0 yarl==1.8.2 aiohttp==3.8.4 serverless-wsgi==3.0.2 flask==2.2.3 --target="${fixturesDirname}/test_dependencies"`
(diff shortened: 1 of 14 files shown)


selcukcihan:
@medikoo this PR will need to be taken care of after https://github.com/serverless/console/pull/812

selcukcihan:
@medikoo this is now ready, I've enhanced the integration test to test the scenario where client bundles their own module named "secrets" and our sdk still works fine by importing the system's "secrets" module. I was going to add more to the list in the test but it interferes with other test modules such as flask, aiohttp which uses modules like "sys", "os, "typing", "contextvars" etc. Which is fair, because the customer can't both use aiohttp and have a module named "typing" as it will shadow the system module and aiohttp will break.

medikoo (approved):
Looks great 👍

PR #817, opened 2023-06-20
Python SDK: Temporarily disable response body observation

medikoo:
Due to reported errors we temporarily disable this feature


Danwakeem (approved):
⚒️

PR #822, opened 2023-06-23
AWS Lambda SDK: Fix handling of request truncation

medikoo:
When large request body was truncated, we accidentally were passing `null` in place of request body to request and that is crashing protobuf payload handling.

This patch ensures we pass `undefined` which is ignored by protobuf handler


Danwakeem (approved):
👍

PR #824, opened 2023-06-26
AWS Lambda SDK: Ensure to expose uncaught exceptions

medikoo:
In parts of a logic where we process request and response, eventual uncaught exceptions (which in such case will be correctly handled and muted) are converted to unhandled rejections, which are not handled and exposed as lambda crashes to users.

This patch ensures those eventual exceptions land as uncaught exceptions.

It's discovered with latest report, where lambda was crashing for user in dev mode (source of the issue was already fixed with #822)


Danwakeem (approved):
:shipit:

PR #826, opened 2023-06-28
fix: Execution should halt if api error is detected

Danwakeem:
## Description
I forgot to add a return here so once the error was logs (or not depending on the function configuration) we would try and access `res.StatusCode` causing a seg fault.

## Screenshot
Screenshot 2023-06-28 at 3 28 12 PM


czubocha (approved):

medikoo (dismissed):
@Danwakeem CI fails

Danwakeem:
@medikoo This is some test flakeyness that I am trying to resolve in another branch. I haven't quite tracked down why this happens as I haven't been able to reproduce the issue locally 😢 The most recent run should be good.

Danwakeem:
@medikoo I will go ahead and merge this since @czubocha approved and the tests passed after re-running 👍

PR #830, opened 2023-07-02
[Python Base SDK] Fix response body monitoring

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-1204/python-sdk-users-seeing-errors-since-last-update
* Reenable response body monitoring in dev-mode
* For aiohttp instrumentation, if response body monitoring is enabled, we were consuming the response for reporting but then the customer was unable to consume it again as the stream was finished. This fix "unreads" the response data, so that customer can reread it.

#### What Was Happening?
* Note that this is a very specific case in which the transient dependency `aiobotocore` library accesses the internals of `aiohttp` instead of relying its public `read` method which we were already using. If `aiobotocore` also used `read`, it would successfully read the response many times because that `read` method caches the response body. However, since `aiobotocore` accesses the underlying stream directly, once we read the stream for response body reporting, it can't reread it through the underlying stream.
* The fix rewinds the underlying stream to its initial position by unreading the data.

### Testing done
Unit and integration tested (integration tests for this specific case will be added in a separate PR for the lambda SDK).

### DIFF
--- a/python/packages/sdk/sls_sdk/lib/instrumentation/http.py
+++ b/python/packages/sdk/sls_sdk/lib/instrumentation/http.py
@@ -127,10 +127,10 @@ class NativeAIOHTTPInstrumenter(BaseInstrumenter):
         self._original_init = None
 
     async def _capture_response_body(self, trace_span, response):
-        # TODO: Turned off temporarily, until we have response body observation fixed
-        return
         # response is a aiohttp.ClientResponse object
-        if not self.should_monitor_request_response:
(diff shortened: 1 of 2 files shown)


selcukcihan:
@medikoo ready for review 🙏

medikoo:
@selcukcihan great thanks for all your work. In terms of any further development in context of this project please reach out to @ac360

Mmarzex (approved):

PR #832, opened 2023-07-03
fix: Ensure NODE_PATH always contains required paths

Mmarzex:
Resolves bug reported in [SC-1236](https://linear.app/serverless/issue/SC-1236/node-sdk-import-module-error-reported-by-user)


Danwakeem (approved):
:shipit:

PR #836, opened 2023-07-06
fix: Instrument @smithy scoped smithy client

Mmarzex:


Danwakeem (dismissed):
Nice 👍

Danwakeem (approved):
Ooo yeah good catch 👍

PR #838, opened 2023-07-06
Fix logging instrumentation

selcukcihan:
### Description
* Related issue https://linear.app/serverless/issue/SC-1249/loggingwarning-cannot-handle-passing-an-exception
* Mimic the logic of `Logger.error` & `Logger.warning` in terms of the interpretation of the arguments. The crucial part that was not working is stringifying the first argument and using it as a formatter: https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L387
* Handle exceptions in the `warning` method as we do in `error`, passing it as is without stringifying it because it will get stringified in a different way within the `create_warning_captured_event` method.

### Testing done
Reproduced customer's exact issue and more in a unit test

### DIFF
--- a/python/packages/sdk/sls_sdk/lib/instrumentation/logging.py
+++ b/python/packages/sdk/sls_sdk/lib/instrumentation/logging.py
@@ -18,50 +18,51 @@ _original_warning = None
 _original_warn = None
 
 
-def _resolve_message(*args) -> str:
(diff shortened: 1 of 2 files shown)


selcukcihan:
@medikoo @Mmarzex ready for review 🙏

Mmarzex (approved):

PR #861, opened 2023-08-09
fix!: Use createSpan instead of createTraceSpan

Mmarzex:


Danwakeem (approved):
✅  Did we just not like the name of `createTraceSpan`?

Mmarzex:
> ✅ Did we just not like the name of createTraceSpan?

It was my bad, I misread the spec.

PR #882, opened 2023-09-29
Fix response tag handling

selcukcihan:
### Description
* Fixes https://linear.app/serverless/issue/SC-1653/python-sdk-error-user-instrumented-functions-and-all-functions-started
* My only explanation is that somehow the customer is using AWS Lambda to implement an http API but failing to return a dictionary with "statusCode" field in it and we fail to properly surface the issue.
* This will fix the status code extraction logic to make sure it doesn't fail and instead just report that the status code is missing.

### Testing done
Unit tested


selcukcihan:
@Mmarzex ready for review please 🙏

Mmarzex (approved):

